PR #1, opened 2018-05-03
Name-update started

ceriottm:
Please help substituting all instances of "proteus" in the codebase. My there are already many for a project that has existed for a few weeks!


tjunge:
Quick reminder, we had decided not to have any first level branches that
are not supposed to be permanent. Put temporary branches under feat
(i.e. feat/rename, no just rename)

PR #5, opened 2018-05-23
Feat/boost dep

felixmusil:
Remove the dependency on external boost unit test library. We now take care of building this static library.


felixmusil:
I am not too sure why the travis build is missing files... I will see that with Federico later

PR #9, opened 2018-06-15
Feat/dev tutorial with json dihedral

felixmusil:
Documentation:
I like the How to section very much. I think that this level of detail is needed.
One comment:  for the moment one needs to add .cc file names in the cmake files for them to be compiled.

Src: 
Very nice work.
I find the level of comments enough. It will be hard to reproduce in other parts rascal...
I thought there would be no neighbourlist in this tutorial to help focus on highlighting the peculiarities of Rascal. Since it is there I am not against keeping it but it increases the line count by 200.
I would rename the JSONTransfer namespace to io and also dedicate a file to it.

My current concern is with the Adaptor framework. The code for this tutorial would need major modifications to highlight the use of Adaptor and the Traits.


mastricker:
I am not sure what you mean with the .cc files names in the cmake files - do you mean for the documentation?
I actually did not look at that yet - I will have a look.

About the NeighbourhoodManagerBase:
I agree with the JSONTransfer for dedication of the file - it will be reusable in other variants. The thing is, that the interface to the specific JSON format is very specific - this implementation now has the bare minimum, which is output from ASE - at least the structure I tested.

About the linked list - this was more an exercise for me to get to grips with the Eigen::Maps and since the iterator does only work on atoms, if there is no neighbourlist, the testing was very limited.

I also agree with the adaptor framework. I will start working on the MaxLevel+1 Adaptor beginning next week. For clarity, this should also include a pair-level generator, like we discussed.

My suggestions about the chain-manager:
I will add the quadruplets by hand into the JSON file. Then, using an example, I will show how to calculate dihedral angles based on those lists and make it a property-field, which will then be put out into a JSON file again (input-computation-output). I'll strip the tutorial of the neighbourlist, put that into the MaxLevelIncreaser adaptor and get the JSON struct definition out of it.
The example will then be:
read file from json -> put data in neighbourhood manager -> calculate the dihedral angles as property -> output as JSON file.

Additionally: We should start thinking of an interface to control parameters of the managers. For example the cutoff distance, etc. Probably also with JSON files? 

So, thanks for the review, that was all I wanted to know, we are on the right track!
I'll work on it. I also don't think we need to reproduce the style in all of the files. As we said, this is specifically 'verbose' as a guideline.

felixmusil:
I agree with you that since it is supposed to be a simple tutorial one can add by hand the dihedral. We will see later on if this is enough for future user to get the hang of it.

I am not sure of a need for an input interface in c++. I would be more inclined to keep this strictly in python but if you find a situation where it is needed then we can make it.

I see what you mean about the JSONTransfer namespace. Let's decide next meeting if we want to follow fully ASE interface. If so I will start an Atoms class. This should be straightforward except for the layout of the 2D arrays.

mastricker:
I have now added an example file to calculate the dihedral angles with the chain manager.
So far, I am not completely satisfied, because the atom/pair/, i.e. the cluster ref is not used. But maybe we do not need it in this case. I cannot think of a 'nicer' way to define the quadruplets right now.

Before polishing this further, I am gonne work on the adaptor. I'll do it in the same branch, since it sort of belongs together, because I will loose the neighbourhood build.

PR #11, opened 2018-08-22
FPS sparsification function

ceriottm:
So the idea is to include in librascal a couple of subroutines I've been working on in the past week. So that every time I need this I have to import _rascal.
I'm still working on it but I'd appreciate comments on how badly I'm breaking the coding convention that I have read but probably not fully interiorized.


felixmusil:
Hi Michele, 
so the first thing to change is the number of character per line which should be 80, e.g. the comments might be better above the line it refers to. You also need the big header at the begining the c++ files you added (copy from other .cc). Otherwise, the comments are plenty and make the code crystal clear.

You need to write your own tests in c++ (some very basic ones see test_lattice.hh/cc) and in python to test the bindings and maybe some more advanced tests (see python_structure_manager_test.py).
You miss the default parameters in the bindings (see https://pybind11.readthedocs.io/en/stable/basics.html#default-arguments).
I get plenty of warnings when compiling. Is it the same for you ? Was it like this from the start of your branch ? Keep in mind that there shouldn't be any warning with the -Wall -Wextra -Weffc++ flags when merging.


Some advices for the implementation:
+ to compute feature_x2 you might want to use rowwise() (https://eigen.tuxfamily.org/dox/group__TutorialReductionsVisitorsBroadcasting.html) to perform the operation as a oneliner so that Eigen can optimize the operation on its own.
+ I would write a function for the distance computation so that later on it is easier to switch between different distance/kernel definitions.

felixmusil:
I am almost done with laying out the basis of a proper library importation from python outside of the librascal. Some issues came up and it took some time to understand and solve them.

ceriottm:
thanks felix. what warnings are you getting? I don't get any.

felixmusil:
I get warnings about -Werror=effc++ in the bindings part. I suspect it has to do with the bindings or my mac...

felixmusil:
I just pushed a way to expose the bindings in the python module. Tell me what you think and I guess we also should discuss this next meeting. 

Basically, I added an install instruction to our cmake procedure and made sure the relative dynamic library linking (from _rascal.so to librascal.dylib) are working in MacOS. This is what I have at hand and it seems more complex to handle than linux but it might not work as is in linux. Please tell me how it works for you. 
There are other ways to tackle this problem such as using a setup.py with setuptools and I don't know what is best in the long run but, for the moment, this way works.

So now you can do:
```
mkdir build; cd build;
cmake ..
make
make install
```
and import the bindings from outside of the library (ipython shell or notebook) with:
```
import sys
sys.path.insert(0,'/path/to/librascal/')
import rascal as rsc
rsc.lib.fps?
```

ceriottm:
So, test has been added, and code has been cleaned. I'd wait to add unit tests for the python bindings because the library interface is still under scrutiny, if I understand correctly. IMO we could merge this and have corrections to the library scheme and a separate regtest as another PR

PR #20, opened 2018-10-16
Feat/adaptor half neighbour list

mastricker:
The adaptor which makes a half/minimal neighbour list from a full one seems to work. I have also added a part of the documentation to comment on the ``Adaptor`` vs ``Manager`` thing and why we call them the same in the following.

The sorting to the half list is according to the atom index: second index of a pair has to be larger than the first.
static_asserts are included so that the adaptor only works on top of a pair list.

Please have a look and suggest possible changes and/or additional tests, which I could include.

Questions: should the arrays sizes used in this adaptor be hardcoded to a size of 2?
It should not be necessary, since it is checked for the MaxOrder, which should be exactly two.

Another thing: The access of e.g. 'distances' from an AdaptorStrict is not included right now.
Do we want to enforce the layering to
full pair list -> minimal pair list -> strict pair list?
Then, the problem would not exist.


felixmusil:
quick comments:
+ you should put all of you includes inside the #ifdef guard. The way it is in the .hh you added is not proper and will create an error in some cases.
+ there is a get_distance function in the manager strict (in master at least).
+ I think that we should start making .hh and .cc again to keep implementation in the header minimal to help readability.
+ I am not sure what you mean by ' enforce the layering' ?
+ in  src/structure_managers/adaptor_half_neighbour_list.hh:250 shouldn't it be this->nb_neigh[i] instead of this->nb_neigh[i-1] ? I am saying this because it was the buggy part of the adaptor strict that I had to fix so that it works properly. The bug is that from the second center and on, the lookup of the neighbourg list is shifted (it skips some neighbours).
+ Since you put an assert we shouldn't have hard coded order.
+ in  src/structure_managers/adaptor_half_neighbour_list.hh:345 you use .back() to get the atom index ? If so, please use .get_atom_index() which does the same but tells what it does or rename the .back() and .front() functions so that they are more explicit on what they do.
+ for testing I would do A=\sum_{i<j} (d_{ij}+d_{ji}) with the minimal NL and B=\sum_{ij;i!=j} d_{ij} with the full NL and check A==B. If it is not right it should be close to what I wrote :)
+ another test would be when you have the half to full to make some cross ckecks between them that would be for later.

mastricker:
Good feedback, thanks!

Fixed:
- header guard
- nb_neigh[i-1]; your comment makes sense, changed that.
- get_atom_index: changed that also; and i took the index_i initialization out of the loop of the pairs, since it will stay the same, missed that one the last time.

Comments: 
- get_distance: can only be accessed via the strict manager. My question relates to this: do we want to be able to query the distance with any structure manager and adaptor, because then we need to have a mechanism for pushing the request through (like position). Therefore I argue, that it should not be included. Because half neighbour list does not contain any distances.
- enforce layering: I mean, that we do not allow to have a different stacking than
full pair list -> minimal pair list -> strict pair list
so, something like this would not be possible:
full pair -> strict pair list -> minimal pair list

About the tests:
- I agree with the back and forth from full-half-full again, will include that in the other test.
- Your other suggestion: is the "d" the distance? And I don't understand the (i!=j) condition. If that is the atom index, this can not happen. We do not have pairs which are 1-1 or 2-2.

I have pushed the first comments - please wait with the merge until I have included the suggested test, which I do not understand right now.

Thanks again,
Markus

PS: How can I access the Travis report to see whats going on? I can only see the details of the test configurations, but not the actual error?

gibertif:
You can check the full log directly on Travis. 
You can either access it by logging on the website or by following the Details link at the end of the pull request and then click on the JOB id in Jobs and Stages.

There is a Log for each Job ID that reports the entire installation + compilation + testing.

I can help you find it and read it, in case slack me.

felixmusil:
>     * Your other suggestion: is the "d" the distance? And I don't understand the (i!=j) condition. If that is the atom index, this can not happen. We do not have pairs which are 1-1 or 2-2.
>       So you're saying, you basically want to sum up all distances?

Yes, d_{ij} is the distance between atom i and atom j and our NL/Adaptors are there to replicate such sums. So by design i!=j in the case of full NL and j>i in the minimal NL. 

Is it clearer now ?

mastricker:
Yes, I added it. Good test - it caught, that i forgot to add a get_neighbour_position()!

Btw: I will clean up the tests a littile, while I am at it. I think we are not consistent with "constructor" tests and actual test cases. The fixtures should provide the complete environment to check to specific stuff. I will comment again, once I am done with it.

mastricker:
So, it seems the half list adaptor seems ready. I have merged with master. Tests pass. The stacking also works.
Along the way: 
Memory issues were solved, some reconfiguring of the get_ methods. But I will look those over once I make a "feat/clean_master" as per the defined TODOs.

Please check if it works for you and if you have any other suggestions at this point.

PR #32, opened 2018-10-24
Feat/test structure managers

ceriottm:
I open a PR early because honestly I don't see much more that can be done in terms of the testing of the common interface. Also since Felix is cleaning up structure managers we might want to avoid duplicate efforts for structure managers that are going to disappear anyway.

I also have questions as to whether (1) we should also test protected features and (2) whether we should also test loops over neighbors, and this neighbour_position function that I still do not understand.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,7 +9,6 @@ set(BUILD_SHARED_LIBS ON)
 
 
 add_compile_options(-Wall -Wextra)
-add_compile_options(-Werror)
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Weffc++")
 # to avoid conflicts with built-in functions
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fno-builtin")
@@ -98,6 +97,7 @@ if (${BUILD_TESTS})
 endif(${BUILD_TESTS})
 
 # compile the library
+add_compile_options(-Werror) # warnings are errors! 
 add_subdirectory(
   ${CMAKE_SOURCE_DIR}/src/
   )
--- a/src/structure_managers/structure_manager_chain.hh
+++ b/src/structure_managers/structure_manager_chain.hh
@@ -212,6 +212,11 @@ namespace rascal {
                       this->cell_data.size()/traits::Dim);
     }
 
+    inline int get_atom_type(const int index) {
+      auto t = this->get_atom_types();
+      return t(index);
+    }
+   
     //! Returns the type of a given atom, given an AtomRef
     inline int get_atom_type(const AtomRef_t& atom) {
       auto index{atom.get_index()};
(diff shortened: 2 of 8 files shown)


mastricker:
I think this is a good start for the "test-homogenization".

Things I noticed (also during the clean up of the master yesterday, which is not done yet):
1) We need to be more specific with the commenting (copied from our coding convention), comments behave differently depending on (also thanks to Till for clarification):
>>>>
Comment Style

Use either the // or /* ... */ syntax for comments that are only relevant within their context (local comments for the reader/maintainer of your code)

Use //!, /** ... */, and //!< syntax for doxygen comments (which will end up being compiled into the API :ref:`reference`), see Doxygen for details on doxygen.
<<<<<
The state now does not adhere to it, but I can do that while I clean up the rest anyways.

2) 80 character line length limit, never gets old.

And maybe you can change that in your editor but:
3) Trailing whitespaces

Bottom line: Code looks good, compiles for me. I support the pull request under the prerequesite, that I take it over from there and include it in the 'clean up'. And then we work at it again once we cleaned up the obsolete structure manager implementations.

mastricker:
I'm for merging all open requests now, since I also worked on the documentation of the code. And Tills fix for the errors. And the newest adaptor for extending the neighbourlist.


Then we can start from the "most polished" point for the merging of the managers and extend the tests.

I would also separate the order=1 and higher for testing. I think it is a good thing to keep the basic test for just order one, because that is what we need to test when someone adds a new manager. And then he can take advantage of the infrastructure.

On Oct 26, 2018 16:26, felixmusil <notifications@github.com> wrote:

it seems like a good begining. I see 2 directions to explore:

  *   Make tests that depend on the traits (maybe not all but at least the order), at the moment only order 1 features are tested. It should be possible to make conditions on the traits as to avoid testing features that don't exists and have a ManagerFixture that initialize adapted mangers (like AdaptorNeighbourList) with the template argument (I am not sure at the moment how one could do that, but this https://en.cppreference.com/w/cpp/language/parameter_pack seems like the way to do).

  *   We need to converge on a few structures that are increasingly complex (for the moment there is a simple cubic and a triclinic structure). And have a simple way to switch the structure.

We probably can merge master into this branch when the obsolete managers will be removed and continue the test streamlining.

What do you think ?

-

ceriottm:
Fine for me to merge -- I wish I could have done more but at least I could delve a bit into the code and helped clarify this get_neighbour_position thingie.

mastricker:
I understand, but this kind of reviewing is really helpful. Being questioned about our design decisions at stragglers.
Stating the obvious: this makes it cleaner an more robust.

On Oct 26, 2018 17:29, Michele Ceriotti <notifications@github.com> wrote:

Fine for me to merge -- I wish I could have done more but at least I could delve a bit into the code and helped clarify this get_neighbour_position thingie.

-

felixmusil:
Let's keep this branch for further work

PR #35, opened 2018-10-26
Feat/fix clang errors

tjunge:
The error came from using a `auto` value (as opposed to rvalue, which would be safe if moved from, but out of the question here, because we use value twice) to store the result of a `Eigen::solve()`:
`auto multiplicator = cell.ldlt().solve(xpos)`**.eval()**`;`

 Eigen's expression templates need to be either evaluated or cast into a type of which we **know** they own their memory, else they cannot be safely stored.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ make all
 ctest -V
 ```
 
-
+cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DBUILD_TESTS=ON  CMAKE_C_FLAGS_RELWITHDEBUBINFO="-03 -g -DNDEBUG" ..
 * Special flags:
   + INSTALL:
     + empty (default) -> install rascal in the build folder
--- a/src/structure_managers/adaptor_neighbour_list.hh
+++ b/src/structure_managers/adaptor_neighbour_list.hh
@@ -2,7 +2,7 @@
  * file   adaptor_neighbour_list.hh
  *
  * @author Markus Stricker <markus.stricker@epfl.ch>
- * @outhor Till Junge <till.junge@epfl.ch>
+ * @author Till Junge <till.junge@epfl.ch>
  *
  * @date   04 Oct 2018
  *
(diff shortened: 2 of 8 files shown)


src/structure_managers/adaptor_neighbour_list.hh
```diff
@@ -2,13 +2,14 @@
  * file   adaptor_neighbour_list.hh
  *
  * @author Markus Stricker <markus.stricker@epfl.ch>
+ * @outhor Till Junge <till.junge@epfl.ch>
```
mastricker: author

mastricker:
Ok, that sounds plausible. I reviewed the other changes, completely agreeable, thank you.

Two things:
- Your @author addition in the license is an @outhor right now, I used the commenting tool here on github.
- -O3 flag for compilation necessary in release with debug?

Other than that, I would merge after those two things are done

README.md
```diff
@@ -27,7 +27,7 @@ make all
 ctest -V
 ```
 
-
+cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DBUILD_TESTS=ON  CMAKE_C_FLAGS_RELWITHDEBUBINFO="-03 -g -DNDEBUG" ..
```
mastricker: necessary with O3?

src/structure_managers/adaptor_neighbour_list.hh
tjunge: oops thot is obsolutely correct
```suggestion
 * @author Till Junge <till.junge@epfl.ch>
```

README.md
tjunge: The error appeared only in O3 in clang, so I'd say yes, for testing

mastricker: Ok, agreed.
I vouch for merge - Felix?

felixmusil:
The -O3 flag is necessary here since the bug did not appear otherwise. Thanks Till for this tricky debug, the auto feature is very nice but it also has some side effects...

PR #38, opened 2018-11-02
Feat/manager cleanup

felixmusil:
remove unused feature from master and simplify tests


felixmusil:
The only missing bit at the moment is to move the manager chain to the examples and then it is ready to merge

felixmusil:
Also in the atomic_structure copies the data onces after reading. 
Would it be possible to keep the data as vectors and make some getters that return EigenMaps ?

mastricker:
If I get you right:
You want a Map onto the json structure?
As far as I understand, this is not possible. The reason is that positions and cell is a vector of vectors - so we can not map onto that. This means, we first have to make a vector of just doubles and then map onto that with a stride of 3. At least I don't know of any way to solve that differently.

And to make it consistent and obvious, i opted to copy them all. But we can at least forget about the type and pbc data.

About the manager_chain. Would you move the source file to the examples section?

felixmusil:
I see. The only way to not copy is to read as a flat vector indeed. I guess that it will not be a problem for the moment.

The manager_chain is used in: its tests and in alanine_chain.cc where it is built but not used. So basically it is unused. What would you like to do with it actually ?

mastricker:
Ok.
I think we can delete it. What was the documentation of the code in the chain manager became *centers. As far as I understand, this has found its way.

So: I will delete it and make the example with the dihedral angles based on _centers?
What do you think?

felixmusil:
Sounds good to me.

mastricker:
I think this here can be closed: I need the refactored tests in order to change the ManagerChain dependency. I will to that in the branch feat/better_tests.

Merge?

PR #49, opened 2018-11-28
Feat/mrclean

felixmusil:
We changed the header and used cpplint.py to change to check for conformance.

note that we used the followings options:

"cpplint.extensions": [
      "cpp",
      "h++",
      "cuh",
      "c",
      "c++",
      "cu",
      "hxx",
      "hpp",
      "cc",
      "cxx",
      "h",
      "hh"
    ],
    "cpplint.filters": [
      "-build/include",
      "-runtime/indentation_namespace",
      "-runtime/references",
      "-legal/copyright",
      "-whitespace/comments",
      "-readability/namespace",
      "-readability/alt_tokens"

    ]


tjunge:
hi @felixmusil @max-veit @ceriottm, 
We (Markus and I)'ve cleaned up the boilerplate as per Max's comment. Should be ready to merge

mastricker:
merging

tjunge:
> Aaargh, I was just about to do this myself (see the updated header and vimscript in the scripts directory). This way it's not such a pain to change automatically, and we have a single reference header to go off of. Which includes, btw, calling this software `librascal` rather than `Rascal` -- right?

What script folder?

max-veit:
> What script folder? Does that mean we shouldn't delete this branch yet?

See latest commit (that got merged with this pull request, but I wasn't done yet so I had to create another PR). It's in the top level because none of the other folders seemed appropriate. We could just as well call it `tools/` and put `cpplint.py` in there as well.

But once that's merged I guess we can go ahead and delete it.

> also, we once decided that the library is called librascal, but the project is rascal, so I think that can stay the way it is.

Ok, header template updated.

PR #62, opened 2018-12-14
Feat/test ci

gibertif:
Added:
clang 4.0
gcc 6.0


.circleci/config.yml
```diff
@@ -206,11 +256,68 @@ jobs:
...
+                                command: |
+                                        cd build 
+                                        make dev_doc
+
 workflows:
         version: 2
         ubuntu:
                 jobs:
                         - build_ubuntu_g++-5.0
+                        - build_ubuntu_g++-6.0
```
mastricker: can we make this 6.4? because it is the default on ubuntu 18.04?

.circleci/config.yml
```diff
@@ -206,11 +256,68 @@ jobs:
...
+                        - run: 
+                                command: |
+                                        cd build 
+                                        make dev_doc
+
 workflows:
         version: 2
         ubuntu:
                 jobs:
                         - build_ubuntu_g++-5.0
```
mastricker: Can this be 5.4 (Ubuntu 16.04 standard)

.circleci/config.yml
```diff
@@ -206,11 +256,68 @@ jobs:
...
+                        - build_ubuntu_g++-6.0
                         - build_ubuntu_g++-7.3
+                        - build_ubuntu_clang-4.0
                         - build_ubuntu_clang-5.0
```
mastricker: Can we add clang-6.0 (default on ubuntu 18.04?)

gibertif:
I will add clang-6.0, and if you want I can specify the .4 for the gcc compilers, but it is alredy the one that is used, i.e. if you open the CI log you see that the compiler used is 6.4. I will add it though, so that we are sure what we are using.

mastricker:
Ok I am not sure, what actually went wrong - can you please resolve the conflicting file - I don't want to mess it up.
When its done, I squash-merge and then merge the other PR with the new master, so everything is up to date.
Thanks!
Markus

PR #68, opened 2018-12-28
Feat/fps redux

ceriottm:
I think this is ready to be merged. It's quite urgent that we clean up the bindings and perhaps separate better the unit tests of the pybind11 wrappers, and the more complex tests of actual methods.


ceriottm:
> Seems good to me. I have a few comments though:
> 
> * rascal/utils/fps.py:89 The return type should be discussed. Why return a dictionary rather than multiple variables ?

This was briefly discussed during one meeting. Since there are different methods available, FPS can return different things. A dictionary makes it easy to access the return values without having to remember which member of the tuple is what. 
This is similar to how many scipy functions behave, e.g. optimize

> * src/utils/sparsify_fps.cc:49/176/180 I would avoid 1 line if without curly brackets if not on the same line. It could lead to some ugly bug latter on.
Can you make these comments as inline code comments? makes it much easier to incorporate

> * tests/test_utils_sparsify.cc We use the alias size_t instead of unsigned int type in the code. You might want to follow this trend to increase the code uniformity.
ditto

PR #69, opened 2019-01-07
Feat/fix instal

felixmusil:
changes to the installation procedure, bug in python executable usage fixed and move of files. 

The 2 new flags to set the installation behavior:

* Special flags:
  + -DINSTALL_PATH:
    + empty (default) -> install rascal in the detected python lib folder
    + custom string -> path to where rascal will be installed
  + -DUSER_INSTALL:
    + OFF (default) -> changes nothing
    + ON -> install in the user's site package of the detected python library

by default it will install in the global python site-packages folder.


ceriottm:
ok so make; ctest works.
make install still has problems with libcephes

```
<ipython-input-1-1a84ae0f7445> in <module>
----> 1 import rascal as ras

~/local/lib/python3.6/site-packages/rascal/__init__.py in <module>
----> 1 from . import lib, representation, utils

~/local/lib/python3.6/site-packages/rascal/lib/__init__.py in <module>
----> 1 from ._rascal import (StructureManager, FeatureManager, Adaptor,
      2                      RepresentationManager, utils, math)
      3 
      4 from ._rascal.utils import sparsification

ImportError: libcephes.so: cannot open shared object file: No such file or directory
```

felixmusil:
I think that it is not a systematic problem with this library linking. Basically, math-rascal.so depends on it so it should have a path to find it.
When you make it should reset the path to nothing so that it finds it since it is in the same folder.
I don't know how to reproduce the error so I won't be able to help you.

How did you solve it before ?

Félix


Le 7 janv. 2019 22:35, Michele Ceriotti <notifications@github.com> a écrit :

ok so make; ctest works.
make install still has problems with libcephes

<ipython-input-1-1a84ae0f7445> in <module>
----> 1 import rascal as ras

~/local/lib/python3.6/site-packages/rascal/__init__.py in <module>
----> 1 from . import lib, representation, utils

~/local/lib/python3.6/site-packages/rascal/lib/__init__.py in <module>
----> 1 from ._rascal import (StructureManager, FeatureManager, Adaptor,
      2                      RepresentationManager, utils, math)
      3
      4 from ._rascal.utils import sparsification

ImportError: libcephes.so: cannot open shared object file: No such file or directory

ceriottm:
I think part of the problem is that this is a C++ project with python bindings, so we should install C++ libraries in whatever is the chosen (system-wide or user) library path. perhaps I broke things changing the CMAKE_INSTALL_PATH

ceriottm:
so, I looked a bit around how these things should be done proper. we should have librascal.so in the system (or user) lib folder, librascal-math.so and libcephes.so in a rascal/ subfolder (apparently there's way to tell librascal.so that it can find those in that subfolder) and then the python bindings in the python path. 
if we have a C++ library that can be linked stand-alone, it shouldn't be caught up in the python environment mess.

felixmusil:
So before doing any rash move Michele I have some clues. 
I reproduce your problem on a ubuntu 18.04 machine while I went to another ubuntu 16.04 machine (not mine) and it works smoothly like on my machine. I will try to dig more into that.

What you are suggesting looks more like a mess to me than a solution Michele. I see how it can grow into a nightmare when several installations have been made without over writing onto each other making it hard to understand what is linked to what being all over the place.  
Note that on a cluster since you can't install the .so as you described they will have to be set up in a custom folder that will need to be given to any external code that would link to it. So your suggestion does not addresses cluster context I think. 

In the current configuration one can't get librascal.so without building the bindings. Do you think it should be possible ?
Also for the moment, all the libraries are installed in one place that can be be passed to any other code at compile time to link it.

ceriottm:
IMO we should 
(1) generate a single C++ library - there's no reason to have a separate libcephes.so and math-rascal.so if we are pulling the source into our repo
(2) it should definitely be possible to link a C++ code to librascal.so without having to bring about the bindings
(3) I'm not rushing to conclusions - I did quite some reading and the reason we're having so many problems is that we're doing a rather non-standard setup. C++ libraries should live in a library path -- not necessarily /usr/lib, you can obviously have a local library path that is defined in LD_LIBRARY_PATH in your bashrc.

felixmusil:
(1) there is a very good reason for separating the c library and the c++ side: avoid polluting the namespace.
(2) we can but nobody mentioned it before so it does not exist. Note that it would introduce 2 build/install/test procedures that will have to be maintained in parallel. Is there really a need to avoid having python as a dependency considering this cost ?
(3) I understand that you read about this topic but I think that you are still rushing to the conclusion that this issue is tied to the library and not its environment. Could you try to also look into it please ? 
Also you say 'so many problems' but I see only one at hand.

ceriottm:
Think this needs some valgrinding. main_test_suite fails when compiling in Release mode, and not when compiling in debug mode.

tests/python_structure_manager_test.py
```diff
@@ -6,7 +6,8 @@
 sys.path.insert(0,'../tests/')
 
 from test_utils import load_json_frame, BoxList, Box
-from python_import_rascal import _rascal as rc
+import rascal.lib._rascal as rc
```
ceriottm: +1

PR #71, opened 2019-01-27
change options handling of the representations

felixmusil:
Here are the changes for this merge:

- change the option mechanism for the representations.
- add check hypers compatibility mechanism in the representation base.
- make sure the python wrappers are copied at each build.

These changes follow our discussion last meeting.


src/representations/representation_manager_base.cc
```diff
@@ -30,9 +30,55 @@
...
+    }
+    return out;
   }
-}
+  /* ---------------------------------------------------------------------- */
+  std::string RepresentationManagerBase::get_hypers_string() {
+    return this->hypers.dump(2);
+  }
+  /* ---------------------------------------------------------------------- */
+  void RepresentationManagerBase::check_hyperparameters(
```
max-veit: This function checks for presence of required options and for correct values of switches, but isn't this validation that we said we could do with json schemas? I can't remember our decision on using schemas, but they do have the additional benefit of type specification and validation of numerical values (that would be a pain to reinvent here).

src/representations/representation_manager_base.hh
```diff
@@ -33,25 +33,18 @@
...
-
   class RepresentationManagerBase {
    public:
     //! type for the hyper parameter class
     using hypers_t = json;
     //! type for representation
-    // TODO(felix) Should the user have freedom for the type ?
     using precision_t = double;
+    //! type used to register the valid key and values of hypers_t
+    using reference_hypers_t = std::map<std::string, std::vector<std::string>>;
```
max-veit: Is `reference_hypers` going to be used for both hypers (generally numerical parameters) and options (switches that change the behaviour of the class)? Does the initial validation pass treat both of these on an equal footing? If so, where does the split between options and hypers happen?

tests/test_feature_manager.hh
```diff
@@ -58,7 +58,13 @@ namespace rascal {
       {{"central_decay", 0.5},
        {"interaction_cutoff", 10.},
        {"interaction_decay", 0.5},
-       {"size", 120}}
+       {"size", 120},
+      {"sorting_algorithm", "distance"}},
```
max-veit: Indentation please (otherwise it looks like it's part of the wrong level)
```suggestion
       {"sorting_algorithm", "distance"}},
```

tests/test_feature_manager.hh
```diff
@@ -58,7 +58,13 @@ namespace rascal {
...
+      {{"central_decay", 0.5},
+       {"interaction_cutoff", 10.},
+       {"interaction_decay", 0.5},
+       {"size", 120},
+      {"sorting_algorithm", "row_norm"}}
```
max-veit: Same here
```suggestion
       {"sorting_algorithm", "row_norm"}}
```

tests/test_representation_manager.cc
```diff
@@ -48,7 +48,8 @@ namespace rascal {
     true_order << 0, 1, 3, 2, 4;
 
     auto test_order = internal::SortCoulomMatrix<
-        Option::CMSortRowNorm>::get_coulom_matrix_sorting_order(test_matrix);
+      internal::CMSortAlgorithm::RowNorm>::get_coulom_matrix_sorting_order(
```
max-veit: Maybe fix the spelling with search and replace?
```suggestion
      internal::CMSortAlgorithm::RowNorm>::get_coulomb_matrix_sorting_order(
```

tests/test_representation_manager.hh
```diff
@@ -56,19 +56,24 @@ namespace rascal {
       {{"central_decay", 0.5},
       {"interaction_cutoff", 10.},
       {"interaction_decay", 0.5},
-      {"size", 120}}
+      {"size", 120},
```
max-veit: Again, please clarify nesting levels with indentation

src/representations/representation_manager_base.cc
felixmusil: I don't remember either but I think that is would be good to keep such checks close to the implementation. Also having them too close to the json reader would mean a lot of work if we change/extend the reader capabilities.
In this function, I assume that the input to check has the same functionality as a stl map so it should allow for varying data representation.

src/representations/representation_manager_base.hh
felixmusil: I agree with you it is tricky. I am treating them as equals here while one could imagine having option dependent hypers. The check here looks whether the field name is valid and if it as only some possible values then check if the value is valid.
The broader question that you raise is: should we check for internal consistency thoroughly ? e.g. given the representation and a set of options should we make sure that all the required hypers are present ? Is that right ?
I am not sure it is needed at this stage but if we want to go down that way we need to make the registration as simple as possible.

src/representations/representation_manager_base.cc
max-veit: Fair enough -- in any case, it's good to keep it encapsulated in this check function so we can add or change functionality later as needed. It might still be a good idea for the future to align the semantics of the reference_hypers with that of the json schema, both so that it's familiar to schema users and so that we can switch later if needed. It's not a huge change; I think it's just instead of mapping a string to a list of strings, we map it to another map that contains fields such as `required` (boolean), `type` (str, int, enum, etc.), and `allowed_values` (or whatever it's called for a enum-like type).

PR #73, opened 2019-02-04
Feat/only cosmetics

mastricker:
Two new make targets have been added to make use of autoformatting features:

`make pretty` 
Uses clang-format with the `.clang-format` specification to autoformat all hh/cc files.
Specific regions like matrix definitions, which should be user readable can be ignored using
```
// clang-format off
SOME CODE TO IGNORE
// clang-format on
```

`make prettypython`
Uses autopep8 with the `.pycodestyle` specification to autoformat python files according to pep8 standard.
Lines, which should be ignored can be specified by using the comment
`# noqa` (no quality assurance)

Any other suggestions for formatting?

Otherwise, if autopep8 and clang-format are installed, formatting code is a breeze now.

The only thing, which is missing, is the autoformatting of python doc strings. But there would be a solution, a package calles [`docformatter`](https://github.com/myint/docformatter)
Any opinions on that?
A small local test on my side showed that it does what it should.


felixmusil:
Hi Markus,
seems to work properly. 

For the docstring in python I started to use the numpy format and many packages provide a template generation for functions class and all. I think that this format is good, do you agree ?

cmake/autopep8.cmake
```diff
@@ -0,0 +1,100 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+# target to run autopep8.py for all configured sources
+set(AUTOPEP8_TARGET prettypython CACHE STRING "Name of python autoformatter")
```
felixmusil: why do you cache this variable ? same for the other AUTOPEP8 variable and CLANG_FORMAT.
Also caching them tend flood the ccmake list of option so I would either not cache them or rename them so that more important variables appear first. What do you think ?

mastricker: No specific reason. I basically adapted what Till had adapted for the linter. It was already in there, so I kept it to be compliant with existing stuff.

As for the flooding: I agree, more important stuff should be further up. But  I think it should be there. I know you are not (or at least were not) using the ccmake to configure. But I like the interface to get a choice of options. I do not want to dig very deep in makefiles for a quick overview.

What do you think?

felixmusil: So how about a prefix in front of such cached variables ? Like Z_ so that it appears at the bottom.

mastricker: Sounds like a good alternative. Did not think of it.
I will do that.

mastricker: Ok, so I cleaned it up. I also marked the linter as advanced. It is optional and has nothing to do with the compilation in the first place

What do you think about the not-advanced-appearance?

mastricker:
Sorry - forgot to document.

I have added the ignore comments for clang and autopep to the readme. So that people, who use it have the chance to find it fast, in case they want to ignore some lines like matrices, etc.

mastricker:
Ok, fixed the conflicts and made some cosmetics on master including header guard formatting corrections.

One thing I noticed: when using the verbose version of ctest, it spits out a very long winding print from the soap vector. I am fully for testing this, but it seems misplaced here - it totally clobbers the output.
What do you think?

Ready to merge after checks turn out well?

felixmusil:
I took care of the printing in the tests and the merging too.

PR #79, opened 2019-04-08
Feat/species filter units

mastricker:
Hi,

 I put it as a PR so we can see in which direction we will merge and drop this one or whatever and the discussion can be tied to lines of code easily.
I guess we will merge this into the `feat/fix_rec_update` and drop the PR?

Included here:
- New base class `Updatable` , which in brings the update functionality to structure, adaptor, species and filter, like we talked about.
- As far as I can see the update functionality should work, but it might be that I forgot to add something. But that will probably come out once we try the recursive update?
- Basic units checking with unit system. We already kind of talked about this. See if you can get the idea and please give me feedback. This has to be worked out.
- `SpeciesManager` now works with a `StandardisedTuple` for the access operator instead of the convoluted impossible tuple type we used before. Till and I find the naming unfortunate, I am still thinking about it. So the name might change.

So in essence, a work in progress towards the recursive update functionality including read in.
Next step is the `RepresentationCalculator` instead of `Manager`. And of course the Behler-Parinello calculator.

Another thing I noticed: Our tests take reather long. I have dialed back the multiplier for the neighbour list built a little. But that only got the test down to 3.13 seconds. I hope we can bring that down to below 1 second (linting not included).

Cheers,
Markus


mastricker:
This branch was merged into `feat/fix_rec_update` and another PR is unnecessary, therefore closed.

PR #83, opened 2019-04-12
Feat/test key standardisation

mastricker:
I made the test for `KeyStandardisation` better with the help of Till's input to ensure that the `reinterpret_cast` is tested.

Further, I added an initial documentation on the specifics of the adaptors as well as the species manager.


felixmusil:
The test and renaming seems fine to me.

For the documentation of the Adaptors, I would move it under the NeighborList section so that StructureManagers and SpecieManager are grouped together under this section.

mastricker:
Ok, good point. I changed the ordering.
Once Alex pushed the `StructureManager` documentation we probably should discuss the ordering and naming again. And also the granularity.
This is just a first attempt, I think we need more details regarding the actual implementation. But on the other hand there could also be a potential mismatch between docs and actual code.

I would argue that the documentation I added should stay rather high level - explaining the rationale behind the design choices. But the actual implementation is documented in the code.
Or we add the code snippets like you (Felix) started to connect the idea to the implementation.

mastricker:
I agree.
I also put up a discussion point for Thursday to discuss the documentation.

PR #89, opened 2019-05-09
Feat/soap dot

felixmusil:
I am adding a kernel builder feature to the block sparse feature manager (it should not stay like too long but the functionality is there now so that the code is useful). 
I also found a small but important bug in the internally sorted map which will make the overall soap thing faster: the check making sure that (a,b) == (b,a) in soap was not working properly.
I improved the soap notebook with some more features and removed the code that should not be use anymore. 
I added the normalization of the soap features and put it as an option.

Note that the kernel computation is not parallel so quite slow. Moreover there might be some slowdown because of the sparsity.


src/representations/representation_manager_soap.hh
```diff
@@ -230,8 +236,8 @@ namespace rascal {
                   // TODO(andrea) pre compute l_factor
                   double l_factor{1 / std::sqrt(2 * l + 1)};
```
mastricker: If I am not mistaken, this should be changed to 
`double l_factor{pow(std::sqrt(2*l + 1), -1)};`
In a small test I made, for small `l`, `pow( X, -1)` outperforms `1 / X` by quite a significant amount.

src/structure_managers/property_block_sparse.hh
```diff
@@ -114,7 +132,7 @@ namespace rascal {
       template <class Key>
       decltype(auto) count(const Key & key) {
         key_type skey{this->copy_sort(key)};
```
mastricker: CamelCase

src/representations/feature_manager_block_sparse.hh
```diff
@@ -264,6 +291,204 @@ namespace rascal {
...
+
+  /**
+   * Matrix multiplication between two block sparse feature matrices
+   */
+  template <typename Precision_t>
+  decltype(auto) dot(FeatureManagerBlockSparse<Precision_t> & X1,
+                     FeatureManagerBlockSparse<Precision_t> & X2) {
+    auto n1_centers{X1.sample_size()};
+    auto n2_centers{X2.sample_size()};
+    Dense_t<Precision_t> mat =
```
mastricker: Maybe I am getting the math wrong here (which is totally possible), therefore for clarification:
When two block sparse feature matrices from two different structures where not keys of no centers intersect, the matrix here stays zero, like initialized.
And that is what is actually wanted, right?

src/representations/feature_manager_block_sparse.hh
```diff
@@ -264,6 +291,204 @@ namespace rascal {
...
+    }
+    return mat;
+  }
+
+  /**
+   * global cosine kernel with a power of zeta
+   */
+  template <typename Precision_t>
+  decltype(auto)
+  cosine_kernel_global(int zeta, FeatureManagerBlockSparse<Precision_t> & X1,
```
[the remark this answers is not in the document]
mastricker: Further: Is the `.dot` function a subset of the cosine?
I.e. could one call `.dot` for two BlockSparse feature matrices and then evaluate the cosine (to the power zeta) on the given matrix to avoid code duplication?

src/representations/feature_manager_block_sparse.hh
```diff
@@ -264,6 +291,204 @@ namespace rascal {
...
+    auto stride1{X1.get_structure_stride()};
+    auto stride2{X2.get_structure_stride()};
+    auto n1_frames{stride1.size() - 1};
+    auto n2_frames{stride2.size() - 1};
+
+    Dense_t<Precision_t> mat(n1_frames, n2_frames);
+    // loop over the structures of the 1st set
+    for (size_t iframe1{0}; iframe1 < n1_frames; iframe1++) {
+      auto & iframe1_st{stride1[iframe1]};
+      auto & iframe1_nd{stride1[iframe1 + 1]};
```
mastricker: I know that you like to use `nd` for `end`...but it took me again half a minute to figure it out. Can we make `_nd` to `_end`, pleeeeeease :)

src/representations/feature_manager_block_sparse.hh
```diff
@@ -264,6 +291,204 @@ namespace rascal {
...
+    auto stride1{X1.get_structure_stride()};
+    auto stride2{X1.get_structure_stride()};
+    auto n1_frames{stride1.size() - 1};
+    auto n2_frames{stride2.size() - 1};
+
+    Dense_t<Precision_t> mat(n1_frames, n2_frames);
+    // loop over the structures of the 1st set
+    for (size_t iframe1{0}; iframe1 < n1_frames; iframe1++) {
+      auto & iframe1_st{stride1[iframe1]};
+      auto & iframe1_nd{stride1[iframe1 + 1]};
```
mastricker: `nd` to `end`, pleaseeeeee

src/representations/feature_manager_block_sparse.hh
```diff
@@ -264,6 +291,204 @@ namespace rascal {
...
+    auto n1_centers{X1.sample_size()};
+    Dense_t<Precision_t> mat =
+        Dense_t<Precision_t>::Zero(n1_centers, n1_centers);
+    for (int icenter1{0}; icenter1 < n1_centers; icenter1++) {
+      auto && keys1{X1.get_keys(icenter1)};
+      mat(icenter1, icenter1) = 1.;
+
+      for (int icenter2{icenter1 + 1}; icenter2 < n1_centers; icenter2++) {
+        auto && keys2{X1.get_keys(icenter2)};
+        auto keys_intersection{intersection_of(keys1, keys2)};
```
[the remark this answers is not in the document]
felixmusil: each environement might have different species so yes

src/representations/feature_manager_block_sparse.hh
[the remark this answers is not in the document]
felixmusil: yes but the doing it like this involves a significant memory overhead: 
10k local environment -> dot generates a 10k x 10k intermediate matrix while there might be only 200 structures so the final kernel matrix is 200x200

src/representations/representation_manager_soap.hh
felixmusil: good to know

src/structure_managers/property_block_sparse.hh
felixmusil: this is std conformance

src/representations/feature_manager_block_sparse.hh
felixmusil: yes in this case. one could introduce some coupling but I would rather do this kind of operation before hand

felixmusil: Thanks for the small benchmark these small good practice add up in the end.

    feature_manager_block_sparse: I think we will need a schematic for the layout
    of the block sparse layout. The description in the beginning of the file is
    helpful, but a drawing would make it much more visual. I get the principle idea, but I am lost on the details. I tried to comment on the parts, which do the actual math.
1. since this will disappear soon I would rather not spend too much time on the details here. If you have questions I can always elaborate though.
2. there are not tests since this is temporary. Michele wants the functionality available now but the overall implementation will have to change quite a bit latter so the tests seems a bit too much to me. Of course I ran some comparisons and checks which allowed me to track down a bug so the computation works.

felixmusil:
Thanks for the small benchmark, these kind of good practice adds up to the code.

This functionality is in a temporary stage of implementation: Michele wants it there now so that the code is can be used but it will have to change in the near future. So I tested the soundness of the implementation and put some comments but since it will have to change quite drastically and soon I would rather not go beyond at the moment.

Issue #91, opened 2019-05-14
Default Functions for all StructureManagerImplementations


agoscinski:
This is a way to make default functions for all StructureManagerImplementation (Root and Adaptor), which can have different functionalities dependent on the traits. The only requirement I see is that you need to initialize the member variable in the StructureManagerInterface. A good candidate would be the get_property() because it has already its member variable in the interface. There is code below which gives an example how this could look like.

If we initialize an empty distance property in the interface, get_distance could be also implemented. The functionality would depend then on the HasDistance trait. The if case checking a trait in the get_has_property() function could be put into a std::enable_if with template parameters. I also have tested this, but I do not put it here, because the workaround to do partial specialization for functions makes the the code more complicated. We already do this kind of partial specialization in other parts of the code, so I guess you can see that this is possible in principle.

The code should be compilable and executable (but you have to manually with g++)
https://github.com/cosmo-epfl/librascal/blob/feat/default_implementation_functions/examples/default_functions_for_implementations.cc

agoscinski:
solved in #130

Issue #86, opened 2019-04-18
implement in structure_manager.hh a conditional get_distance that looks at the stack if any has the trait HasDistance is true.


felixmusil:

agoscinski:
this line looks wrong
https://github.com/cosmo-epfl/librascal/blob/master/src/structure_managers/adaptor_increase_maxorder.hh#L56
I think should inherit from underlying manager like HasDirectionVectors

felixmusil:
I agree with you.

Le 4 juil. 2019 13:27, agoscinski <notifications@github.com> a écrit :

this line looks wrong
https://github.com/cosmo-epfl/librascal/blob/master/src/structure_managers/adaptor_increase_maxorder.hh#L56
I think should inherit from underlying manager like HasDirectionVectors

mastricker:
I will need to think about that a little.

agoscinski:
the original issue is solved in #130 
see
https://github.com/cosmo-epfl/librascal/blob/feat/get_property_forwarding/src/structure_managers/structure_manager.hh#L630-L643
https://github.com/cosmo-epfl/librascal/blob/feat/get_property_forwarding/src/structure_managers/adaptor_strict.hh#L249-L253

but the issue if `AdaptorMaxOrder` should have the trait `HasDistance{false}` is still open. Should I open new issue for this?

PR #92, opened 2019-05-21
Bug/#84 86

agoscinski:
Solve issues #84 #85 #86 
### Implementation:
- Added get and create property functionality to structure_manager.hh and adapted the usage of the distance and direction vector property in AdaptorStrict to this new functionality
- Added ClusterRefClusterKey which was before ClusterRefKey. New inheritance is ClusterRefBase-> ClusterRefClusterKey<Order, Layer> -> ClusterRefKey<Order, Layer, ParentLayer, NeighbourLayer> -> ClusterRef<Order>. The new ClusterRefKey has a parent and a neighbour
- replaced compute_cluster_layer at several places with get_layer function because I do not think it make sense, feel free to disagree and comment on this
- Changed size_t to int at places where I believe an atom_index is used, I marked the positions with #ATOM_INDEX in the code. Changed name of member variable neighbours in adaptors to neighbours_atom_index. Added neighbours_cluster_index for #84 functionality.

### Still open to do
- For the case consider_ghost_neighbours = false, the neighbour atoms in pairs can be a ghost atom, but it is not in the cluster_indices_container of order 1. We decided to update in this case the real atom corresponding to the ghost atom. This is a bit tricky because they both have different atom_indices, but I havent tried to solve it yet. This is also the reason why the test fails currently fails.
- Sometimes the atom index is used even though the cluster index order 1 should be used. Since with StructureManagerCenters there is no difference for 0 layers, tests have to be still added, and some code to be adapted
- Finish all small issues TODO(alex)

### Question parent of ClusterRefKey:
The parent is saved as ClusterRefClusterKey, such that the parent of the parent can only be retrieved if we downcast it to ClusterRefKey. When it is downcasted we have to careful to have the correct layer information, which is the LayerByOrder of the top level manager (the last added manager/adaptor). We could also save the information in ClusterRefKey, but then ClusterRefKey would need the LayerByOrder as template argument. Since I we currently do not have any functionality which requires parents of parents, I would put this as an enhancement issue for later.

### Question reduce code repetition
To solve issue #86 completety, an implementation of the get_property would be needed in each adaptor. To reduce code repetition for the default behaviour (forward to lower level manager if not(trait::HasDistances)) I would like to implement what is proposed in #91.


mastricker:
Hi,

I read your comments and it sounds sensible, but I have not gone over the code yet.

Is there a reason why all build checks fail?
If that is sorted out, I will have a look...

Cheers,
Markus

agoscinski:
I only did checks main_test_suit, will check this

agoscinski:
python bindings should be fixed

mastricker:
There is still something wrong or missing. The binding tests work but the main test suite fails on my computer, and in the CI also.

On my computer it is the property test:
```
1:   Test suite "Property_tests" has failed with:
1:     8 test cases out of 9 passed
1:     1 test case out of 9 failed
1:     1 test case out of 9 aborted
1:     2190 assertions out of 2191 passed
1:     1 assertion out of 2191 failed
```

agoscinski:
The feature is almost fully tested. But I did one major change which deserves already some discussion.

    1. Property and TypedProperty has a template parameter to to declare the MangerImplementation type he gets on construction. Before he only knew that it was a BaseManager. Since a Property only needs to know the information of the manager to know what the clusters neighbour is (look at the atomic index of the cluster and use it in the map atomic index to cluster index of order=1,layer=ManagerLayer), we don't need to save any neighbour in the ClusterRefKey. The Parent of a ClusterRefKey is not used anyway, so we don't need to change ClusterRefKey. So I use the old ClusterRefKey again and changed the Property class instead.
    2. Can I refactor atom_indices to atomic index, so that we mean with atom index a cluster index of order 1 with type size_t like we do it with pair index and triplet index, and atomic index is the unique identifier of the atom with type int which does not change with filtering in higher layers.

mastricker:
Hi,

about 1.:
I do not really follow, sorry. What is it exactly? Is it related to the `[]` operator?

about 2.:
The change sounds reasonable at first glance. Although I am not sure if we should use variable names which are so similar. I am thinking: maybe we should use the same variable as lammps. They are using `inum` for the list of `atom_indices`. This might show the intention for having two very different, but easily confusable _list of atomic indices_. One is an integer name, the other is an actual index inside librascal? Although the former is also used for access in lammps-related properties.

agoscinski:
1. I edited the original comment, hope that it is clearer now. Yes the neighbour is now accessed in the operator[] function of the Property using the manager. https://github.com/cosmo-epfl/librascal/blob/bug/%2384-86/src/structure_managers/property.hh#L204 
2. I am okay with every name as long we use a different name.

felixmusil:
I don't like inum since it is not very clear what it refers to :) 
How about atom_indices -> atom_tags (these correspond to the external indexing, the one from lammps for instance) and the keep atom_indices for the internal indexing ?

mastricker:
Agreed, I like it.

agoscinski:
I have another question, currently if consider_ghost_neighbours=false, in the case of the access of an atomic property with an ghost atom we update the atom in the origin cell corresponding to the ghost atom. Is this something you would also like to have, even if consider_ghost_neighbours=true? So an additional parameter could be implemented.

felixmusil:
I think that the wrapping behavior should always happen.

mastricker:
Before I comment, I have to check what lammps does with that. Because in a parallel setting, a ghost atom might be an atom that is related to another domain (on a different process), not a periodic image of an atom in the current cell.
Hence, the contribution to the ghost atom should be put to the ghost atom and not to the original atom. But then again, in lammps we already have a neighbour list, which is not related to our neighbour list built.

agoscinski:
Please check in code (src and tests) the following tags
#BUG8486@(yourname)
#BUG8486@(all)
for yourname I used: felix, markus, till
I will now stop committing now, until the questions are all resolved. You can answer by commit, write here or by slack.

All tests should pass except of lint/src. If someone could figure out the lint errors. I was not able to find out where to add an extra space.

mastricker:
I'll quite the linter. Get back then.
Thanks!

mastricker:
Ok, so I went especially over the marked things for me, this are my comments. I will add more, once I have gone over them and the rest of the code:

- `property_typed'
I think the solution for the `consider_ghost_neighbour`-dependent resizing is
good. I have no idea right now how to do that otherwise.  I am not sure I
understand the problem you outlined in line 234. What do you mean by it does not
always work?
The C++ documentation says:
Returns the number of elements in the container, i.e. `std::distance(begin(), end())`.
And for `::capacity`:
Returns the number of elements that the container has currently allocated space for.
Why would we want to set unused (but allocated) memory to zero?
- `adaptor_increase_maxorder.hh`: my name is tagged to the comment which states
  that the ghost atoms are appended to the cluster index containter. I do not
  really see how that is different from before?
And line 1036:
That comment is referring to our new tag/index distinction, right?
I agree with the changes.

agoscinski:
@mastricker 
- Okay I agree with you, I had some bug before and I though it was because resize() changes only the capacity and not the size, but I have just tried this in a separate file and they are the same when resize() is used (vector is filled with zeros then). I will change it back to size()
- before increase max order did not have ghost atoms in the cluster indices container of order 1:  this->manager was used before and now I use this->manager->with_ghosts()

src/structure_managers/adaptor_strict.hh
```diff
@@ -239,6 +237,35 @@ namespace rascal {
       return this->manager->get_shared_ptr();
     }
 
+    // BUG8486@(till) I deleted the non const getters, because they are not
```
felixmusil: this is fine. you can delete the comment

felixmusil:
I went though most of the changes and it seems fine to me. How about you Markus ?

Alex, you should add your name to the preamble of the files you modified.

mastricker:
The short things I looked at were fine.
Did Till have a look at the questions posed to him?

Otherwise seems fine.
Cheers,
M

On Jun 2, 2019 22:14, felixmusil <notifications@github.com> wrote:

I went though most of the changes and it seems fine to me. How about you Markus ?

Alex, you should add your name to the preamble of the files you modified.

-

PR #93, opened 2019-06-02
Feat/soap improvements

felixmusil:
I think we added enough lines to warrant a merge. We optimized the spherical expansion and the soap representation quite a bit:
+ implementation of a specialized hypergeometric 1f1 function with an interface dedicated to spherical expansion
+ improved elementary functions like integer power, shape of the spherical harmonics...
+ rework of the spherical expansion to make it modular
+ rework of the computational details of the spherical expansion (precomputations, storage type and computations)
+ rework of the soap representation to make it more modular
+ rework of the computational details of the soap representation (precomputations, storage type and computations)
+ added cutoff functions (cosine and radial scaling)
+ property block sparse performances have been improved (storage, access, resize)
+ file reading capabilities have been fully moved to atomic_structure
+ python interface and hyperparameter definition is converging to the format we discussed 
+ tests for the new corresponding features

One thing we have to investigate: the python binding are orders of magnitude slower than the c++ counterpart. Cross language profiling is a big pain so I hope the RepresentationCalculator will address the issue...

Note that we tried to use Eigen's vectorization infrastructure as much as possible but I am not sure if we can do better...

We are not using cephes anymore and I don't think we will need it... Should I remove it ?

There are some additional improvements that should come afterwards:
+ improve spherical harmonics calculation (precompute terms depending on hyper parameters and polynomials summation)
+ look more into vectorization
+ openmp parallelization 

There are still some things to clean up before merging that I will take care of soon.


ceriottm:
I'd wait for @agoscinski to reshape the Legendre polynomials into a class -- or the whole spherical harmonics calculator into a single class, actually. Precompute as much as possible. Also, I think that if we rid of CEPES we could give more telling names to files in math/

felixmusil:
I would not wait since there is no problems having another pull request. It is better to not keep diverging branches for too long.

Le 3 juin 2019 01:08, Michele Ceriotti <notifications@github.com> a écrit :

I'd wait for @agoscinski<https://github.com/agoscinski> to reshape the Legendre polynomials into a class -- or the whole spherical harmonics calculator into a single class, actually. Precompute as much as possible. Also, I think that if we rid of CEPES we could give more telling names to files in math/

Issue #99, opened 2019-06-24
cpplint complains about blank lines that clang-format doesn't remove


max-veit:
This is especially frustrating as one cpplint complaint makes all the tests on the CI error out, and you'd hope that `make pretty` would at least get rid of this most stupid class of "errors". But no, blank lines at the beginning of code blocks (specifically classes) aren't removed by clang-format, even though they send cpplint into a tizzy.

More generally, I'd like to argue for removing `make lint` for all but one of the CircleCI targets (or make a target just for linting), because otherwise it's hard to see whether the tests failed due to a genuine error or just a quick-fix formatting non-conformance.

ceriottm:
I presume there is no way to instruct clang-format to remove those lines? Alternatively, can't we change the CPPLINT settings so that those lines are ignored, and then have a make lint-pedantic target for those of us who are irked by those white lines? 
I grew to actually like the fact that the code is consistently and cleanly formatted, and I think that having the circleCI complaint when we don't do make lint before pushing is a small price to pay.

mastricker:
Hi,

I see where this frustration comes from. I will have a look, if that can be integrated into the make pretty.
And I tend to agree with Michele.

One way to circumvent this is to do a local `make lint` and just follow the 2-3 errors it throws. I see that this should be automated, but it did not bother me enough to actually do it.

Again, will have a look if it can be integrated.

PR #100, opened 2019-06-28
added empty line removal to clang-format

mastricker:
This is a small one, solving issue #99 

I had to clear my head - dug a little in the clang-format documentation and added the corresponding line into the `.clang-format` file.
I also reformatted the template parameters, because it looked ugly, the linter is happy with `&&` instead of `and`.

Empty lines at the beginning of functions are now removed automatically with `make pretty-cpp`.
I think with this we do not need to take out the linter from the CI.
If other things pop up, we will try to automate them also.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,10 +36,6 @@ jobs:
                         - run:
                                 command: |
                                         cd build
-                                        make lint
-                        - run:
-                                command: |
-                                        cd build
                                         make
                         - run:
                                 command: |
@@ -49,6 +45,10 @@ jobs:
                                 command: |
                                         cd build
                                         make dev_doc
+                        - run:
+                                command: |
+                                        cd build
+                                        make lint
 
         build_ubuntu_g++-6.4:
                 docker:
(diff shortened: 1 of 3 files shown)


max-veit:
Wait wait, hang on. 

> Empty lines at the beginning of functions are now removed automatically with `make pretty-cpp`.
> I think with this we do not need to take out the linter from the CI.
> If other things pop up, we will try to automate them also.

I still disagree, because the linter _masks errors_ that we'd otherwise see right away on the CI. It's impractical to run `make pretty` before every push; such a strategy would also contribute to meaningless commits and diff noise. If nothing else, let's move `make lint` to the end of the build process so we can at least see if the damn thing builds before failing with what is essentially "Wait, this code isn't pretty enough. Build rejected!"

max-veit:
I just committed this reordering of the build process -- this keeps the linter in the CI while at least letting us see the real reason an automated build fails. Please let me know if you're happy with this solution.

mastricker:
It's fine.
At least locally that is what my 'make' command does anyways.

On Jul 2, 2019 17:45, Max Veit <notifications@github.com> wrote:

I just committed this reordering of the build process -- this keeps the linter in the CI while at least letting us see the real reason an automated build fails. Please let me know if you're happy with this solution.

-

Issue #65, opened 2018-12-19
Make JSON schemas to provide a json serialization of StructureManagers and Calculators


max-veit:
Related to #54, but on the JSON end: Write a [JSON schema](https://json-schema.org/specification.html) to specify the format of the JSON data your RepresentationManager can take. Most relevant to us is that you can specify required values, data types (e.g. float), and numeric restrictions (e.g. >= 0).

felixmusil:
I would not make a json schema for the hyperparameters format, a good documentation should suffice. 

Making this possible for serializing entire objects is useful though to be able to dump representations and trained models.
I would leverage the [functionalities](https://github.com/nlohmann/json#arbitrary-types-conversions) of the json library we are using, and since many of our types are only [move constructible](https://github.com/nlohmann/json#how-can-i-use-get-for-non-default-constructiblenon-copyable-types) it won't be as elegant as advertised.

PR #102, opened 2019-07-10
Feat/center select

felixmusil:
Add the selection of central atoms in the manager centers: if you want descriptors for a subset of the unit cell then there is a new parameter allowing to make such decision. This is very useful when using the representations for structural classification.

The main modifications:
+ atomic structure has a is_a_center_atom optional argument that is used in manager centers to decide if an atom should be part of the iteration over centers.
+ json (de)serialization to eigen matrix and array has been added to json_io and is used by atomic structure.
+ the center atoms in manager centers are referred first (first block of atoms_index) and then the no center atoms are referred to. the atom tags are different when some centers are omitted so the order of iteration over neighbors changes.
+ adaptor neighbour list has been modified to fit the changes. atom_tag_list only contains the tags of the atoms that will be iterated over (no center atoms are omitted) and ghost atoms tag start from the number of atoms in the cell so atom_tag_list indices are not continuous if there are some omitted atoms. The omitted atoms properties in atom_types and atom_index_from_atom_tag_list are filled with largest values.
+ I renamed some intermediate variable to make clear the difference between the iteration index provided by the iteration over the manager, the atom_tags which are arbitrary indices that follow atoms and the atom_index that correspond to an index in a list of positions/types... (in the structure_manager_centers at least).  
+ I test this functionality in manager centers and adaptor strict (for both NL and Strict). Do you think this is enough ? do you have other ideas ?

Note that it includes already feat/soap_gradients so I would wait for it to be merged before merging this one.


src/structure_managers/adaptor_neighbour_list.hh
```diff
@@ -555,24 +559,25 @@ namespace rascal {
...
+      auto nb_atoms{this->consider_ghost_neighbours
+                        ? this->n_centers + this->n_ghosts
+                        : this->get_size()};
       return nb_atoms;
     }
 
     //! Returns position of an atom with index atom_tag
     inline Vector_ref get_position(const size_t & atom_tag) {
-      if (atom_tag < n_centers) {
+      if (atom_tag < this->n_atoms) {
```
mastricker: I guess this is needed because of the changed number of center atoms?

src/structure_managers/adaptor_neighbour_list.hh
```diff
@@ -555,24 +559,25 @@ namespace rascal {
...
+        return this->get_ghost_position(atom_tag - this->n_atoms);
       }
     }
 
     //! ghost positions are only available for MaxOrder == 2
-    inline Vector_ref get_ghost_position(const size_t & atom_tag) {
+    inline Vector_ref get_ghost_position(const size_t ghost_atom_index) {
       auto p = this->get_ghost_positions();
-      auto * xval{p.col(atom_tag).data()};
+      auto * xval{p.col(ghost_atom_index).data()};
```
mastricker: Is that just a rename or does it bear any other functionality?

src/structure_managers/adaptor_neighbour_list.hh
```diff
@@ -628,28 +637,30 @@ namespace rascal {
...
      * cell.
      */
-    size_t get_atom_index(const int atom_tag) const {
+    inline size_t get_atom_index(const int atom_tag) const {
       return this->atom_index_from_atom_tag_list[atom_tag];
     }
 
     //! Returns the type of a given atom, given an AtomRef
     inline const int & get_atom_type(const int & atom_tag) const {
+      // return this->atom_types[this->get_atom_index(atom_tag)];
```
mastricker: can this go away? or should this be the version which is actually correct?

src/structure_managers/adaptor_neighbour_list.hh
```diff
@@ -780,6 +790,9 @@ namespace rascal {
      */
     size_t n_ghosts;
 
+    //! number of atoms in the unit cell
```
mastricker: maybe add: "and considered as centers?"

src/structure_managers/adaptor_neighbour_list.hh
```diff
@@ -780,6 +790,9 @@ namespace rascal {
...
+    size_t n_atoms{};
+
     //! counts the number of time the neighbour list has been updated
     size_t n_update{0};
```
mastricker: Is this an artefact? Or do you need that for anything specific in the performance analysis?

src/structure_managers/adaptor_neighbour_list.hh
```diff
@@ -973,8 +988,8 @@ namespace rascal {
     // solve inverse problem for all multipliers
     auto cell_inv{cell.inverse().eval()};
     auto multiplicator{cell_inv * xpos.eval()};
-    auto xmin = multiplicator.rowwise().minCoeff();
-    auto xmax = multiplicator.rowwise().maxCoeff();
+    auto xmin = multiplicator.rowwise().minCoeff().eval();
```
mastricker: Damn, how did that ever work - or did that change with the version of Eigen?

src/structure_managers/adaptor_neighbour_list.hh
```diff
@@ -1006,49 +1021,51 @@ namespace rascal {
     // of current atoms to start the full list of current i-atoms and ghosts
     // This is done before the ghost atom generation, to have them all
     // contiguously at the beginning of the list.
-    int ntot_atoms{0};
-    for (auto atom : *this->manager) {
-      int atom_tag = atom.get_atom_tag();
-      size_t cluster_index = this->manager->get_atom_index(atom_tag);
-      auto atom_type = atom.get_atom_type();
-      this->atom_tag_list.push_back(atom_tag);
+    // for (auto center : this->manager) {
```
mastricker: Still needed? (I mean the commented out section?

src/structure_managers/adaptor_neighbour_list.hh
```diff
@@ -1006,49 +1021,51 @@ namespace rascal {
...
           auto flag_inside =
               internal::position_in_bounds(ghost_min, ghost_max, pos_ghost);
 
           if (flag_inside) {
             // next atom tag is size, since start is at index = 0
-            auto new_atom_tag = this->n_centers + this->n_ghosts;
-            // get_size_with_ghosts();
+            auto new_atom_tag{this->n_atoms + this->n_ghosts};
             this->add_ghost_atom(new_atom_tag, pos_ghost, atom_type);
-            ntot_atoms++;
```
mastricker: Why is not needed anymore? I am a little confused.

src/structure_managers/adaptor_neighbour_list.hh
```diff
@@ -1076,19 +1093,22 @@ namespace rascal {
     // #BUG8486@(markus) I changed it so it could work with other
```
mastricker: So is this solved then?

src/structure_managers/make_structure_manager.hh
```diff
@@ -302,39 +302,45 @@ namespace rascal {
   }
 
   namespace internal {
-    template <typename StructureManagerPtr, int TargetLevel>
+    template <typename StructureManager, int TargetLevel>
```
mastricker: Why is it not a pointer anymore? Or is it a pointer and just not named like it any more?

src/structure_managers/make_structure_manager.hh
```diff
@@ -302,39 +302,45 @@ namespace rascal {
...
+  extract_underlying_manager(std::shared_ptr<StructureManager> manager) {
+    // static_assert(TargetLevel < 0, "target_level should be negative");
+    // using LayerByOrder = typename StructureManager::traits::LayerByOrder;
+    static constexpr int n_step_below =
+        StructureManager::traits::StackLevel - TargetLevel;
+    // static_assert(n_step_below >= 0, "TargetLevel is larger than the number
+    // of manager in the stack");
     auto test{
-        internal::UnderlyingManagerExtractor<StructureManagerPtr, TargetLevel>(
+        internal::UnderlyingManagerExtractor<StructureManager, n_step_below>(
```
mastricker: is the `n_step_below` referring to the new traits with `StackLevel`?

src/structure_managers/structure_manager.hh
```diff
@@ -911,7 +911,7 @@ namespace rascal {
 
     //! return the index of the atom/pair/etc. it is always the last one,
     //! since the other ones are accessed an Order above.
-    inline int get_atom_tag() const { return this->back(); }
+    inline decltype(auto) get_atom_tag() const { return this->back(); }
```
mastricker: Why not keep it an `int`, the lammps manager will point to it.

src/structure_managers/structure_manager_centers.cc
```diff
@@ -37,8 +37,9 @@ namespace rascal {
   /* ---------------------------------------------------------------------- */
   // function for setting the internal data structures
   void StructureManagerCenters::build() {
-    this->natoms = this->atoms_object.positions.size() / traits::Dim;
-
+    auto && is_center_atom = this->get_is_center_atom();
```
mastricker: Again, would rename that to a something like a `_mask`

src/structure_managers/structure_manager_centers.hh
```diff
@@ -271,6 +284,12 @@ namespace rascal {
       return 0;
     }
 
+    inline int get_atom_tag(const int & raw_index) const { return raw_index; }
+
+    inline size_t get_atom_tag(const size_t & raw_index) const {
+      return raw_index;
```
mastricker: Why raw index and why two functions one for size_t and one for int?

src/structure_managers/structure_manager_centers.hh
```diff
@@ -326,7 +347,10 @@ namespace rascal {
      */
     std::vector<size_t> offsets{};
 
-    //! Total number of atoms in structure
+    //! Total number of center atoms in the structure
+    size_t n_center_atoms{};
+
+    //! Total number of center atoms in the structure
```
mastricker: A better comment is needed. This is the same comment as for the `n_center_atoms`, but the above means something different, right? This could change based upon the (by me called `mask`).

src/structure_managers/structure_manager_lammps.cc
```diff
@@ -51,6 +51,7 @@ namespace rascal {
     this->vatom = vatom;
     this->offsets.reserve(inum);
     this->offsets.resize(1);
+    this->atom_index_from_atom_tag_list.clear();
     // #BUG8486@(all) it should be this->inum
```
mastricker: solved?

tests/test_adaptor.hh
```diff
@@ -276,6 +279,140 @@ namespace rascal {
     json factory_args{};
   };
 
+  struct MultipleStructureManagerNLFixtureNoCenter {
```
mastricker: what does `NoCenter` refer to here? `CenterMask` maybe?

tests/test_adaptor_neighbour_list.cc
```diff
@@ -35,6 +35,7 @@
 namespace rascal {
 
   BOOST_AUTO_TEST_SUITE(neighbour_list_adaptor_test);
+  // TODO(felix) add adaptor test for the no center option
```
mastricker: Yes, that needs to be done before merge, otherwise I fear it will be forgotten.

tests/test_adaptor_strict.cc
```diff
@@ -409,6 +411,92 @@ namespace rascal {
...
+
+  BOOST_FIXTURE_TEST_CASE_TEMPLATE(multiple_no_center_test, Fix,
+                                   Fixtures_no_center, Fix) {
+    bool verbose{false};
+    auto & managers = Fix::managers;
+    int n_it{static_cast<int>(managers.size())};
+    for (int i_it{0}; i_it < n_it; i_it += 2) {
+      auto & manager = managers[i_it];
+      auto & manager_no_center = managers[i_it + 1];
+      auto is_center_atom = extract_underlying_manager<0>(manager_no_center)
```
mastricker: While I understand what is done here, it seems a little convoluted. Is there no data structure in the fixture to get this? Or do you specifically want to test for this exact thing?

tests/test_properties.cc
```diff
@@ -731,7 +731,7 @@ namespace rascal {
         sparse_features.emplace_back(*manager, sparse_features_desc);
         this->keys_list.emplace_back();
         test_data_t test_data{};
-        //for (auto atom : manager) { // NOTE(max) replaced since `atom` unused
+        // for (auto atom : manager) { // NOTE(max) replaced since `atom` unused
```
mastricker: I guess this can go away, right?

tests/test_representation_manager.cc
```diff
@@ -32,6 +32,72 @@
...
+      //                       RepresentationManagerSortedCoulomb>,
+      RepresentationFixture<MultipleStructureSphericalExpansionNoCenter,
+                            RepresentationManagerSphericalExpansion>,
+      RepresentationFixture<MultipleStructureSOAPNoCenter,
+                            RepresentationManagerSOAP>>;
+
+  /**
+   * Test that selecting subsets of centers will give the same representation
+   */
+  BOOST_FIXTURE_TEST_CASE_TEMPLATE(multiple_nocenter_test, Fix,
```
mastricker: Again, suggest using something along the lines of `mask` instead of the nocenter

tests/test_structure_manager_centers.cc
```diff
@@ -40,23 +40,31 @@ namespace rascal {
...
-                          atom_cluster.get_position()[ii]);
-      }
-      if (verbose) {
-        std::cout << "atom (" << atom_cluster.back()
-                  << "), atom counter = " << atom_counter << std::endl;
+    int i_manager{0};
+    for (auto & manager : this->managers) {
+      auto & structure = this->structures[i_manager];
+      auto & is_a_center_atom = structure.is_a_center_atom;
+      BOOST_CHECK_EQUAL(manager->get_size(), is_a_center_atom.count());
```
mastricker: I need to think about this a little bit more, `get_size()` is needed for iteration, and you only want to iterate over the _marked_ centers. So it makes total sense here. But on the other hand I still find it unusual. Maybe it is the naming (mask instead of nocenter) that would help me wrap my head around it.

tests/test_structure_manager_centers.cc
```diff
@@ -69,57 +77,72 @@ namespace rascal {
...
-
-      auto pos = atom.get_position();
-      auto pos_reference = positions.col(index);
-      auto position_error = (pos - pos_reference).norm();
-      BOOST_CHECK(position_error < tol / 100);
+    int i_manager{0};
+    for (auto & manager : this->managers) {
+      auto natoms = manager->size();
+      auto natoms2 = manager->get_size();
+      auto natoms3 = manager->get_nb_clusters(1);
```
mastricker: These kind of checks also need to be done for the neighbour list.

bindings/bind_py_structure_manager.cc
```diff
@@ -240,6 +240,20 @@ decltype(auto) add_structure_manager_implementation(py::module & m,
 
 template <typename StructureManagerImplementation>
 void bind_update_unpacked(PyManager<StructureManagerImplementation> & manager) {
+  using ArrayB_t = AtomicStructure<3>::ArrayB_t;
```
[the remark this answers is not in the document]
felixmusil: array of boolean

examples/test_nl.cc
```diff
@@ -69,33 +107,171 @@ int main() {
...
+  //   auto i_idx = uni(rng);
+  //   atomic_structure.is_a_center_atom(i_idx) = false;
+  // }
+  // atomic_structure.is_a_center_atom = false;
+  // // atomic_structure.is_a_center_atom(0) = true;
+  // atomic_structure.is_a_center_atom(4) = true;
+  atomic_structure.is_a_center_atom(2) = false;
+  atomic_structure.is_a_center_atom(3) = false;
+
+  atomic_structure.is_a_center_atom(6) = false;
```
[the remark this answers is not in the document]
felixmusil: this is what it is. What are you suggesting ?

src/atomic_structure.hh
```diff
@@ -255,15 +216,27 @@ namespace rascal {
       return is_similar;
     }
 
+    inline bool is_identical(const PositionsInput_t & positions,
+                             const AtomTypesInput_t & atom_types,
+                             const Eigen::Ref<const Eigen::MatrixXd> cell,
+                             const PBCInput_t & pbc,
+                             const double & skin2) const {
+      auto is_a_center_atom = ArrayB_t::Ones(atom_types.size());
+      return this->is_identical(positions, atom_types, cell, pbc,
```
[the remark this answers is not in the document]
felixmusil: this is in another branch already

src/atomic_structure.hh
```diff
@@ -255,15 +216,27 @@ namespace rascal {
...
+                                is_a_center_atom, skin2);
+    }
+
     inline bool is_identical(const PositionsInput_t & positions,
                              const AtomTypesInput_t & /*atom_types*/,
                              const Eigen::Ref<const Eigen::MatrixXd> cell,
                              const PBCInput_t & pbc,
+                             const ArrayB_ref & is_a_center_atom,
                              const double & skin2) const {
       bool is_similar{true};
```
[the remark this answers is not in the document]
felixmusil: identical in the finite number world is relative to the threshold to me so identical and similar are synonyms in this context

src/json_io.hh
```diff
@@ -35,11 +35,244 @@
...
+        throw std::runtime_error(error);
+      }
+    }
+
+    internal::ResizeEigen<_Rows, _Cols>::apply_arr(n_rows, n_cols, s);
+
+    internal::FillEigen<_Rows, _Cols>::apply_arr(j, s);
+  }
+}  // namespace Eigen
+
```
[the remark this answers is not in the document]
felixmusil: it is tested to make sure the data is the same after a round of conversion

src/representations/representation_manager_soap.hh
```diff
@@ -753,7 +753,7 @@ namespace rascal {
         soap_vector_gradients[center].resize(
             pair_list, n_spatial_dimensions * n_row, n_col);
 
-        //TODO(max,felix) needs work
+        // TODO(max,felix) needs work
```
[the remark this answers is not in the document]
felixmusil: we know what it means

src/structure_managers/adaptor_filter.hh
```diff
@@ -57,6 +57,8 @@ namespace rascal {
     constexpr static bool HasDirectionVectors{
         parent_traits::HasDirectionVectors};
     constexpr static int Dim{parent_traits::Dim};
+    constexpr static int StackLevel{ManagerImplementation::traits::StackLevel +
```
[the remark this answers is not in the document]
felixmusil: the extract_underlying_manager is used to get a manager from the manager stack and StackLevel is used to know how large is the stack and when to stop going down in the stack for extraction.

src/structure_managers/adaptor_neighbour_list.hh
```diff
@@ -300,11 +304,11 @@ namespace rascal {
     //! get dimension dependent neighbour indices (surrounding cell and the cell
     //! itself
     template <size_t Dim, class Container_t>
-    std::vector<int>
-    get_neighbours_atom_tag(const int current_atom_tag,
-                            const std::array<int, Dim> & ccoord,
-                            const Container_t & boxes) {
-      std::vector<int> neighbours_atom_tag{};
+    void get_neighbours_atom_tag(const int current_atom_tag,
```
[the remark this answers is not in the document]
felixmusil: the reference is modified so no need to return. it avoids creating intermediate variables neighbours_atom_tag over and over again

src/structure_managers/adaptor_neighbour_list.hh
```diff
@@ -325,7 +328,7 @@ namespace rascal {
       std::array<int, dimension> nidx{};
       for (auto dim{0}; dim < dimension; ++dim) {
         auto val = position(dim);
-        nidx[dim] = static_cast<int>(std::floor(val / rc));
+        nidx[dim] = std::max(1, static_cast<int>(std::floor(val / rc)));
```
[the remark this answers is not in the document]
felixmusil: yes indeed

src/structure_managers/adaptor_neighbour_list.hh
felixmusil: yes

src/structure_managers/adaptor_neighbour_list.hh
felixmusil: rename

src/structure_managers/adaptor_neighbour_list.hh
felixmusil: this is where I am not sure about the design. here like at other places atom_tag is used to access arrays, these arrays are designed to work like these so no problem should arise but there might be a better way to do it. What do you think ?

src/structure_managers/adaptor_neighbour_list.hh
felixmusil: this is true from n_centers but not for n_atoms

src/structure_managers/adaptor_neighbour_list.hh
felixmusil: I don't follow you. it is a useful member variable

src/structure_managers/adaptor_neighbour_list.hh
felixmusil: yes I ended up with a problem here and needed to eval() the expressions

src/structure_managers/adaptor_neighbour_list.hh
felixmusil: the variable is not needed anymore since n_ghosts does the same job

src/structure_managers/adaptor_neighbour_list.hh
felixmusil: I guess

src/structure_managers/make_structure_manager.hh
felixmusil: std::shared_ptr<StructureManager> is explicitly required

src/structure_managers/make_structure_manager.hh
felixmusil: yes

src/structure_managers/structure_manager.hh
felixmusil: auto is used here because the type returned is dictated by the storage rather than the function

src/structure_managers/structure_manager_centers.hh
felixmusil: atom_tag_list has int and somewhere I don't remember it is used to get_atom_tag so it is there to avoid compilation error

src/structure_managers/structure_manager_lammps.cc
felixmusil: no clue

tests/test_adaptor.hh
felixmusil: it refers to the fact that some centers are omitted in contrast to other similar struct. NoCenter is how we have been calling this functionality since the begining.

tests/test_adaptor_neighbour_list.cc
felixmusil: it is tested in adaptor_strict, forgot to rm the comment

tests/test_adaptor_strict.cc
felixmusil: is_center_atom is accessible only from manager center, so you need to extract it.

examples/test_nl.cc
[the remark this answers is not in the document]
mastricker: `center_mask`, similar to a `strain_mask` in ASE constraints for example. Same bool array, just a different name.

src/structure_managers/structure_manager_centers.cc
```diff
@@ -47,8 +48,17 @@ namespace rascal {
...
     for (size_t id{0}; id < this->natoms; ++id) {
-      this->atoms_index[0].push_back(id);
-      this->offsets.push_back(id);
+      if (center_atoms_mask(id)) {
+        this->atoms_index[0].push_back(id);
+        this->offsets.push_back(id);
+      }
+    }
+
+    for (size_t id{0}; id < this->natoms; ++id) {
```
max-veit: Wait, what? So the references to the central atoms still get stored, but now they're just pushed to the end of the array? And iteration over the centers only iterates over then first `n_centers` atoms, implicitly skipping the atoms that were not selected as centers? Where is that mentioned/documented?

Another issue: This changes the ordering of the atoms from what was passed in, potentially breaking user and developer expectations. This definitely needs to be documented prominently.

src/structure_managers/structure_manager_centers.hh
max-veit: See, this is also inconsistent with the earlier `decltype(auto) get_atom_tag()` change. We urgently need to agree on an interface for these things, otherwise these definitions are going to keep changing and introducing subtle bugs.

src/structure_managers/structure_manager_centers.hh
```diff
@@ -298,6 +317,8 @@ namespace rascal {
       return this->atoms_object;
     }
 
+    int get_n_atoms() const { return this->natoms; }
```
max-veit: and shouldn't this be a `size_t`?

tests/test_adaptor.hh
```diff
@@ -276,6 +279,140 @@ namespace rascal {
...
+        "reference_data/small_molecule.json"};
+    // const std::vector<std::string> filenames{
+    //     "reference_data/CaCrP2O7_mvc-11955_symmetrized.json"};
+    const std::vector<double> cutoffs{{2., 3.}};
+    const std::vector<double> skins{{0., 0.3}};
+
+    json factory_args{};
+  };
+
+  struct MultipleStructureManagerNLStrictFixtureCenterMask {
```
max-veit: These names are getting impossible to decipher; for a while, I thought this was a duplicate of the class above.

tests/test_representation_manager.hh
```diff
@@ -151,6 +151,61 @@ namespace rascal {
                                   {"normalize", true}}};
   };
 
+  struct MultipleStructureSOAPCenterMask
```
max-veit: It seems a bit ridiculous to have to make a new one of these representation classes, with the same boilerplate code, every time we want to try out a feature of the (completely unrelated!) structure manager. I'm not suggesting we do it in this PR, but let's refactor the test code soon; this is getting out of hand.

examples/test_json.cc
```diff
@@ -0,0 +1,116 @@
+/**
+ * file   test_nl.cc
+ *
+ * @author Felix Musil <felix.musil@epfl.ch>
+ *
+ * @date   2018
+ *
+ * @brief Example for Neighbour list
```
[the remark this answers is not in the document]
felixmusil: would you prefer playground rather than test because that is what it is

src/atomic_structure.hh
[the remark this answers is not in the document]
felixmusil: the function body and signature to test if identical would be the same (probably the threshold would be smaller) so that is the same to me. The fact that we use it to choose to rebuild or not the NL is not part of the function but how it is used.

src/json_io.hh
```diff
@@ -35,11 +35,244 @@
...
 using json = nlohmann::json;
 
+/**
+ * Utilities to convert json to Eigen types and the oposite
+ */
+namespace Eigen {
+  namespace internal {
+
+    template <int _Rows, int _Cols>
+    struct ResizeEigen {
```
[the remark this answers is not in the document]
felixmusil: ResizeEigen is called if it is a fixed size matrix and it corresponds to a no op. If it was throwing an error if executed then some checks should be done before hand but that would defeat the purpose of this construct.

src/json_io.hh
```diff
@@ -35,11 +35,244 @@
...
+          s(icol) = j[icol].get<_Scalar>();
+        }
+      }
+    };
+
+  }  // namespace internal
+  /* ---------------------------------------------------------------------- */
+  /**
+   * By convention 1D array are stored as a single vector and not a matrix
+   * with one dimension set to 1.
```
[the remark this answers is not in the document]
felixmusil: yes that is the case. When converting from json a single vector the input matrix input shape will dictate which dimension is to be set to 1.

src/structure_managers/adaptor_neighbour_list.hh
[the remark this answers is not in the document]
felixmusil: The function name is a bit misleading but modifying by reference is a very valid c/c++ construct since it is the easiest way to modify on several variables while one can only return one.
The clear does not free memory but only reset the size of the array so it effectively does not allocate new memory.
RVO might be able to optimize out this but I checked with the profiler and this change corresponded to some performance gain.

src/structure_managers/adaptor_neighbour_list.hh
```diff
@@ -940,15 +954,15 @@ namespace rascal {
 
       // outer mesh, including one layer of emtpy cells
       auto lmesh = std::fabs(mesh_min[i]) + max_coord + 3 * cutoff;
-      int n = std::ceil(lmesh / cutoff);
+      auto n = std::ceil(lmesh / cutoff);
```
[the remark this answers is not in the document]
felixmusil: it is exactly what is intended here: all computation is done with doubles so there is no point to cast into an int to then cast it back to double the next line. So this modification keep the type consistent with double.

src/structure_managers/structure_manager_centers.cc
felixmusil: I added some documentation

examples/test_json.cc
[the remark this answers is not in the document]
max-veit: Yeah, something like `sandbox.cc`; ideally this wouldn't get pushed to master.

src/structure_managers/adaptor_neighbour_list.hh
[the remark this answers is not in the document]
max-veit: Ok then, but you still have an implicit conversion when it's assigned into the array later; I would highly recommend making this conversion explicit (and using `double` instead of `auto` to clarify what is happening here).

src/structure_managers/structure_manager_centers.cc
felixmusil: it does not change the ordering of the input atoms, it removes some from the list that will be iterated over with. Is is clearer ?

src/structure_managers/structure_manager_centers.hh
felixmusil: This is more something that should be fixed at some point since fundamentally the function should not be needed. The fix is not trivial so it will not be fixed here.

src/json_io.hh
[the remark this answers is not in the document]
max-veit: ...what? Then I don't understand the purpose of this construct. Where is the check that ensures this doesn't get called with the wrong sizes on fixed-size matrices?

tests/test_representation_manager.hh
felixmusil: We started by making the tests avoiding such code repetition. In the end it made the tests very hard to understand so I moved to a more verbose organization that avoid repeating the functional code but states all the relevant inputs in 1 structure (avoiding scattering them over).
Here it is repeated but inherit from a different neighbourlist definition.
To avoid repretition the inheritance is templated now.

ceriottm:
"demo" would convey the message, but I agree with Max that to be pulled
into master a demo should be turned into  a set of tests, or of a
well-documented, stable example.

examples/test_json.cc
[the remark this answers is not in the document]
felixmusil: The purpose of this file is to keep track of the development of the code so that when I want to build a new functionality I can start from there (avoiding writing over and over again the same set up code) to see if it compiles and makes at least some sense.
I agree with you that it is meant to be deleted at some point since the development effort will mostly have converged.

src/structure_managers/structure_manager_centers.hh
felixmusil: yes

tests/test_adaptor.hh
felixmusil: Well... I can only agree with you

src/structure_managers/structure_manager_centers.cc
max-veit: No, not really, because this sequence of for-loops makes it look like you're first iterating over the center atoms, then over the atoms that are not marked as centers, but then doing exactly the same thing with them (adding their IDs to the same two arrays). So this has the effect of reordering the arrays so that the IDs of the centers come first, no? Then where is the actual center selection done? Do you only iterate up to `n_center_atoms` and skip the rest of the array?

src/structure_managers/structure_manager_centers.cc
```diff
@@ -62,7 +72,7 @@ namespace rascal {
   // returns the number of cluster at Order=1, which is the number of atoms
   size_t StructureManagerCenters::get_nb_clusters(size_t order) const {
     if (order == 1) {
-      return this->natoms;
+      return this->n_center_atoms;
     } else {
       throw std::string("ERREUR : Order != 1");
```
max-veit: Just noticed the spelling... it reminds me of Molpro's `Global error FEHLER`, if you've ever had the joy of experiencing that message.

src/atomic_structure.hh
```diff
@@ -139,141 +158,137 @@ namespace rascal {
...
+    j = json{{"cell", cell},
+             {"atom_types", s.atom_types},
+             {"pbc", s.pbc},
+             {"positions", s.positions},
+             {"center_atoms_mask", s.center_atoms_mask}};
+  }
+
+  /* ---------------------------------------------------------------------- */
+  template <int Dim>
+  void from_json(const json & j, AtomicStructure<Dim> & s) {
```
max-veit: Is there any documentation on the format the JSON must be in, and what all the fields represent? This is related to #65 but even for the atomic structure format the schema and documentation are missing. I know we started with an ASE-like format, but now that we've changed some names and added capabilities (like `center_atoms_mask`), these need to be written down somewhere other than in the source code.

src/structure_managers/structure_manager_centers.cc
felixmusil: > So this has the effect of reordering the arrays so that the IDs of the centers come first, no?

the value of id is given by the original order provided by the user, the mask on the center orders these indices differently. This reordering does not matter since atoms that are not centers don't belong to the main loop over the manager: the center atoms will come in the order they are provided by the user.
What can happen is that the order the neighbors are iterated over changes depending on which center is selected. I don't see how it can be an issue since representations should not be sensitive to this arbitrary ordering.

>Then where is the actual center selection done?  Do you only iterate up to n_center_atoms and skip the rest of the array?

it is done here by the magic of the order and the fact that the manager size == n_center_atoms.

max-veit: Right, so the iteration over neighbours is guaranteed to happen in the same order _only_ if the atomic symbols, positions, cell, cutoff, and the center mask are all the same. The reason this might be important is e.g. in indexing dense arrays from a representation manager, you might have some expectation as to how these are ordered -- especially if you're processing these outside of librascal. What happens if you get an (atomic) kernel matrix and you want to know which row corresponds to which atom? In the case of center selection, it looks like the order of the selected centers should be preserved, but this should definitely be written down somewhere.

src/atomic_structure.hh
felixmusil: it is the ASE json format. center_atoms_mask can be exported with the ase writer if is it in the arrays of the Atoms object under the 'center_atoms_mask' keyword

src/json_io.hh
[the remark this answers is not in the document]
felixmusil: there are no checks since it is all templated

src/structure_managers/adaptor_neighbour_list.hh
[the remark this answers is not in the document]
felixmusil: >  the change breaks compatibility with code that expected the old behaviour

it is used here only so no code expects the old behaviour.

src/structure_managers/adaptor_neighbour_list.hh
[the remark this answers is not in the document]
felixmusil: mesh_max is an Eigen::Matrix<double, traits::Dim, 1> so no implicit convertion

src/structure_managers/adaptor_neighbour_list.hh
felixmusil: note that xmin is not a double.

src/structure_managers/structure_manager_centers.hh
felixmusil: > Won't removing atom tags from the interface make a lot of things more complicated?

they should not be removed from since they are useful also in the ClusterRefs so don't worry.

src/atomic_structure.hh
max-veit: Added an example JSON file that illustrates this behaviour

src/structure_managers/adaptor_neighbour_list.hh
mastricker: Ok, then we leave it for now. Maybe adding a comment line why we want to keep the design around as a reminder for now?

tests/test_adaptor_strict.cc
mastricker: I see, because afterwards all center atoms are centers and the others are not...

examples/test_json.cc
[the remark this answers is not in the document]
mastricker: I tend to agree with Felix regarding the functionality, but with Max for the design.
My ideal version would be
1. Example folder, where we have 'c++' examples on how to use and iterate - kind of like this folder started
2. A test/sandbox/temp/playground folder which is fair game

Ideally the sandbox-programs would migrate to examples once functionality is added. And I would like to keep the examples simple.
I once worked with a finite element library called 'deal.ii' and I think I mentioned that before. They have a very nice way of combining examples with a tutorial. [Link](https://www.dealii.org/current/doxygen/deal.II/Tutorial.html). So when you want to use their library, you do the first couple of tutorial, get the hang of it and then use one of the more complex examples, which is close to your problem and modify it.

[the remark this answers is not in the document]
agoscinski: I like this idea. If we make clear that the sandbox folder contains unstable examples, it should be fine. If we want to make it super professional in the future, we only keep this folder in a main development branch and ignore it when merging, but I don't think this is necessary for a science library.

src/json_io.hh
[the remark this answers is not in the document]
max-veit: Wait, but that doesn't stop someone from calling `apply_mat()` with the wrong sizes, even if it is templated. For instance, you could call `ResizeEigen<3, 3>::apply_mat(4, 4)` which, by Eigen rules, shouldn't even compile - but here I think it does, and just silently discards the sizes (the 4, 4) passed in. That can easily mask bugs later on.

examples/test_json.cc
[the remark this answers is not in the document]
max-veit: Are you saying that the code doesn't need to be professional because it's a "science library"? On the contrary, because so many people rely on scientific libraries to get accurate results for their research, that's all the more reason to keep the code clean, stable, and professional. And that applies to the examples as well; eventually we'll want to convert all the sandbox code into usable, predictable, well-documented examples (and the library that Markus linked is no different: All their examples serve a purpose and are extensively documented!).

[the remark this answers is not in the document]
max-veit: So do we want to make a `sandbox/` folder with a `README` explaining that it's only for development examples, not tutorial examples?

src/json_io.hh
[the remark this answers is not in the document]
max-veit: Ugh, I'm not sure why the automated build breaks; it's just fine on my machine. One option could be to just call the resize and let Eigen complain if the sizes differ...

src/structure_managers/structure_manager_lammps.cc
mastricker: Ok, let's leave it now.

examples/test_json.cc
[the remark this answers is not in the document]
mastricker: Suggestion:
We revert the examples, which have been working as examples and were commented and intended as such.
But we allow (at least for now) an additional sandbox folder so one can copy an example and adjust it to his/her liking. 
Ideally all functionality is in examples, but we are still too fluent I guess. And we don't have enough examples. I think we will have those examples with whatever we put in the papers. The whole code for evaluation can be in there. And some data along with it stored somewhere else. That is at least how the linked library worked: they started with the very basics, but there are examples where they solve fluid equations for the earth mantel in FEM, which was a joint publication at some point.

And I agree with Max with the science library. I have worked too long on varieties of this program and that script which was from a science background.

examples/test_json.cc
```diff
@@ -0,0 +1,116 @@
...
+// constexpr auto sum_seq(std::index_sequence<I...>, const int& sum = 0) {
+//   return std::make_tuple(a[I]...);
+// }
+
+// template<size_t I, size_t... Is,size_t Sum = 0 >
+// constexpr auto sum(const int& sum) {
+//   return sum + 1 + I;
+// }
+
+int main() {
```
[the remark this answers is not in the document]
mastricker: This should be resolved in the documentation upgrade branch?

src/json_io.hh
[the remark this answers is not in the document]
mastricker: Is this solved?

src/structure_managers/structure_manager_centers.cc
mastricker: Is this solved?

mastricker:
Still missing the atomic number/element type in the 'is similar' in `atomic_structure.hh`. But I will give it a go, pending the addition of the is-similar.

examples/test_json.cc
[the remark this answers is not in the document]
max-veit: Well no, I resolved it already in 82466e1. All new code must be properly documented, and I'm inclined to start strictly applying that policy from this PR onwards.

src/json_io.hh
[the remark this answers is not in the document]
max-veit: Not quite, not for the case of matrices where both the row and the column size are fixed. (it was sneakily reverted in the merge commit a0aac28...) But for the other two cases, it will call Eigen's resize directly which raises (I think) a run-time error; checking at compile time seems to be either hard or impossible.

src/structure_managers/structure_manager_centers.cc
max-veit: Not sure, I think the implications of the reordering still need to be documented.

felixmusil:
Let's wait for the representation_calculator branch before merging this guy.

src/json_io.hh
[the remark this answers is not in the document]
felixmusil: of course I reverted these asserts since they cause failures that are not warranted. Please check your modifications before committing even if you feel that they should work.

PR #122, opened 2019-09-18
Feat/ipynb hook

rosecers:
PR to merge pre-commit hooks into master branch. 

Current functionality: 
- strips ipython notebooks of output before committing to the repo.
- exits if nbstripout not installed

Known issues:
- requires user to enact ``git config core.hooksPath shared_hooks'' to point to this folder for git hooks.


Luthaf:
> requires user to enact ``git config core.hooksPath shared_hooks'' to point to this folder for git hooks.

This could be made into a script that we run automatically with cmake. This way, if you build the code (which should be done by all contributors :smiley:), then you have the hook installed.

max-veit:
Looks good, works as expected (though I haven't tested what happens when you try to commit without `nbstripout` installed). One comment: There is still some diff noise that happens when you run a notebook with a different python version other than the one, see e.g. this sample edit:
```diff
diff --git a/examples/coulomb_example.ipynb b/examples/coulomb_example.ipynb
index b8ad9753..bf56f39d 100644
--- a/examples/coulomb_example.ipynb
+++ b/examples/coulomb_example.ipynb
@@ -26,6 +26,15 @@
     "+ make install -j 4"
    ]
   },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "print(\"bla bla bla mad wombat\")"
+   ]
+  },
   {
    "cell_type": "code",
    "execution_count": null,
@@ -208,9 +217,9 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "py36",
+   "display_name": "Python 3",
    "language": "python",
-   "name": "py36"
+   "name": "python3"
   },
   "language_info": {
    "codemirror_mode": {
@@ -222,7 +231,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.6.8"
+   "version": "3.7.3"
   },
   "toc": {
    "base_numbering": 1,
```
But second, and more worryingly, it seems to clear the output in _my local working copy_, which is something I never want to happen (what if I'm working on a notebook with something that took a long time to generate, then I decide to commit my intermediate progress? Then the output gets stripped out and the notebook changes right under me, possibly making me lose my results). Therefore, I think we should implement this as a _filter_ in the `.gitattributes` file instead of as a pre-commit hook. This is, for instance the installation method recommended in `nbstripout` itself: https://github.com/kynan/nbstripout#usage And if it slows things down unacceptably, we can always try this alternate implementation: https://github.com/fastai/fastai-nbstripout

And finally, we just need to think about a server-side hook to prevent people from pushing notebooks with output. But it can't be too strict, because I want us to be able to use "pre-filled" notebooks as ready-made tutorials (see e.g. [here](https://github.com/cosmo-epfl/librascal/blob/feat/sphinx_merge_doc/docs/source/tutorials/tutorials.rst) where an IPython notebook is used as a ready-made html tutorial). So we should be able to commit a "final version" of a notebook with output, so it renders properly. What do you all think?

ceriottm:
yeah, the solution I had found adds a .gitattribute file containing
```
*.ipynb filter=nbstripout
*.ipynb diff=ipynb
```
and adds to .git/config
```
[filter "nbstripout"]
    clean = \"/usr/bin/python3\"
\"/home/michele/local/lib/python3.6/site-packages/nbstripout\"
    smudge = cat
    required = true
[diff "ipynb"]
    textconv = \"/usr/bin/python3\"
\"/home/michele/local/lib/python3.6/site-packages/nbstripout\" -t
```
IDK how to get this enforced on the users though.

max-veit:
There's no way to force someone to adopt a repository's git configuration, as that would be a security risk. We could either go with Guillaume's `cmake` idea or a server-side hook that just rejects commits that haven't been filtered and stripped.

ceriottm:
If we can have a server-side hook that gives a meaningful error message
then I'm on board.
M

max-veit:
https://git-scm.com/book/en/v2/Customizing-Git-An-Example-Git-Enforced-Policy#_an_example_git_enforced_policy

Luthaf:
> https://git-scm.com/book/en/v2/Customizing-Git-An-Example-Git-Enforced-Policy#_an_example_git_enforced_policy

Is this possible on Github? It looks like it is doable on the enterprise version, but I have never heard of it (and could not find any information) for `github.com`.

max-veit:
Damn, you're right, looks like it's an enterprise-only thing (you have to have control of the server, as you said).

Our strategy then will be to warn all current (and potential) contributors about the policy (post it prominently in the README!), and as a last resort include a check in the CI, in much the same vein as `make lint`, to prevent ipynb output diffs from being committed (except, of course, as final, polished outputs for tutorials). If it turns out there are big output diffs in a branch's history then we might have to request a rebase to strip them out.

max-veit:
Okay, just to get this closed and merged in: The only thing left to do on this PR is to move the filter to a `.gitattributes` (so it doesn't clear the local working copy, only the staged version). I'm more or less fine with the remaining diff noise as long as the output gets cleared. As for the server-side hook, that's not possible at the moment, so we'll just have to add it as a CI target (on a separate branch).

mastricker:
What is missing here? The filter move to the `.gitattributes`?

max-veit:
Yes, as stated above

ceriottm:
Also make sure we add prominently to the README and the dev guide the
instructions to enable nbstripout.

max-veit:
And add a `CONTRIBUTING.rst` so everyone knows to run this before even submitting a PR.

These two tasks together should make about a 15-minute job; I just haven't gotten the chance to do it yet. If anyone else has the time, please feel free to finish this off.

Luthaf:
>  Is there a way of asking in the script whether any ipynb files have been added or changed and

Which script? 

It should be possible to do something like this by using `git diff --name-only | grep "ipynb"`

max-veit:
`shared_hooks/pre-commit`. Which also isn't run unless explicitly activated by the user, but perhaps we can keep it as an additional safety mechanism.

PR #123, opened 2019-09-19
Misc. spelling, renaming, and documentation

max-veit:
Still WIP, since I don't (yet) have a way of spellchecking the entire codebase. But error messages should get the highest priority, since that's what users will see in their most frustrated and confused moments.

### DIFF
--- a/bindings/bind_py_models.cc
+++ b/bindings/bind_py_models.cc
@@ -29,7 +29,7 @@
 
 namespace rascal {
 
-  decltype(auto) add_kernel(py::module & mod, py::module &) {
+  decltype(auto) add_kernel(py::module & mod, py::module & /*m_internal*/) {
     py::class_<Kernel> kernel(mod, "Kernel");
     // use custom constructor to pass json formated string as initializer
     // an alternative would be to convert python dict to json internally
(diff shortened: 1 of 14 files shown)


bindings/bind_py_models.cc
```diff
@@ -29,7 +29,7 @@
 
 namespace rascal {
 
-  decltype(auto) add_kernel(py::module & mod, py::module &) {
+  decltype(auto) add_kernel(py::module & mod, py::module & /*m_internal*/) {
```
Luthaf: Is `decltype(auto)` needed here? From my understanding, it should be equivalent to `auto` in this case.

max-veit: Not related to these changes, but: I'm pretty sure it's necessary for function return types, since you need to _declare_ the type that the function is returning: https://en.cppreference.com/w/cpp/language/auto

Luthaf: This is completely a drive-by comment, feel free to ignore it here =)

But looking at the examples in the documentation link, you only need decltype(auto) to do perfect forwarding in a completely generic context, which is not the case here. See the first example in your link:

```cpp
template<class T, class U>
auto add(T t, U u) { return t + u; } // the return type is the type of operator+(T, U)
```

felixmusil: the decltype is not necessary here but it does not hurt either. Most of the decltype in the code are like that by the way.

Luthaf: It does hurt in that it will confuse both people that know what decltype is doing, and people how don't, but already have some knowledge of auto

max-veit: Ok, now I'm more confused than ever. But now I agree that it really should be just `auto`, since that takes the type of `kernel` (since the return statement is just `return kernel;`), which is a plain `py::class_<Kernel>` (which IMO would have been much clearer to just write, less mental pointer-chasing to find out what this thing actually returns).

max-veit:
Merged into #124

PR #124, opened 2019-09-19
Feat/representation calculator

felixmusil:
some additional changes to the branch that were missed


max-veit:
Have a look at #123 which implements these changes already. I think it would be best to merge these changes into that PR (branch `feat/spellcheck`) since some of the doc here is more complete (or correct) than what's written in that branch.

PR #130, opened 2019-09-26
Property request forwarding and default manager function

agoscinski:
Short summary
- default implementation for all manager implementations dependent on traits
- new traits "IsRoot" is true for all root implementations as StructureManagerCenters
- added property forwarding
- to validate a property the function get_property_ptr/ref with a true flag is used signature: get_property_ptr/ref(string property_name, bool validate)

Adds forwarding of property requests. Here an example what problem this feature solves:

For this kind of manager stack:
```
AdaptorImpl2 with prop2
AdaptorImpl1 with prop1
RootImpl
```

Problem with get_property_ref/ptr
```
AdaptorImpl2 with prop2 -> invoke get_property_ref/ptr("prop1") -> Adaptor2 does not have prop1 and returns error "No property of name prop1" 
AdaptorImpl1 with prop1
RootImpl
```

Fix
```
AdaptorImpl2 with prop2 -> invoke get_property_ref/ptr("prop1") -> does not hav prop1
      | forwards request to lower stack 
      v
AdaptorImpl1 with prop1 -> return prop1
RootImpl
```

We have two different cases: 
- at the root level if the property is not found an error "No prop found" should be returned
- but for adaptors it has to be forwarded to previous managers

To prevent code repetition in all manager implementations, I implemented this fix with the "default functions dependent on traits" feature for the managers. This is explained in detail in the new example file in the examples folder `examples/example_traits_dependent_function.cc`. The main result is that the `get_distance` and `get_direction_vector` functions are now only in the structure manager present.

There exists now only get_property_ptr/ref as getter for properties which has two boolean flags:
- validate_property, if property found the property is validated
- force creation, creates the property at the top layer if not found (optional, by default false)


src/structure_managers/structure_manager.hh
```diff
@@ -419,77 +469,74 @@ namespace rascal {
...
+      error << "No property of name '" << name << "' has been registered";
+      throw std::runtime_error(error.str());
     }
 
     template <typename UserProperty_t>
-    UserProperty_t & get_property_ref(const std::string & name) {
-      return *this->template get_property_ptr<UserProperty_t>(name);
+    UserProperty_t & get_property_ref(const std::string & name,
+                                      bool validate_property,
+                                      bool force_creation) {
```
felixmusil: I would make it optional.
```suggestion
                                      bool force_creation = false) {
```

src/structure_managers/structure_manager.hh
```diff
@@ -580,7 +627,87 @@ namespace rascal {
...
+    get_direction_vector(const ClusterRefKey<Order, Layer> & pair) {
+      return this->get_previous_manager()->get_direction_vector(pair);
+    }
+
+    template <size_t Order, size_t Layer,
+              bool HasDirectionVectors = traits::HasDirectionVectors,
+              typename std::enable_if_t<not(HasDirectionVectors), int> = 0>
+    inline const Vector_ref
+    get_direction_vector(const ClusterRefKey<Order, Layer> &) const {
+      throw std::runtime_error("Trying to get direction vectors from a manager "
```
felixmusil: Is it better to throw a runtime error or a compilation error ? I would say compilation by not implementing this function even if the compilation error is not so straightforward it is still understandable and avoids long compilations

src/models/kernels.hh
```diff
@@ -80,13 +80,11 @@ namespace rascal {
         size_t ii_A{0};
         for (auto & manager_a : managers_a) {
           size_t ii_B{0};
-          auto && propA{
-              manager_a->template get_validated_property_ref<Property_t>(
-                  representation_name)};
+          auto && propA{manager_a->template get_property_ref<Property_t>(
+              representation_name, true, true)};
```
felixmusil: There are a few places where you force_creation and you shouldn't. here for instance we build a kernel and if one forgot to compute the representation I think it is more informative to fail making sure the user is not surprised to get a kernel filled with zeros.

src/structure_managers/structure_manager.hh
```diff
@@ -419,77 +469,74 @@ namespace rascal {
...
+      std::stringstream error{};
+      error << "No property of name '" << name << "' has been registered";
+      throw std::runtime_error(error.str());
     }
 
     template <typename UserProperty_t>
-    UserProperty_t & get_property_ref(const std::string & name) {
-      return *this->template get_property_ptr<UserProperty_t>(name);
+    UserProperty_t & get_property_ref(const std::string & name,
+                                      bool validate_property,
```
felixmusil: make it optional
```suggestion
                                      bool validate_property = true,
```

max-veit:
An idea to address the diff noise: Try rebasing onto the latest version of master. You'll still have lots of conflicts to fix, but it'll get rid of the stuff that was changed in master and also incorporated here.

You'll need to force-push if you do this, which can break previous comments - YMMV.

src/structure_managers/structure_manager.hh
agoscinski: force creation and validate property is now optional, and by default force creation is false and validate is true

src/structure_managers/structure_manager.hh
agoscinski: changed to static_assert, so the compiler error is more understandable

src/models/kernels.hh
agoscinski: I changed the ones in kernels to not forced, but in the representation calculator all are still forced, is this okay like this?

src/structure_managers/structure_manager.hh
agoscinski: optional now see https://github.com/cosmo-epfl/librascal/pull/130#discussion_r336030905

src/structure_managers/structure_manager.hh
agoscinski: had to remove these functions, because they are always compiled if HasDistance is false, even when they are not used anywhere

docs/source/dev_guide/forwarding_of_property_requests.rst
```diff
@@ -0,0 +1,12 @@
+.. _forwarding_of_property_request:
+
+.. contents::
+   :local:
+
+Forwarding of property requests
+-------------------------------
+
+This c++ file is a simplification of our CRTP structure to show how the forwarding of requests through the manager stack work. The forwarding mechanism is used when getting a property with the `get_property_ptr` or `get_property_ref` function and for the `get_distance` and `get_direction_vector` function.
```
felixmusil:
```suggestion
This c++ file is a simplification of our CRTP structure to show how the forwarding of requests through the manager stack works. The forwarding mechanism is used when getting a property with the `get_property_ptr` or `get_property_ref` function and for the `get_distance` and `get_direction_vector` function.
```

src/rascal/structure_managers/adaptor_full_neighbour_list.hh
```diff
@@ -63,6 +63,7 @@ namespace rascal {
         typename LayerIncreaser<MaxOrder,
                                 typename parent_traits::LayerByOrder>::type{})};
     using LayerByOrder = std::index_sequence<AtomLayer, 0>;
+    typedef ManagerImplementation PreviousManager_t;
```
felixmusil: I would prefer the `using` syntax compared to the typedef. There are other instances too that should be changed

src/rascal/structure_managers/structure_manager.hh
```diff
@@ -181,6 +183,9 @@ namespace rascal {
         typename internal::ClusterIndexConstructor<ClusterIndex_t,
                                                    StructureManager_t>;
 
+    constexpr static bool IsRootImplementation =
```
felixmusil: please comment here the rational for future reference

src/rascal/structure_managers/structure_manager.hh
```diff
@@ -320,72 +325,114 @@ namespace rascal {
...
-    void attach_property(const std::string & name,
-                         std::shared_ptr<PropertyBase> property) {
-      if (this->has_property(name)) {
-        std::stringstream error{};
-        error << "A property of name '" << name
-              << "' has already been registered";
-        throw std::runtime_error(error.str());
-      }
-      this->properties[name] = property;
+    inline bool has_layer_property(const std::string & name) const {
```
felixmusil: I would not make this function as it seems redundant and you could use `count` instead of find to search for the name as I feel it is more readable.

src/rascal/structure_managers/structure_manager.hh
```diff
@@ -320,72 +325,114 @@ namespace rascal {
...
-      return not(this->properties.find(name) == this->properties.end());
-    }
-
-    template <typename Property_t>
-    void create_property(const std::string & name) {
-      auto property{std::make_shared<Property_t>(this->implementation())};
-      this->attach_property(name, property);
+    template <bool IsRoot = IsRootImplementation,
+              std::enable_if_t<IsRoot, int> = 0>
+    inline bool has_stack_property(const std::string & name) {
```
felixmusil: Maybe `has_property_in_current_stack` ?

src/rascal/structure_managers/structure_manager.hh
```diff
@@ -400,88 +447,162 @@ namespace rascal {
...
         auto property{std::make_shared<UserProperty_t>(
             this->implementation(), metadata, exclude_ghosts)};
-        this->register_property(property, name);
+        this->properties[name] = property;
         return property;
       }
+      std::stringstream error{};
+      error << "No property of name '" << name << "' has been registered";
+      throw std::runtime_error(error.str());
     }
```
felixmusil: the logic seems a bit funky here. Also I would set `const bool force_creation = true` as default since it would be the case most of the times. 
How about:
```
if (this->has_stack_property(name)) {
...
} else if (not this->has_stack_property(name) and force_creation) { 
...
} else {
      std::stringstream error{};
      error << "No property of name '" << name << "' has been registered";
      throw std::runtime_error(error.str());
}
```

src/rascal/structure_managers/structure_manager.hh
```diff
@@ -400,88 +447,162 @@ namespace rascal {
...
 
-    template <typename T, size_t Order, Dim_t NbRow = 1, Dim_t NbCol = 1>
-    std::shared_ptr<Property_t<T, Order, NbRow, NbCol>>
-    get_property(const std::string & name) const {
-      return this
-          ->template get_validated_property<Property_t<T, Order, NbRow, NbCol>>(
-              name);
+    /**
+     * For MD when properties needs to be updated
```
felixmusil:
```suggestion
     * to keep track if the property is up to date with the structure
```

src/rascal/structure_managers/structure_manager.hh
```diff
@@ -400,88 +447,162 @@ namespace rascal {
...
+        this->properties[name]->set_updated_status(is_updated);
+        return;
+      }
+      std::stringstream error{};
+      error << "A property of name '" << name << "' does not exist"
+            << " in manager '" << this->get_name() << "' on top layer";
+      throw std::runtime_error(error.str());
+    }
+
+    /*
```
felixmusil:
```suggestion
    /**
```

src/rascal/structure_managers/structure_manager.hh
```diff
@@ -400,88 +447,162 @@ namespace rascal {
...
+        auto property = this->properties.at(name);
+        if (validate_property) {
+          this->template validate_property_t<UserProperty_t>(property);
+        }
+        return std::static_pointer_cast<UserProperty_t>(property);
+      }
+      std::stringstream error{};
+      error << "A property of name '" << name << "' does not exist"
+            << " in manager '" << this->get_name() << "'";
+      throw std::runtime_error(error.str());
```
felixmusil: put that block in an `else` please

felixmusil: is this function meant to be used by user ? if not please move it to the protected section

src/rascal/structure_managers/structure_manager.hh
```diff
@@ -400,88 +447,162 @@ namespace rascal {
...
+          ->template forward_get_property_request<UserProperty_t>(
+              name, validate_property);
+    }
+
+    template <size_t Order, size_t Layer,
+              bool HasDistances = traits::HasDistances,
+              typename std::enable_if_t<HasDistances, int> = 0>
+    inline double & get_distance(const ClusterRefKey<Order, Layer> & pair) {
+      static_assert(HasDistances == traits::HasDistances,
+                    "HasDistances is used for SFINAE, please don't specify it");
```
felixmusil:
```suggestion
                    "The manager does not have distances.");
```

src/rascal/structure_managers/structure_manager.hh
```diff
@@ -400,88 +447,162 @@ namespace rascal {
...
+    }
+
+    template <size_t Order, size_t Layer,
+              bool HasDirectionVectors = traits::HasDirectionVectors,
+              typename std::enable_if_t<HasDirectionVectors, int> = 0>
+    inline const Vector_ref
+    get_direction_vector(const ClusterRefKey<Order, Layer> & pair) {
+      static_assert(
+          HasDirectionVectors == traits::HasDirectionVectors,
+          "HasDirectionVectors is used for SFINAE, please don't specify it");
```
felixmusil:
```suggestion
          "The manager does not have direction vectors.");
```

src/rascal/structure_managers/structure_manager.hh
agoscinski: To be clear
AdatporNL < StructureManagerCenter > is the whole stack
AdatporNL is a layer and StructureManagerCenter is a layer.

If you ask for a property, it has first to check if the property exists layer by layer, therefore I don't see any redundancy in having the two functions `has_layer_property` and `has_stack_property`.

src/rascal/structure_managers/structure_manager.hh
agoscinski: To be clear
AdatporNL < StructureManagerCenter > is the whole stack
AdatporNL is a layer and StructureManagerCenter is a layer.

So `has_stack_property` looks in the whole stack not just the top layer. I could rename it to `has_whole_stack_property`

src/rascal/structure_managers/structure_manager.hh
agoscinski: I don't see the benefit in readability.

src/rascal/structure_managers/structure_manager.hh
agoscinski: Again I don't see the benefit in readability.

agoscinski: it is not meant to be used by the user, but by other adaptors, therefore it needs to be public
AdaptorNL< StructureManagerCenter >
AdatporNL invokes `forward_get_property_request` from StructureManagerCenter https://github.com/cosmo-epfl/librascal/blob/d8dc2fe43a2cee5d276e940dfa041cecd9e1072b/src/rascal/structure_managers/structure_manager.hh#L583-L585

src/rascal/structure_managers/structure_manager.hh
felixmusil: `is_property_in_current_level`

src/rascal/structure_managers/structure_manager.hh
felixmusil: `is_property_in_stack`

PR #131, opened 2019-09-27
Bugfix/paths

ceriottm:
Just a small CMake change to allow me compile from a symlinked folder

### DIFF
--- a/bindings/bind_py_models.cc
+++ b/bindings/bind_py_models.cc
@@ -62,16 +62,20 @@ namespace rascal {
    */
   void add_kernels(py::module & mod, py::module & m_internal) {
     // Defines a particular structure manager type
-    using ManagerCollection_t =
+    using ManagerCollection_1_t =
         ManagerCollection<StructureManagerCenters, AdaptorNeighbourList,
                           AdaptorStrict>;
-
+    using ManagerCollection_2_t =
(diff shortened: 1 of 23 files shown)


Luthaf:
The fix is strange. What was your error ?

ceriottm:
was getting an error about it not finding the build folder, because my
source folder is within a symlink, and there was a clash between the path
with resolved link and path which is behind the link. apparently this is
because it then sees the folder as external to the build tree, and then
needs explicit declaration of the build path.

Luthaf:
I am not following you here. What command are you running exactly, and what is the corresponding error? There should be no need to declare the full build path to cmake as `add_subdirectory`, which is used to declare a *source* path. But there might be an issue with the configuration of pybind11 or something else.

For your folders, you say you have `/path/to/source => /some/other/path` and then `/path/to/source/build`, right?

Luthaf:
Ok, I understand the issue now. In the `add_subdirectory(${_pybind11_external_dir}/pybind11)` call, `${_pybind11_external_dir}` is an absolute path, which makes `add_subdirectory` use the same absolute path as build directory. Then when changing the current path (through the symbolic link @ceriottm is using), the build path does not correspond anymore and cmake bails out.

A cleaner fix would be to pass a relative path to `add_subdirectory`, I'll look into this.

ceriottm:
yeah I think that would be ideal.

Issue #126, opened 2019-09-24
Segfault when atoms are outside the periodic unit cell


max-veit:
Many of you know what I'm talking about; this has been considered a "feature" of librascal for quite a while. But I'm marking this as a bug because a user should _never_ see a segfault for what is considered perfectly valid input for (and often, output from) a number of MD programs. (in fact, they should never see a segfault _ever_, but this is one of the most egregious examples).

If doing a simple bounds-check of the atoms before calculating the neighbourlist / applying adaptors / whatever is too performance-intensive (which I simply don't believe, especially given the tiny fraction of the total compute time it takes to build the neighbourlist in the first place), can we at least add a check in the Python interface? Something with `ase.wrap()` or whatever before the atoms are converted to `librascal` objects?

felixmusil:
I am assigning this task to you @max-veit. The wrapping of the atoms in the unit cell is simple to integrate in AtomicStructure class I think.
The segfault on very skewed cell is more tricky to understand though since it is probably a limiting case that is not properly handdled in the AdaptorNeighbourlist class.

max-veit:
Ugh. Is there an example of such a skewed cell somewhere for testing? I understand why it wouldn't be in the normal test suite (Boost can catch exceptions, but nothing can catch a segfault), but it would be good to have such an example available at least.

felixmusil:
I sent you some structures that should reproduce the segfault because it is skewed

ceriottm:
skewed cells were tested extensively by @mastricker and I believe a test should still be present somewhere. I'd be much in favor to have some basic sanity checks implemented in the release mode, possibly with a precompiler switch to turn them off if someone wants to squeeze the last flop

felixmusil:
Yes that is why it took me some time to pin point the problem.
It is not only about skewness but also triclinicity. I am talking about cells that are almost skewed but but not enough to be easily changed. The solution I found that was working has been to use spglib to sanitize the cell, the transformation is quite involved to streamline.
Since it has a c++ and python interface we can always integrate it to our project ?



Le 24 sept. 2019 22:15, Michele Ceriotti <notifications@github.com> a écrit :

skewed cells were tested extensively by @mastricker<https://github.com/mastricker> and I believe a test should still be present somewhere. I'd be much in favor to have some basic sanity checks implemented in the release mode, possibly with a precompiler switch to turn them off if someone wants to squeeze the last flop

felixmusil:
this is solved in #144

PR #132, opened 2019-09-27
branch to investigate problems in AdaptorNeighborlist

felixmusil:
We get an segfault in AdaptorNeighborlist with a seemingly sane structure.

The error is reproduced in playground.cc if filename == "reference_data/alloy.json" and Michele made it work in "reference_data/alloy-fixed.json" by wrapping and shifting the atoms by the center of mass:
```
frame = read("../tests/reference_data/alloy.json","0")
frame.wrap(center=(0., 0., 0.))
write("../tests/reference_data/alloy-fixed.json",frame)
```

Basically with  "reference_data/alloy.json" when binning the j-atoms in the boxes some atoms end up outside of the bounds leading to a segfault.

It seems from Michele's investigations that alloy-fixed.json is still interpreted properly since the soap vectors are very (too) different from the ones of quippy.


max-veit:
see also #126 for problems with "insane" structures (i.e. the ones that violate the -- still unwritten -- assumptions behind this linked-cell algorithm)

ceriottm:
yeah, I know that @max-veit raised a red flag already quite some time ago - I kind of thought it was a corner-case issue, so I thought it was not priority, but given that it seems this crashes (or worse, gives wrong results) when one passes a simple replicated FCC lattice, I think it needs maximum attention and a quick, robust fix - to be followed up with less urgency with a more efficient implementation, possibly relying on sanitizing the simulation box and/or folding the atoms

mastricker:
I'll have a go at it once I'm back on Friday.

My understanding was the same: corner case an we leave it for the future. Seems more pressing now.

This branch has everything that is needed to replicate the issue?

On Sep 29, 2019 17:36, Michele Ceriotti <notifications@github.com> wrote:

yeah, I know that @max-veit<https://github.com/max-veit> raised a red flag already quite some time ago - I kind of thought it was a corner-case issue, so I thought it was not priority, but given that it seems this crashes (or worse, gives wrong results) when one passes a simple replicated FCC lattice, I think it needs maximum attention and a quick, robust fix - to be followed up with less urgency with a more efficient implementation, possibly relying on sanitizing the simulation box and/or folding the atoms

-

examples/linked_cells.ipynb
```diff
@@ -68,38 +62,42 @@
...
   {
    "cell_type": "code",
-   "execution_count": null,
-   "metadata": {
-    "hidden": true
-   },
-   "outputs": [],
+   "execution_count": 14,
+   "metadata": {},
+   "outputs": [
```
max-veit: errrrghhh, please install the notebook cleaner hook, we really don't need another series of outputs recorded in history forever

mastricker:
Is this solved now?

On Oct 3, 2019 20:42, felixmusil <notifications@github.com> wrote:

Closed #132<https://github.com/cosmo-epfl/librascal/pull/132>.

-

PR #134, opened 2019-09-27
enable RadialScaling to the list of possible cutoff

felixmusil:
Adds RadialScaling in spherical expansion to the list of options and in the bindings.


bindings/rascal/representations/base.py
```diff
@@ -26,3 +26,49 @@ def CalculatorFactory(rep_options):
              'The available combinations are: {}').format(
                 name, list(_representations.keys())))
     return _representations[name](*rep_options['args'])
+
+def cutoff_function_dict_switch(cutoff_function_type, **kwargs):
+    """
+    return appropriate dict for the cutoff function parameters
+    """
+    if cutoff_function_type == 'Cosine':
```
max-veit: Let's call this 'ShiftedCosine' for compatibility with B-P

bindings/rascal/representations/spherical_covariants.py
```diff
@@ -81,17 +82,13 @@ class documentation
...
-            cutoff=dict(
-                value=interaction_cutoff,
-                unit='AA'
-            ),
-            smooth_width=dict(
-                value=cutoff_smooth_width,
-                unit='AA'
-            ),
+        cutoff_function_parameters.update(
+            **dict(interaction_cutoff=interaction_cutoff,
```
max-veit: I think the `**dict` syntax is unnecessary here -- it first gathers the keywords into a dict, then explodes them back out into keywords. If I'm not mistaken, you should be able to just remove `**dict([...])` and keep everything inside, to the same effect.

bindings/rascal/representations/spherical_expansion.py
```diff
@@ -66,17 +67,13 @@ class documentation
...
-            cutoff=dict(
-                value=interaction_cutoff,
-                unit='A'
-            ),
-            smooth_width=dict(
-                value=cutoff_smooth_width,
-                unit='A'
-            ),
+        cutoff_function_parameters.update(
+            **dict(interaction_cutoff=interaction_cutoff,
```
max-veit: see my `**dict` comment above

bindings/rascal/representations/spherical_covariants.py
max-veit: Also, is it possible to just pass these parameters along to SphericalExpansion and let the parsing be done once there, rather than three times (once in each of the Spherical* representations)?

bindings/rascal/representations/spherical_invariants.py
```diff
@@ -80,17 +81,13 @@ class documentation
...
-            cutoff=dict(
-                value=interaction_cutoff,
-                unit='AA'
-            ),
-            smooth_width=dict(
-                value=cutoff_smooth_width,
-                unit='AA'
-            ),
+        cutoff_function_parameters.update(
+            **dict(interaction_cutoff=interaction_cutoff,
```
max-veit: same here

src/representations/cutoff_functions.hh
```diff
@@ -123,6 +123,10 @@ namespace rascal {
      *        |
      *        ╰ c / (c + (r/r_0)^m), else
      *
+     * c -> rate
+     * r_0 -> scale
+     * m -> exponent
```
max-veit: Good. Perhaps we should add a comment to the effect that `c` is almost always left at 1?

tests/test_calculator.hh
```diff
@@ -267,7 +267,19 @@ namespace rascal {
...
          {"cutoff", {{"value", 2.0}, {"unit", "AA"}}},
-         {"smooth_width", {{"value", 1.0}, {"unit", "AA"}}}}};
+         {"smooth_width", {{"value", 1.0}, {"unit", "AA"}}}},
+        {{"type", "RadialScaling"},
+         {"cutoff", {{"value", 4.0}, {"unit", "AA"}}},
+         {"smooth_width", {{"value", 0.5}, {"unit", "AA"}}},
+         {"rate", {{"value", .0}, {"unit", "AA"}}},
+         {"exponent", {{"value", 4}, {"unit", ""}}},
+         {"scale", {{"value", 2.5}, {"unit", "AA"}}}},
+        {{"type", "RadialScaling"},
```
max-veit: Just out of curiosity, what do the extra parameter sets do to the test runtime? I thought the gradient tests were relatively fast, but I don't remember the last time I ran them in debug.

bindings/rascal/representations/base.py
max-veit:
```suggestion
    if cutoff_function_type == 'ShiftedCosine':
```

felixmusil: this renaming is outside of the aim of this branch :)

bindings/rascal/representations/spherical_covariants.py
felixmusil: SphericalExpansion is directly used at the c++ level so you need to make the hypers dict in every calculators.

tests/test_calculator.hh
felixmusil: not much as I recall. the additional tests are restricted to the spherical expansion.

felixmusil: the overall time for the tests is 7 sec in release

bindings/rascal/representations/base.py
max-veit: Is it though? When do we do the renaming then? Because this seems like a pretty good place, then we don't have the extra pain wrangling names when B-P is merged in.

max-veit: I count 14 files where the renaming would need to be done (excluding the kernel with exactly the same name, "Cosine", which I think is slightly misleading -- it's a dot product -- but that's a discussion for another day):

```
bindings/rascal/representations/spherical_covariants.py
65:                 cutoff_function_type="Cosine", normalize=True,

bindings/rascal/representations/spherical_expansion.py
54:                 gaussian_sigma_constant=0., cutoff_function_type="Cosine",

bindings/rascal/representations/spherical_invariants.py
63:                 cutoff_function_type="Cosine",

examples/spherical_invariants_example.cc
72:  json fc_hypers{{"type", "Cosine"},

examples/test_nl.cc
99:  json fc_hypers{{"type", "Cosine"},
143:      {"zeta", 2}, {"target_type", "Structure"}, {"name", "Cosine"}};

examples/spherical_expansion_example.cc
72:  json fc_hypers{{"type", "Cosine"},

examples/spherical_invariants_profile.cc
77:  json fc_hypers{{"type", "Cosine"},

examples/spherical_expansion_profile.cc
74:  json fc_hypers{{"type", "Cosine"},

src/representations/calculator_spherical_expansion.hh
676:      if (fc_type.compare("Cosine") == 0) {
677:        this->cutoff_function_type = CutoffFunctionType::Cosine;
679:            make_cutoff_function<CutoffFunctionType::Cosine>(fc_hypers);
683:                               ": \'Cosine\'.");
796:    case CutoffFunctionType::Cosine: {
797:      this->compute_by_radial_contribution<CutoffFunctionType::Cosine>(

src/representations/cutoff_functions.hh
47:    enum class CutoffFunctionType { Cosine, RadialScaling, End_ };
82:    struct CutoffFunction<internal::CutoffFunctionType::Cosine>

tests/test_calculator.hh
123:        {{"type", "Cosine"},
126:        {{"type", "Cosine"},
191:        {{"type", "Cosine"},
194:        {{"type", "Cosine"},
269:        {{"type", "Cosine"},
272:        {{"type", "Cosine"},
352:        {{"type", "Cosine"},
355:        {{"type", "Cosine"},
441:        {{"type", "Cosine"},
478:        {{"type", "Cosine"},

scripts/generate_spherical_covariant_ref_data.py
77:                          "cutoff_function_type": "Cosine",

scripts/generate_spherical_invariant_ref_data.py
86:                                    "cutoff_function_type": "Cosine",

scripts/generate_spherical_expansion_ref_data.py
77:                                          "cutoff_function_type": "Cosine",
```
I'd say this is 15 minutes of work at most; if you agree, I'll go ahead and do the renaming.

ceriottm:
:+1: for renaming, and also for renaming the "cosine" kernel that is really
just a dot product

felixmusil:
I don't agree that it is just a dot product since it is a proper kernel only if the representation vectors are normalized, if not the matrix you get is not a kernel strictly speaking.
For efficiency though we assume that the normalization has already been done.

bindings/rascal/representations/base.py
felixmusil: If you feel like it just go on with the renaming of the cutoff function. Note that you need to generate again the reference data that have this keyword so that the test can work properly.

max-veit: Yeah, I noticed after the tests failed. Just this simple rename resulted in ~16 MiB of changes... I'm really not a fan of these huge binary "reference data" files (which really just compare to previous versions).

max-veit:
But is "Cosine" really a clearer name than "DotProduct"? I don't think so. But let's discuss this at the next meeting. For now let's stick with the cutoff-function rename, which is well within the scope of this branch.

Issue #30, opened 2018-10-24
read the docs integration


felixmusil:

rosecers:
![](https://github.trello.services/images/mini-trello-icon.png) [Work on read the docs implementation for lib_rascal documentation](https://trello.com/c/uRTWXYti/7-work-on-read-the-docs-implementation-for-librascal-documentation)

max-veit:
@Luthaf assigning you because you mentioned it might be possible to do an automatic build on gh-pages?

Luthaf:
I'll look into this on Monday ! I have a working version with Travis CI (http://chemfiles.org/chemfiles/), I just need to adapt it to Circle CI

Luthaf:
This is done, documentation is now online and auto updated at https://cosmo-epfl.github.io/librascal/index.html

PR #135, opened 2019-09-30
Simplify Circle Ci config

Luthaf:
- Remove un-used files
- Remove duplication in config.yml
- Use a single point of knowledge (requirements.txt) for all Python dependencies


Luthaf:
This is a first step toward #30.

Luthaf:
@agoscinski, I think I discussed this with you: pip does not support conda style yaml files, and since some dependencies of rascal only exist on PyPi, I went with a pip-style requirements.txt file. This should still be usable with conda environments, using conda's pip.

Luthaf:
Also ping @gibertif, since you wrote the circle Ci config initially.

Luthaf:
Also, do we need to test on all version of GCC and Clang? We could just use the oldest and newest version of these (gcc 5 & 9, clang 4 & 9), thus cutting on CI run time.

README.rst
```diff
@@ -12,13 +12,13 @@ collection of tools to this end.
 
 librascal is currently considered a standalone code. However, we aim to
 provide enough flexibility to interface it with other codes such as
-LAMMPS and PLUMED-2.0.  It can be used as a C++ library as well as a
-python module.  To be able to call it from python, we have used the
+LAMMPS and PLUMED-2.0. It can be used as a C++ library as well as a
```
max-veit: French spacing...? In all seriousness, though, the double space makes it easier to read in code; this is of course ignored by the typesetting to give the "modern" single-space.

Luthaf: No, I have [proselint](https://github.com/amperser/proselint) configured locally, and it was complaining :smile: 

I though this was standard in English too, but I can change it back if you want.

Luthaf: The exact message was "Inconsistent spacing after period (1 vs. 2 spaces)." so I guess both are accepted in English, and proselint just want me to be more consistent :smile:

requirements.txt
```diff
@@ -0,0 +1,10 @@
+numpy
+scipy
+# for documentation
+sphinx==2.1.2
+breathe==4.13.1
+sphinx_rtd_theme
+nbsphinx
+# Developers tools
+jupyter
+cpplint
```
agoscinski: I added some libraries, not essential for the tests but for running the tutorials and developer tools

max-veit: I wouldn't call running the tutorials a "developer tool"

PR #142, opened 2019-10-03
Added deepclean custom target to clean building folder

agoscinski:
Using the target deepclean with `make deepclean` does
```
shopt -s extglob
rm -fr -- !(external|third-party)
```

Had to use the command
`find ! -regex "\\./\\(external\\|third-party\\).*" -delete`
Because the other did not work. Also I had to put it in the script because cmake has problems with \|  and \\ are ignored, see for yourself and try
`COMMAND echo '\\(external\\|third-party\\).*'`
If someone has a solution to prevent using a script file, then please do so. I have tried a lot and it did not work and this solution was the best I could find.

Feature was called lazyclean before, therefore different branch name


Luthaf:
Locally, I got this:

```
execute_process(
    COMMAND find -E . -not -regex "./(external|third-party).*" -delete
    WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
)
```

And it seems to be working. It needs the `-E` flag, but this might be a difference between gnu find and BSD find (the one by default on macOS)

---

What is the use case for this? If the idea is to clean built objects while keeping external code, `rm -rf src` (inside the build folder!) should also do the trick.

felixmusil:
the goal is to regenerate the cmake files since in some cases they don't get updated properly even when regenerating. 

the -E argument is not in the gnu find btw.

Luthaf:
> the goal is to regenerate the cmake files since in some cases they don't get updated properly even when regenerating.

That is really weird. Do you have more details ? I would rather fix the underlying issue if possible.

felixmusil:
I don't think that it can be fix through cmake.
Cmake does not restart generating config files from scratch which is nice but sometimes it does not detect properly the changes and misses them.
There are many moments where I end up restarting the build folder, e.g. linking mismatch between branches or the python environment is not the good one.
I think that restart the configuration from scratch is the simple and all-rounder solution unless you have a better way to achieve this ?

Le 6 oct. 2019 17:40, Guillaume Fraux <notifications@github.com> a écrit :

the goal is to regenerate the cmake files since in some cases they don't get updated properly even when regenerating.

That is really weird. Do you have more details ? I would rather fix the underlying issue if possible.

agoscinski:
I tried
```
execute_process(
    COMMAND find -E . -not -regex "./(external|third-party).*" -delete
    WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
)
```
but I get 
`/bin/sh: 1: Syntax error: "(" unexpected`
using it as COMMAND in a cmake target. Also if I execute the command in my terminal, it deletes everything.

max-veit:
`lazyclean` indeed -- it's taking more time to get this working than any of us will spend typing `shopt -s extglob; rm -rf -- !(build|external)` (or just `rm -- -rf ^(build|external)` on zsh :D ) and recalling the appropriate command with ctrl-R

agoscinski:
@max-veit I had this too, but then I used it by accident in the librascal main folder and lost the progress for the whole day. The solution in this branch already works on linux systems.

max-veit:
> @max-veit I had this too, but then I used it by accident in the librascal main folder and lost the progress for the whole day.

Then you really should be committing more often. If you want, you can always clean up your changes with `git rebase --interactive origin/branchname` (but see [here](http://think-like-a-git.net/sections/git-makes-more-sense-when-you-understand-x/example-4-lsd-and-chainsaws.html) for an alternative perspective on that command). As long as the glob doesn't touch dotfiles (and I'm pretty sure it doesn't...) you won't lose anything you haven't committed.

agoscinski:
changed to python script so that it should work now on multiple platforms

CMakeLists.txt
```diff
@@ -366,3 +366,10 @@ if(AUTOPEP8_FOUND)
   autopep8_add_subdirectory("${CMAKE_SOURCE_DIR}/bindings" "")
   autopep8_add_subdirectory("${CMAKE_SOURCE_DIR}/scripts" "")
 endif()
+
+# Removes every file in the build folder except the external libraries to
+# prevent redownloading external libraries
+add_custom_target(
+    deepclean
+    COMMAND python ${CMAKE_SOURCE_DIR}/scripts/developer_utils.py deepclean
```
Luthaf: Does the script work with Python 2? It might need a few `from __future__ import ...`

agoscinski: yes it does

felixmusil: it is better to use the python detected by Cmake compared to using the one in path to avoid surprises.

```suggestion
    COMMAND "${PYTHON_EXECUTABLE}" ${CMAKE_SOURCE_DIR}/scripts/developer_utils.py deepclean
```

mastricker: I agree with @Felix (and the comment @Luthaf ). Since we are imposing the major version, we should also use it consistently.
On my machine, the `python` command would be `2.7`.

PR #144, opened 2019-10-03
Bugfix/neighbors

felixmusil:
Fix the rational for generating periodic images in AdaptorNeighborList. 

now raise an error if atoms are outside of the unit cell.

- [x] fix the implementation 

- [x] fix the tests


tests/test_structure.hh
```diff
@@ -126,7 +126,7 @@ namespace rascal {
     ManagerFixtureTwoHcp()
         : ManagerFixtureTwo<StructureManagerCenters>{}, pbc{{true, true, true}},
           cell_1(3, 3), cell_2(3, 3), positions_1(3, 2), positions_2(3, 2),
-          atom_types(2), cutoff{0.8} {
+          atom_types(2), cutoff{0.4} {
```
felixmusil: are there still some neighbors with such a small cutoff ?

mastricker:
Man, it is a work in progress, please be patient.

On Oct 7, 2019 14:18, felixmusil <notifications@github.com> wrote:

@felixmusil commented on this pull request.

________________________________

In tests/test_structure.hh<https://github.com/cosmo-epfl/librascal/pull/144#discussion_r331984842>:

> @@ -126,7 +126,7 @@ namespace rascal {
     ManagerFixtureTwoHcp()
         : ManagerFixtureTwo<StructureManagerCenters>{}, pbc{{true, true, true}},
           cell_1(3, 3), cell_2(3, 3), positions_1(3, 2), positions_2(3, 2),
-          atom_types(2), cutoff{0.8} {
+          atom_types(2), cutoff{0.4} {


are there still some neighbors with such a small cutoff ?

-

mastricker:
Quick question: What was the reason to kill the `epsilon` in the neighbour list?

The only failing test was the hcp cell - the number of neighbours was not the same for a basal x-y and a prism x-y plane. I went back to the beginning:
I have checked and rechecked and rechecked the definition of both the hcp cells and I can not find a flaw.
I am now sure that the failing test is a numerical issue. If I include the `epsilon` again  in the mesh definition, it works flawlessly. Even if it is just a `1%` shift of the mesh to not put the mesh overlapping (in theory) with the origin. So the "zero" in this test case is iffy.

Please do convince me of the opposite, if I missed something.

ceriottm:
so the epsilon shift was responsible, together with some quirks in the algebra and an issue in the detection of the limits of the cell, of all our trouble. @tjunge has described the epsilon shift as a trick to get more "visually understandable" folded cells. my concern is that if we introduce a shift now, there will be a case in which it hits us back. we need to understand why only 11 replicas are found - in a jupyter nb implementation of the same algorithm I get all 12.

mastricker:
I'm on it.
Ovito gives me 11 in the basal plane but I don't know how it is exactly implemented.

 Have to check the prism plane again tomorrow.

On Oct 7, 2019 22:43, Michele Ceriotti <notifications@github.com> wrote:

so the epsilon shift was responsible, together with some quirks in the algebra and an issue in the detection of the limits of the cell, of all our trouble. @tjunge<https://github.com/tjunge> has described the epsilon shift as a trick to get more "visually understandable" folded cells. my concern is that if we introduce a shift now, there will be a case in which it hits us back. we need to understand why only 11 replicas are found - in a jupyter nb implementation of the same algorithm I get all 12.

-

Issue #127, opened 2019-09-24
Error importing librascal when compiled with gcc9


max-veit:
Utterly puzzling and frustrating, probably related to [this upstream pybind11 issue](https://github.com/pybind/pybind11/issues/994) which they currently seem unwilling or unable to investigate.

Steps to reproduce (on the current `master`):
- Make a clean conda environment with the necessary packages (`conda list` 
[attached](https://github.com/cosmo-epfl/librascal/files/3647299/conda.list.txt))
- Install GCC 9.1.0
- Run `cmake`, let it install `pybind11` (error occurs with versions 2.2.2, 2.3.0 (rascal default), and 2.4.2 (latest stable)) (
[`CMakeCache.txt` attached](https://github.com/cosmo-epfl/librascal/files/3647397/CMakeCache.txt))

- Build everything with `make`
- Try to import a rascal representation from the build directory (though the exact same thing happens when running in another directory and adding `build/` to the python path):
```python
Python 3.7.3 (default, Mar 27 2019, 22:11:17) 
[GCC 7.3.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os; os.getcwd()
'/local/scratch/mveit/codes/librascal/build'
>>> from rascal.representations import SphericalInvariants
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/local/scratch/mveit/codes/librascal/build/rascal/representations/__init__.py", line 1, in <module>
    from .coulomb_matrix import SortedCoulombMatrix
  File "/local/scratch/mveit/codes/librascal/build/rascal/representations/coulomb_matrix.py", line 4, in <module>
    from ..neighbourlist import AtomsList
  File "/local/scratch/mveit/codes/librascal/build/rascal/neighbourlist/__init__.py", line 1, in <module>
    from .structure_manager import AtomsList, get_neighbourlist, convert_to_structure_list
  File "/local/scratch/mveit/codes/librascal/build/rascal/neighbourlist/structure_manager.py", line 1, in <module>
    from .base import (NeighbourList, NeighbourListFactory, is_valid_structure,
  File "/local/scratch/mveit/codes/librascal/build/rascal/neighbourlist/base.py", line 1, in <module>
    from ..lib import NeighbourList
  File "/local/scratch/mveit/codes/librascal/build/rascal/lib/__init__.py", line 1, in <module>
    from ._rascal import (NeighbourList, Models,
[...]

PR #146, opened 2019-10-04
Use a custom Docker image to speed up Circle CI

Luthaf:
A lot of time (~ 10min) is spent installing dependencies on Circle CI. This PR creates a custom docker image with dependencies pre-installed to try to reduce the build time. 

~~Unfortunately, I hit the error from #127 when building with GCC 9, so this will have to be fixed first.~~

**Unresolved:**

- [x] The docker image is currently pushed as luthaf/rascal-ci. Should we setup a docker hub account for rascal, independent if myself? Set-up as cosmoepfl/rascal-ci


felixmusil:
Having a main image is indeed much better. 
Why is GCC 9 important to this PR ?

Luthaf:
>  Why is GCC 9 important to this PR ?

Not really important by itself, I also wanted to improve build time by using only selected compilers, as discussed in the last meeting. But if I can't fix the issue, I'll fallback to gcc 7

.circleci/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+RUN apt-get update && \
+    apt-get install -y software-properties-common wget && \
+    wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -  && \
+    add-apt-repository ppa:ubuntu-toolchain-r/test && \
+    add-apt-repository "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-9 main"
+
+RUN apt-get update && \
+    apt-get install -y \
+        git \
+        cmake \
```
felixmusil:
```suggestion
        cmake \
        vim \
```

felixmusil: in the case one would need to interact with files on the image

Luthaf:
> A lot of time (~ 10min) is spent installing dependencies on Circle CI. This PR creates a custom docker image with dependencies pre-installed to try to reduce the build time.

After more checking, the 10min "installing dependencies" step was an outlier, and the usual timing for this step is more 2min. This makes this PR less interesting, since it introduces more complexity for less gains.

felixmusil:
Still it makes more sense to download and install the common utilities beforehand so I would still move forward with this PR.

.circleci/Dockerfile
Luthaf: sounds good!

Luthaf:
> The docker image is currently pushed as luthaf/rascal-ci. Should we setup a docker hub account for rascal, independent if myself?

@ceriottm what do you think about this?

PR #147, opened 2019-10-04
fix the demangling of gcc 9 __PRETTY_FUNCTION__

felixmusil:
solves issue #127

### DIFF
--- a/src/rascal_utility.hh
+++ b/src/rascal_utility.hh
@@ -31,8 +31,12 @@
 // Detects which compiler is used
 #if defined(__clang__)
 #define CLANG_COMPILER
-#elif defined(__GNUC__) || defined(__GNUG__)
-#define GCC_COMPILER
+#elif (defined(__GNUC__) || defined(__GNUG__)) && (__GNUC__ > 8)
+// there is a slight modification in the output of __PRETTY_FUNCTION__ in gcc 9
+// here we expect it to be forwarded to newer versions
+#define GCC_COMPILER_9_AND_UPPER
+#elif (defined(__GNUC__) || defined(__GNUG__)) && (__GNUC__ <= 8)
+#define GCC_COMPILER_8_AND_LOWER
 #endif
 
 #include <utility>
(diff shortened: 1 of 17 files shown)


src/rascal_utility.hh
```diff
@@ -263,7 +267,16 @@ namespace rascal {
...
 // The output of of Pretty Function depends on the compiler
 // the #define strings is a pain to split
-#if defined(GCC_COMPILER)
+#if defined(GCC_COMPILER_9_AND_UPPER)
+#define FUNCTION_MACRO __PRETTY_FUNCTION__
+#define PREFIX                                                                 \
+  "static const string rascal::internal::GetTypeNameHelper<T>::GetTypeName() " \
+  "[with T = "  // NOLINT
+#define SUFFIX_1                                                               \
+  "; std::string = std::__cxx11::basic_string<char>]"  // NOLINT
```
Luthaf: What happens if one compiles this with `-D_GLIBCXX_USE_CXX11_ABI=0`?

felixmusil: I have no clue what this flag means...

max-veit:
Well, this fixes the issue for me, thanks! I'll let @Luthaf have the last word, but consider this a tentative approval from my end.

src/rascal_utility.hh
felixmusil: I just compiled with gcc-9 and this flag with cmake and it works

Luthaf: > I have no clue what this flag means...

GCC > 5 has a different ABI when in C++11 mode (hence the  `__cxx11` in the name), and this flag force the use of the older ABI. It is required on some (old) systems, where libstdc++ is compiled with the old ABI.

But if this works for you, it should be good.

Luthaf:
Could you add some tests for this function in case future compiler updates break it Like asserting we get the right type name for basic types and a few templated types.

--- 

The overall implementation of this feature feels fragile to me, what is the rational for not using [`typeid`](https://en.cppreference.com/w/cpp/types/type_info)?

src/rascal_utility.hh
felixmusil: ok sorry for the confusion. I compile with this flag and boost has some linking error because of it. I guess I would have to recompile boost test with this flag for it to work.
Again what is the point of this flag ?

felixmusil:
> The overall implementation of this feature feels fragile to me, what is the rational for not using typeid?

I agree with you. I integrated it to avoid RTTI and the cost it entails. I did a new  [research](https://stackoverflow.com/questions/579887/how-expensive-is-rtti) and it seems that the cost might have been inflated in my previous reading. 

Anyway we should see if it has an impact in our cases. If it does not, then we can move to the RTTI and ABI demangling.

src/rascal_utility.hh
Luthaf: > Why would we want to use this flag while compiling ?

This flag is required on "older" Linux distributions, where the base system compiler is older than GCC 5. This means that the standard C++ library as well as most of the code will be using the older ABI, so even when using more recent compilers on this platform, you want to be compatible and use this flag. A specific example would be CentOS 7 (the latest version of CentOS), where the base compiler is GCC 4.8.5. More modern compilers such as GCC 7 are provided, but use this flag by default to be compatible with the system.

This flags was also required for me on macOS, since the Homebrew version of Boost seems to be using it, meaning that I needed to provide it.

---

Since this is somehow niche usage, I am okay with the fix in this PR, assuming we move away from this implementation and use RTTI when possible.

PR #152, opened 2019-10-08
Use typeid for geting type name

### LINKED ISSUE #149: Use RTTI/typeid instead of __PRETTY_FUNCTION__ parsing to get type names
See #147 for more background.

Luthaf:
Fix #149


src/rascal_utility.cc
```diff
@@ -0,0 +1,83 @@
...
+#include <iterator>
+
+#include "rascal_utility.hh"
+
+#ifdef __GNUG__
+#include <cxxabi.h>
+#endif
+
+std::string rascal::internal::type_name_demangled(const char * name) {
+#ifdef __GNUG__
```
felixmusil: Is this flag also present with clang ?

src/rascal_utility.cc
```diff
@@ -0,0 +1,83 @@
...
+  int status = 0;
```
felixmusil:
```suggestion
  int status{0};
```

src/rascal_utility.cc
```diff
@@ -0,0 +1,83 @@
...
+  char * demangled = abi::__cxa_demangle(name, nullptr, nullptr, &status);
```
felixmusil:
```suggestion
  char * demangled{abi::__cxa_demangle(name, nullptr, nullptr, &status)};
```

src/rascal_utility.cc
```diff
@@ -0,0 +1,83 @@
...
+  auto result = std::string(demangled);
```
felixmusil: To follow the code convention strictly here one should use bracket. I personally feel bracket initialization is has no advantage when using auto so I would keep this as is.

src/rascal_utility.hh
```diff
@@ -249,128 +236,33 @@ namespace rascal {
...
-     */
-    inline std::string get_filename_extension(const std::string & filename) {
-      auto const pos = filename.find_last_of(".");
-      std::string extension{""};
-      if (pos != std::string::npos) {
-        extension = filename.substr(pos + 1);
-      }
-      return extension;
-    }
+    //! Reads a binary file
```
felixmusil:
```suggestion
/**
 * Read a binary file
 * @throw if the file can't be opened.
 */
```

src/rascal_utility.hh
```diff
@@ -249,128 +236,33 @@ namespace rascal {
...
+    void replace(std::string & string, const std::string & pattern,
+                 const std::string & replacement);
 
-      // reserve capacity
-      vec.reserve(fileSize);
-
-      // read the data:
-      vec.insert(vec.begin(), std::istream_iterator<BINARY>(file),
-                 std::istream_iterator<BINARY>());
+    //! return a pretty form of the template typename
```
felixmusil:
```suggestion
    //! return a human readable name for the template typename
```

src/rascal_utility.hh
```diff
@@ -249,128 +236,33 @@ namespace rascal {
...
-      if (not file.is_open()) {
-        throw std::runtime_error(std::string("Could not open the file: ") +
-                                 filename);
-      }
-      // Stop eating new lines in binary mode!!!
-      file.unsetf(std::ios::skipws);
 
-      // get its size:
-      std::streampos fileSize;
+    //! Get the demangled name of a c++ type
```
felixmusil:
```suggestion
    //! Get the demangled name of a c++ type returned by typeinfo(T).name()
```

src/rascal_utility.cc
Luthaf: Yes, clang pretends to be GCC 4.2. Intel compiler also defines this (see https://stackoverflow.com/q/38499462)

src/rascal_utility.cc
mastricker: Actually there is a case in the coding convention stating exactly this:

> Never assign a braced-init-list to an auto local variable. In the single element case, what this means can be confusing.

So it should _not_ be with braces.

src/rascal_utility.cc
Luthaf: Could this be enforced with a cpplint/clang-format rule instead of having to manually check and change it?

src/rascal_utility.cc
Luthaf: Outside of this PR, but what is the rational for enforcing brace initialization as the coding style?

Luthaf:
> Just out of curiosity: Is this where you found the character-by-character read of the read_from_binary?

Yes =)

src/rascal_utility.cc
mastricker: If I remember correctly the idea is that when using the brace initializer, no narrowing is allowed. And to be consistent, we encourage it in the coding convention to use `{}` when possible `()` when necessary and try to avoid `=`.

src/rascal_utility.cc
mastricker: I'll have a look.

src/rascal_utility.cc
Luthaf: > If I remember correctly the idea is that when using the brace initializer, no narrowing is allowed.

Only when using a variable (see below) but OK.

```cpp
long long a{44};
short b{a};    // error
short c{44ll}; // ok
```

---

My main issue with it is that the syntax is a bit alien, especially for people coming from Python (which mean a lot of scientists). It also make the code a bit harder to parse when reading it, even more so if the initializer is long:

```
char * demangled = abi::__cxa_demangle(name, nullptr, nullptr, &status);
// VS
char * demangled{abi::__cxa_demangle(name, nullptr, nullptr, &status)};
```

The brace initialization looks more like a function call than a variable creation.

---

Anyway, as long as we can find a way to have it automatically enforced, I can deal with either. But I would rather not spend reviewer and implementer time checking for this =)

Luthaf:
@felixmusil you need to approve the changes before we can merge this.

src/rascal_utility.cc
mastricker: Does not seem to have the option, neither clang-format nor cpplint.
I would file this under: Whatever gets caught, will should change but we try not to be to picky about it. If it's caught in review, so be it. If not, well we have no ways to check automatically anyways.

PR #154, opened 2019-10-09
Python interface for selecting centres

### LINKED ISSUE #148: Add friendly Python interface to center_select
A common use case of the [center_select feature](https://github.com/cosmo-epfl/librascal/pull/102) is to mark only atoms of a given species as centres. In order to make this feature useful in practice, we need a Python interface that processes an `AtomsList` and sets the `center_atoms_mask` for each structure based on the given criteria (including, but possibly not limited to, atomic species).

max-veit:
Fix #148, in progress

Take advantage of the existing infrastructure in the Python bindings. Now you just have to process the atoms through one of these functions (we might also make convenience functions for operating on lists) and pass them to representation.transform() as usual.

### DIFF
--- a/bindings/rascal/neighbourlist/structure_manager.py
+++ b/bindings/rascal/neighbourlist/structure_manager.py
@@ -263,5 +263,7 @@ def mask_center_atoms_by_species(frame, species_select=[],
             old_mask = np.ones((frame.get_number_of_atoms(),), dtype='bool')
     else:
         old_mask = frame.arrays['center_atoms_mask']
+    # Python's "bitwise" operators do per-element logical operations in NumPy
+    # see for instance https://docs.scipy.org/doc/numpy/reference/ufuncs.html#comparison-functions
     mask = (old_mask | id_select) & ~id_blacklist
     frame.arrays['center_atoms_mask'] = mask
--- a/bindings/rascal/neighbourlist/structure_manager.py
+++ b/bindings/rascal/neighbourlist/structure_manager.py
@@ -1,8 +1,12 @@
+from collections.abc import Iterable
+from functools import reduce
+from operator import and_
+
+import numpy as np
+
 from ..lib import neighbour_list
 from .base import (NeighbourListFactory, is_valid_structure,
  adapt_structure, StructureCollectionFactory)
-from collections.abc import Iterable
-import numpy as np
 
 class AtomsList(object):
     """
@@ -122,6 +126,11 @@ def unpack_ase(frame):
     -------
     StructureManagerCenters
         base structure manager.
+
+    If the frame has an ase.atoms.arrays entry called
+    'center_atoms_mask' then it will be used as the center mask
+    (surprise) for any representations computed on this
+    StructureManager.
     """
     cell = frame.get_cell()
     positions = frame.get_positions()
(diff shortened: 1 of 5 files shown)


Luthaf:
Should these functions also be added to the documentation? At least in the developer doc part, and maybe in the main doc too.

max-veit:
Sure, but we don't have the Python doc included automatically yet (I tried `automodule` or something similar; it's a mess). Is this something we should add to the checklist for implementing a new feature?

max-veit:
Oh great, and it looks like we need yet another Sphinx extension to work with our NumPy-formatted docstrings. Vive [`napoleon`](http://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html#module-sphinx.ext.napoleon)!

Luthaf:
> Is this something we should add to the checklist for implementing a new feature?

Is there such checklist somewhere?

But yes, I think we should always have doc added to both the code and the live HTML version with each new feature.

max-veit:
> > Is this something we should add to the checklist for implementing a new feature?
> 
> Is there such checklist somewhere?

Not yet, but it should be part of the `CONTRIBUTING.rst` that I promised somewhere else to write when I get the chance.

Luthaf:
> Hang on, how were you able to approve the changes if it's still just a draft pull request?

I can approve but the merge button is (was) still disabled

max-veit:
Tsk tsk, approving before checking that the example runs...

Anyway, I fixed the bugs in the code (with some refactoring to allow "stacking" of different function calls), then I'll just add some unit tests and it'll be ready for a proper review.

bindings/rascal/neighbourlist/structure_manager.py
```diff
@@ -235,4 +255,13 @@ def add_center_atoms_mask_species(frame, species_select=[],
...
-    add_center_atoms_mask_by_id(frame, id_select, id_blacklist)
+    if 'center_atoms_mask' not in frame.arrays:
+        # add a default mask
+        if species_select:
+            old_mask = np.zeros((frame.get_number_of_atoms(),), dtype='bool')
+        else:
+            old_mask = np.ones((frame.get_number_of_atoms(),), dtype='bool')
+    else:
+        old_mask = frame.arrays['center_atoms_mask']
+    mask = (old_mask | id_select) & ~id_blacklist
```
Luthaf: Could you add a comment here, I guess not everyone is familiar with boolean binary operations

max-veit: These are standard Python operators and the result (and logic) is clearly described in the docstring. I don't believe in comments that just repeat the code, which is what this would have to be.

max-veit: If you insist, I can add an RTFM to here: https://docs.scipy.org/doc/numpy/reference/ufuncs.html#comparison-functions

But I really don't think it's necessary.

Luthaf: These operators are not often used in Python, the preferred style to use `and`, `or`, `not`. This is not possible here for the reason showed in your link, but might still confuse newcomers coming to this code.

I would still at least add something like 'combine all masks using bitwise operators, since standard Python operators should not be used with numpy array (<link>)'.

max-veit: If a newcomer is digging into the code and sees this line, I'd better hope they already know the difference between `and` and `&` when it comes to NumPy arrays (practically anyone who's worked with numpy before has seen that iconic "The truth value of an array is ambiguous..." error).

max-veit: Seriously, I'm not doing the job of standard documentation. The comment you suggested belongs in the Numpy docs, not in ours.

Luthaf: I was slightly confused by reading this, and it is not much work for something that can help newcomers to the code.

But if you really don't want to, I'll be fine 😃

PR #158, opened 2019-10-10
Simplified layer utilities in cluster ref key - We need to talk about `compute_cluster_layer`

agoscinski:
For the layer utilities in the cluster_ref_key.hh:
- functions have been simplified
- tests have been added
- doc has been added

The `compute_cluster_layer` function has been replaced with the `get_layer` function. `compute_cluster_layer` is still needed, there is a simplified and renamed version called  `get_min_layer`.

The `compute_cluster_layer` function returned for `Order` values outside of the range of LayersByOrder still 0. This could cause undefined behaviour. I added static_asserts to it. There was some usage of the `compute_cluster_layer` function with `Order` values outside of the range in the `py_bind_structure_manager.cc` in the `add_iterator` function. This is now fixed. I believe that this did not cause any bug, because the function was only compiled, but never invoked.


bindings/bind_py_structure_manager.cc
```diff
@@ -124,11 +124,16 @@ namespace rascal {
...
-  //! Bind iterator and ClusterRef for Order >= 2
+  /**
+   * Bind iterator and ClusterRef for Order >= 2
+   *
+   * @pre Order >= 2
+   */
   template <typename StructureManagerImplementation, size_t Order>
   decltype(auto) add_iterator(
       py::module & m,
       PyClusterRef<StructureManagerImplementation, Order - 1> & py_cluster) {
```
agoscinski: Before this `PyClusterRef` was compiled with `Order=1` resulting in a compiler buffer overflow when accessing the `LayerByOrder`

src/structure_managers/cluster_ref_key.hh
```diff
@@ -41,140 +41,125 @@
...
-  //! existing layer by order.
-  template <size_t MaxOrder, class T>
+
+  /**
+   * Increases each layer of all existing orders.
+   *
+   * @tparam T the std::index_sequence representing the layer by order sequence.
+   */
+  template <class T>
   struct LayerIncreaser {};
```
agoscinski: Removed the MaxOrder template parameter which never had any use.

src/structure_managers/cluster_ref_key.hh
```diff
@@ -41,140 +41,125 @@
...
+
   /**
-   * Dynamic access to depth by cluster order
+   * Extends the layer by order index sequence by an additional cluster order.
+   * Used when order is increased by an adaptor. See traits of`AdaptorMaxOrder`.
+   *
+   * @tparam T the std::index_sequence representing the layer by order sequence.
    */
+  template <class T>
+  struct LayerExtender;
```
agoscinski: Same here: Removed the MaxOrder template parameter which never had any use.

src/structure_managers/cluster_ref_key.hh
```diff
@@ -41,140 +41,125 @@
...
   };
 
-  template <size_t MaxOrder, size_t... Ints>
+  template <size_t... LayersByOrder>
   using LayerExtender_t =
-      typename LayerExtender<MaxOrder, std::index_sequence<Ints...>>::type;
+      typename LayerExtender<std::index_sequence<LayersByOrder...>>::type;
 
   /* ---------------------------------------------------------------------- */
-  //! Dynamic access to all layers by cluster dimension (probably not necessary)
```
agoscinski: Is now not needed anymore see `get_min_layer`

src/structure_managers/cluster_ref_key.hh
```diff
@@ -41,140 +41,125 @@
...
+   * specific layer in a manager stack.
+   *
+   * @tparam ActiveMaxOrder the maximum order to check the `LayersByOrders` for
+   *         the minimum value.
+   * @param layers_by_order the type captures the layers of each order in an
+   *        index sequence.
+   *
+   * @return the smallest value in the given index sequence up to Order elements
    */
-  template <size_t index, size_t... Ints>
```
agoscinski: renamed the `get` function to `get_index` because the compiler confused this sometime with std::get

src/structure_managers/cluster_ref_key.hh
```diff
@@ -41,140 +41,125 @@
...
+    static_assert(ActiveMaxOrder > 0,
+                  "ActiveMaxOrder should be greater than 0.");
+    static_assert(ActiveMaxOrder <= sizeof...(LayersByOrder),
+                  "ActiveMaxOrder should not greater than the MaxOrder.");
+    // transforms the LayersByOrder to an array
+    constexpr size_t arr[] = {LayersByOrder...};
+    return *std::min_element(std::begin(arr), std::begin(arr) + ActiveMaxOrder);
   }
 
-  /* ---------------------------------------------------------------------- */
```
agoscinski: This is nowhere used, so I removed it

src/structure_managers/structure_manager.hh
```diff
@@ -560,8 +560,7 @@ namespace rascal {
 
     template <size_t Order>
     constexpr static size_t cluster_layer_from_order() {
-      static_assert(Order > 0, "Order is <1 this should not be");
```
agoscinski: I moved the static assert to the get_layer function

tests/test_cluster_ref_key.cc
```diff
@@ -0,0 +1,117 @@
...
+    BOOST_CHECK(arr_order_one == arr_order_one_ref);
+
+    std::array<size_t, 2> arr_order_two = get_layers(order_two_seq{});
+    std::array<size_t, 2> arr_order_two_ref = {1, 1};
+    BOOST_CHECK(arr_order_two == arr_order_two_ref);
+
+    std::array<size_t, 3> arr_order_three = get_layers(order_three_seq{});
+    std::array<size_t, 3> arr_order_three_ref = {1, 1, 0};
+    BOOST_CHECK(arr_order_three == arr_order_three_ref);
+    print_index_sequence(std::make_index_sequence<5>{});
```
felixmusil:
```suggestion
     if (verbose) {
       print_index_sequence(std::make_index_sequence<5>{});
     }
```

tests/test_cluster_ref_key.hh
```diff
@@ -0,0 +1,79 @@
...
+#include "tests.hh"
+#include <tuple>
+#include "structure_managers/cluster_ref_key.hh"
+
+/**
+ * Prints the index sequence of an std::index_sequence type by recursively
+ * printing the head element.
+ *
+ * @param _ the std::index_sequnce to print
+ */
```
felixmusil: to be deleted

max-veit:
What exactly is the `invalid` label meant to signify here?

src/structure_managers/cluster_ref_key.hh
```diff
@@ -41,140 +41,125 @@
...
+  template <size_t ActiveMaxOrder, size_t... LayersByOrder>
+  constexpr size_t get_min_layer(
+      const std::index_sequence<LayersByOrder...> & /*layers_by_order*/) {
+    static_assert(ActiveMaxOrder > 0,
+                  "ActiveMaxOrder should be greater than 0.");
+    static_assert(ActiveMaxOrder <= sizeof...(LayersByOrder),
+                  "ActiveMaxOrder should not greater than the MaxOrder.");
+    // transforms the LayersByOrder to an array
+    constexpr size_t arr[] = {LayersByOrder...};
+    return *std::min_element(std::begin(arr), std::begin(arr) + ActiveMaxOrder);
```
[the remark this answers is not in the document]
mastricker: I am not following the logic of the naming `ActiveMaxOrder`. That is the `Order`for which the function returns the current layer?

src/structure_managers/cluster_ref_key.hh
```diff
@@ -41,140 +41,125 @@
...
+   *
+   * @return the smallest value in the given index sequence up to Order elements
    */
-  template <size_t index, size_t... Ints>
-  constexpr size_t get(std::index_sequence<Ints...>) {
-    return get<index>(std::make_tuple(Ints...));
+  template <size_t ActiveMaxOrder, size_t... LayersByOrder>
+  constexpr size_t get_min_layer(
+      const std::index_sequence<LayersByOrder...> & /*layers_by_order*/) {
+    static_assert(ActiveMaxOrder > 0,
```
mastricker: Again. At least in my memory, the 0th order was the one for atoms. Either I am missing something now and I don't understand it, or we do have an unused entry in the index sequence, the 0th one.

tests/test_cluster_ref_key.cc
```diff
@@ -0,0 +1,117 @@
+/**
+ * @file   test_cluster_ref_key.cc
+ *
+ * @author Alexander Goscinski <alexander.goscinski@epfl.ch>
+ * @date   10 Thu 2019
```
mastricker: That date is funky :)

tests/test_cluster_ref_key.cc
mastricker: Plus a verbose flag at the top of the test.

tests/test_cluster_ref_key.hh
```diff
@@ -0,0 +1,79 @@
+/**
+ * @file   test_cluster_ref_key.hh
+ *
+ * @author Alexander Goscinski <alexander.goscinski@epfl.ch>
+ * @date   10 Thu 2019
```
mastricker: Funky date again.

tests/test_cluster_ref_key.hh
```diff
@@ -0,0 +1,79 @@
...
+  };
+
+  /**
+   * Prints the index sequence `Ints` defined in the std::index_sequence type of
+   * the parameter `sequence` by recursively printing the head element.
+   *
+   * @param sequence sequence to be printed
+   */
+  template<size_t... Ints>
+  void print_index_sequence(std::index_sequence<Ints...> /*sequence*/) {
```
mastricker: Would I be right in declaring this a helper-function?
If so, could you add that to the description.

mastricker:
For reference:
The point we were not sure with the proposed changes was that it would still be possible to make a property and attach it to a e.g. and adaptor strict and fill it with some values. Then you put another adaptor strict on top with a smaller cutoff. Now you iterate over the second one, but access the property which is attached to the lower one. There the capability of choosing _the other_ cluster index from below is needed. This was not entirely clear, but it work.

PR #164, opened 2019-10-16
Disable circle Ci on gh-pages (for real)

Luthaf:


Luthaf:
And it works! https://app.circleci.com/jobs/github/cosmo-epfl/librascal/9828

The previous version was not tested and failed (see https://app.circleci.com/jobs/github/cosmo-epfl/librascal/9708)

Luthaf:
The deploy-docs.sh is creating a "dummy" circle  CI config file, I took the previous one from an online example that might have been using the previous version of circle CI and was invalid with the current version.

So I rewrote it with the current config style and tested it by enabling doc deploy on this PR (I then removed the corresponding commit, to ensure doc is only built on master)

Issue #121, opened 2019-09-18
Tests take too long, cut them down below 10 s in debug


tjunge:

agoscinski:
there has been some dramatic improvements of test time in the feat/interpolator branch by reducing the length parameter in the kernel tests, it is now around twice as fast.
```
  struct StrictNLCollectionFixture : MultipleStructureManagerNLStrictFixture {
    using Parent = MultipleStructureManagerNLStrictFixture;
    using ManagerTypeHolder_t = typename Parent::ManagerTypeHolder_t;

    StrictNLCollectionFixture() : Parent{} {};

    ~StrictNLCollectionFixture() = default;

    std::string filename{"reference_data/dft-smiles_500.ubjson"};
    int start{4};
    int length{3};
  };
```

mastricker:
I guess this one is solved? I have < 10 seconds seconds right now if I run them in parallel?

Luthaf:
Even if @agoscinski works improves the results here, test runtime in debug is still more than 3 min, so this should stay open.

Luthaf:
no problem! =)

mastricker:
Just adding things I see as time consumers here for reference once the works starts:

- `rascal::kernerls_test::multiple_kernel_compute_test`
- `read_binary_file` seems to take an awful lot of time for a 300kb file

Luthaf:
Enabling -Og (optimizations that should not interfere with debugging) --- as done in #165 --- brings the test time from 147s to 9s locally.

Else, the biggest offenders are test_calculator (98s at -O0/5s at -Og) and test_kernel (15s at -O0/1s at -Og).

agoscinski:
move `-Og` flag to a new `CMAKE_BUILD_TYPE`. Suggested name `DebugOg`

PR #165, opened 2019-10-16
Use modern target-based cmake

Luthaf:
CMake good practice is to use targets to set include path and compilation option instead of the global, folder based `include_directories` and friends, this PR does just that!

It also moves the test configuration in `tests/CMakeLists.txt` and create a separate binary for each test file.

I also removed most of the `always_rebuild` target, since [https://cmake.org/cmake/help/latest/command/add_custom_target.html](add_custom_target) is always considered out of date anyway.

---

Since this touches a fundamental piece of infrastructure for librascal, I would appreciate if you could take the time to checkout the branch locally and test it. Please check for weird interactions between different options, and option changes.

How to checkout a PR locally: https://help.github.com/en/articles/checking-out-pull-requests-locally


CMakeLists.txt
```diff
@@ -52,71 +52,47 @@ ENDIF()
...
-set(ENABLE_PROFILING "OFF" CACHE BOOL
-  "If on, adds gprof profiling information to some of the examples. \
-  (GNU compilers only)")
+if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" OR
+    "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR
+    "${CMAKE_CXX_COMPILER_ID}" STREQUAL "AppleClang")
 
-SET(TYPE_ARCHITECTURE "native" CACHE STRING "Choose the type of architecture to compile to in release mode. Default is to let the compiler decide")
+    set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -Og")
+    set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -Og")
```
Luthaf: Do we want to enable this flag by default? As mentioned in #121, this cuts down the runtime of tests to 9s from 147s. The build time goes from 11min (at -O0) to 12min (at -Og).

CMakeLists.txt
```diff
@@ -92,31 +92,23 @@ SET(TYPE_ARCHITECTURE "native" CACHE STRING "Choose the type of architecture to
...
-  # using Visual Studio C++
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Weffc++")
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra")
+
+if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" OR
+    "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR
+    "${CMAKE_CXX_COMPILER_ID}" STREQUAL "AppleClang")
+
+    set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -Og")
+    set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -Og")
```
Luthaf: Do we want to enable this flag by default? As mentioned in #121, this cuts down the runtime of tests to 9s from 147s. The build time goes from 11min (at -O0) to 12min (at -Og).

felixmusil: what is the diff between these two optimization flags ? I have a hard time finding it.

src/external/json.hpp
```diff
@@ -30,6 +30,11 @@ SOFTWARE.
 #ifndef INCLUDE_NLOHMANN_JSON_HPP_
 #define INCLUDE_NLOHMANN_JSON_HPP_
 
+#if defined(__GNUC__) && !defined(__clang__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
+#endif
```
felixmusil: why edit the json header ?

src/representations/calculator_spherical_invariants.hh
```diff
@@ -38,6 +38,8 @@
 #include "rascal_utility.hh"
 #include "math/math_utils.hh"
 
+#include "wigxjpf.h"
+
```
felixmusil: this is an external library so I prefer to import it through the math_interface header.

src/external/json.hpp
Luthaf: Because gcc emits warning when compiling it since it is not longer considered a system header after https://github.com/cosmo-epfl/librascal/pull/165/commits/32ea71755bcf39f8648d7a68950177bafd84638e

CMakeLists.txt
Luthaf: -O0 is the default and don't optimize anything. See https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html for the options enabled by -Og

src/math/CMakeLists.txt
```diff
@@ -1,36 +0,0 @@
...
-# Boston, MA 02111-1307, USA.
-#
-
-
-set (math_SRC
-  ${CMAKE_CURRENT_SOURCE_DIR}/math_utils.cc
-  ${CMAKE_CURRENT_SOURCE_DIR}/math_interface.cc
-  )
-
-target_sources("${LIBRASCAL_NAME}" PRIVATE ${math_SRC})
```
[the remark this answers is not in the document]
Luthaf: This code now lives in https://github.com/cosmo-epfl/librascal/pull/165/files#diff-95e351a3805a1dafa85bf20b81d086e6R33-R42

src/representations/calculator_spherical_invariants.hh
Luthaf: Except that this means the header is included in all compilation units, thus driving the compile time up. But you are right that this should use `#include <wigxjpf.h>` since it is an external library.

src/structure_managers/structure_manager.hh
```diff
@@ -666,7 +666,7 @@ namespace rascal {
     template <typename T, size_t Size, int... Indices>
     decltype(auto) append_array_helper(const std::array<T, Size> & arr, T && t,
                                        std::integer_sequence<int, Indices...>) {
-      return std::array<T, Size + 1>{arr[Indices]..., std::forward<T>(t)};
+      return std::array<T, Size + 1>{{arr[Indices]..., std::forward<T>(t)}};
```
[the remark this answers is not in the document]
Luthaf: sub-object initialization warning with clang-4.

tests/test_adaptor_center_contribution.cc
```diff
@@ -25,7 +25,9 @@
  * Boston, MA 02111-1307, USA.
  */
 
-#include "tests.hh"
+#include <boost/test/unit_test.hpp>
+#include <boost/mpl/list.hpp>
```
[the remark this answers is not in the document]
Luthaf: Same as `math_interface.hh`: is contains things not needed everywhere, which drives the compile time up. Typically, `boost/mpl/list.hpp` is only used by half of the tests.

Also, it means that we have to jump to more headers to find where something is defined.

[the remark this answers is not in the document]
Luthaf: There is even a tool and a best practice based on this: "include what you use" (https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhyIWYU.md)

The core idea is that each file should explicitly `#include` ALL used headers, and only these.

CMakeLists.txt
felixmusil: Seems like `-Og` is meant for this so yes I would use it by default

src/math/CMakeLists.txt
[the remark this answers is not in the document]
felixmusil: I don't understand what you mean.

src/representations/calculator_spherical_invariants.hh
felixmusil: the code it refers to is only used in this header so I though even if it is included it would be ignored. So at most it increases the size of the file to be compiled but not the compilation time. Is my impression wrong ?

felixmusil:
> I also removed most of the always_rebuild target, since https://cmake.org/cmake/help/latest/command/add_custom_target.html is always considered out of date anyway.

I tested and it does not keep the previous behavior:  
make rascal tests, add a file to tests/reference_data or delete files from build/reference_data, make again and the missing/added files are in build/reference_data.

Could you please make sure this behavior is there again ? it is also useful in when working with the python wrappers.

Also what are the drawbacks of not using the 'CMake good practice' since what we had was working properly ?

src/math/CMakeLists.txt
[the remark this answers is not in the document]
Luthaf: The cmake code that adds the files in `src/math/*.cc` as source file to the rascal library target is now in the file mentionned above. So I merged `src/math/CMakeLists.txt` into `src/CMakeLists.txt`

Luthaf:
> I tested and it does not keep the previous behavior:
make rascal tests, add a file to tests/reference_data or delete files from build/reference_data, make again and the missing/added files are in build/reference_data.
>
> Could you please make sure this behavior is there again ? it is also useful in when working with the python wrappers.

The `librascal/build/reference_data` folder is no longer used. Instead, the tests are run from `librascal/tests` directly (see https://github.com/cosmo-epfl/librascal/pull/165/files#diff-3dea97423f83302e647a3bab62bf1157R18-R20) and have access to the files themselves instead of a copy.

> Also what are the drawbacks of not using the 'CMake good practice' since what we had was working properly ?

Here are a few explanations of why we should use modern cmake: https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1

The TL;DR is that the modern version remove hidden state and makes future changes easier, as well as allowing others to use rascal as a dependency in their project more easily.

---

> Why do you not work on the cosmo-epfl version of the repo ? I feel it is quite clunky for me to review this PR compared most others especially that I need to build the library to check a few things.

That's why I added some instructions on how to checkout this PR locally in the first message.

src/representations/calculator_spherical_invariants.hh
Luthaf: Having more code to compile increase compilation time, since the files need to be parsed even if they are not used, which takes up time and compiler memory.

felixmusil:
> That's why I added some instructions on how to checkout this PR locally in the first message.

I followed them and I found it clunky so I would like to know the reasons for this added complexity on the reviewer side.

>The librascal/build/reference_data folder is no longer used. Instead, the tests are run from librascal/tests directly (see https://github.com/cosmo-epfl/librascal/pull/165/files#diff-3dea97423f83302e647a3bab62bf1157R18-R20) and have access to the files themselves instead of a copy.

I see then you should avoid copy if it is not used anymore no ? Also when you work on the python wrappers it is essential that the modified files are present in the build folder after doing make, how is it dealt with in this PR ?

Luthaf:
> I see then you should avoid copy if it is not used anymore no ? Also when you work on the python wrappers it is essential that the modified files are present in the build folder after doing make, how is it dealt with in this PR ?

There is no copy anymore. If you see a reference_data folder in your build directory, it is a leftover from previous builds. The Python wrappers files are still copied from `librascal/bindings/rascal` to `librascal/build/rascal`, and then use the test files in `tests/reference_data`. The tests use PYTHONPATH to get the right version of the binding files.  

> I followed them and I found it clunky so I would like to know the reasons for this added complexity on the reviewer side.

Another, maybe easier way to deal with PR originating from forks is to use a different remote:

```
git remote add luthaf https://github.com/luthaf/librascal
git fetch luthaf
git checkout -b cmake-use-targets luthaf/cmake-use-targets
# profit!
```

felixmusil:
> Another, maybe easier way to deal with PR originating from forks is to use a different remote:

Thanks, that is exactly what I was looking for since the experience is the same as for any other PR.

>There is no copy anymore. If you see a reference_data folder in your build directory, it is a leftover from previous builds. The Python wrappers files are still copied from librascal/bindings/rascal to librascal/build/rascal, and then use the test files in tests/reference_data. The tests use PYTHONPATH to get the right version of the binding files.

You are right that's my mistake. After regenerating the build folder it works as expected and the wrappers are properly updated.

bindings/CMakeLists.txt
```diff
@@ -28,55 +28,43 @@
...
-    OUTPUT always_rebuild
-    COMMAND ${CMAKE_COMMAND} -E echo
+add_custom_target(copy_python_bindings ALL
+    COMMAND ${CMAKE_COMMAND} -E remove_directory ${CMAKE_BINARY_DIR}/rascal
+    COMMAND ${CMAKE_COMMAND} -E copy_directory
+        ${CMAKE_CURRENT_SOURCE_DIR}/rascal ${CMAKE_BINARY_DIR}/rascal
+    COMMAND ${CMAKE_COMMAND} -E copy
+        ${CMAKE_CURRENT_BINARY_DIR}/_rascal.* ${CMAKE_BINARY_DIR}/rascal/lib/
+    DEPENDS _rascal
+    COMMENT "Copying python bindings"
```
felixmusil: I would change this message since the python part of the library is not just about the bindings of c++ features. 

```suggestion
    COMMENT "Copying python bindings and library"
```

Luthaf:
I rebased this on top of master, so that it is ready for another round of review!

Luthaf:
ping @felixmusil @tjunge I think this is ready, could I get a review?

CMakeLists.txt
```diff
@@ -52,71 +52,47 @@ ENDIF()
...
-set(INSTALL_PATH "" CACHE STRING
-  "Path to install the libraries")
+set(INSTALL_PATH "" CACHE STRING "Path to install the libraries")
 
-set(ENABLE_DOC "OFF" CACHE BOOL
-  "If on, a target 'dev_doc' (which builds the documentation) is added.")
-
-set(BUILD_TESTS "OFF" CACHE BOOL
-  "If on, builds the unit tests")
+SET(TYPE_ARCHITECTURE "native" CACHE STRING "Choose the type of architecture to compile to in release mode. Default is to let the compiler decide")
```
mastricker: Could we make a linebreak here?

CMakeLists.txt
```diff
@@ -220,152 +170,63 @@ endif()
...
+include(cpplint)
 # Checking hh/cc files for compliance with coding conventions
 if(CPPLINT_FOUND)
   # register the folders to check for cpp code conventions compliance
   cpplint_add_subdirectory("${CMAKE_SOURCE_DIR}/src" "")
   cpplint_add_subdirectory("${CMAKE_SOURCE_DIR}/tests" "")
   cpplint_add_subdirectory("${CMAKE_SOURCE_DIR}/bindings" "")
-  cpplint_add_subdirectory("${CMAKE_SOURCE_DIR}/examples"
-  "--filter=-build/namespaces")
+  cpplint_add_subdirectory("${CMAKE_SOURCE_DIR}/examples" "--filter=-build/namespaces")
```
mastricker: Any reason for making this one line instead of the break?

CMakeLists.txt
```diff
@@ -220,152 +170,63 @@ endif()
...
+# Check for clang-format
+include(clangformat)
 # Add all hh/cc files for autoformatting with clang-format
 if(CLANG_FORMAT_FOUND)
   # register the folders to apply clang-format with given configuration
   # .clang-format in project root folder
-  clang_format_add_subdirectory("${CMAKE_SOURCE_DIR}/src" "")
-  clang_format_add_subdirectory("${CMAKE_SOURCE_DIR}/tests" "")
-  clang_format_add_subdirectory("${CMAKE_SOURCE_DIR}/bindings" "")
+  clang_format_add_subdirectory("${CMAKE_SOURCE_DIR}/src")
```
mastricker: I didn't know that this is possible. Interesting.

bindings/CMakeLists.txt
```diff
@@ -28,55 +28,43 @@
...
-target_link_libraries(pyRascal PRIVATE "${LIBRASCAL_NAME}")
-# Want to rename the output, so that the python module is called _rascal
-set_target_properties(pyRascal PROPERTIES OUTPUT_NAME _rascal)
-# Sets rule to install the bindings library in the rascal folder
-install(TARGETS pyRascal DESTINATION "${PYTHON_SITE_PACKAGES}/rascal/lib")
+# SYSTEM: use include_directory(SYSTEM)
+# NO_EXTRAS: do not setup LTO (link time optimization) and striping of the
+#            shared library 
+pybind11_add_module(_rascal ${BINDING_SRCS} SYSTEM NO_EXTRAS)
+target_compile_options(_rascal PRIVATE -Werror)
```
mastricker: is `-Wall`included additionally? I'm a little lost.

bindings/CMakeLists.txt
```diff
@@ -28,55 +28,43 @@
...
-  )
-
-
-add_custom_command(
-    OUTPUT always_rebuild
-    COMMAND ${CMAKE_COMMAND} -E echo
+add_custom_target(copy_python_bindings ALL
+    COMMAND ${CMAKE_COMMAND} -E remove_directory ${CMAKE_BINARY_DIR}/rascal
+    COMMAND ${CMAKE_COMMAND} -E copy_directory
+        ${CMAKE_CURRENT_SOURCE_DIR}/rascal ${CMAKE_BINARY_DIR}/rascal
```
mastricker: `CMAKE_CURRENT_SOURCE_DIR` is something like `pwd` right?

src/CMakeLists.txt
```diff
@@ -30,24 +30,23 @@
...
-target_link_libraries("${LIBRASCAL_NAME}" PUBLIC "${WIGXJPF_NAME}")
-
-add_subdirectory(structure_managers)
-add_subdirectory(representations)
-add_subdirectory(utils)
-add_subdirectory(math)
-
-install(TARGETS "${LIBRASCAL_NAME}" DESTINATION lib)
+set(rascal_sources
+    json_io.cc
```
mastricker: Is this convention to set `.cc` files by hand?

src/representations/calculator_spherical_invariants.hh
```diff
@@ -38,6 +38,8 @@
 #include "rascal_utility.hh"
 #include "math/math_utils.hh"
 
+#include "wigxjpf.h"
```
mastricker: Should this not also be something with an `external` folder?

tests/test_adaptor.hh
```diff
@@ -31,7 +31,6 @@
 #ifndef TESTS_TEST_ADAPTOR_HH_
 #define TESTS_TEST_ADAPTOR_HH_
 
-#include "tests.hh"
```
mastricker: Just out of curiosity: How do you check for unused includes?
Seems like a couple of them piled up...

tests/test_adaptor_filter.cc
```diff
@@ -25,11 +25,12 @@
  * Boston, MA 02111-1307, USA.
  */
 
+#include <boost/test/unit_test.hpp>
+
 #include "structure_managers/adaptor_filter.hh"
 // TODO(markus): replace the following line after the branch feat/better_tests
 // is merged into master
```
mastricker: I guess this can go away. Somehow I cannot suggest changes directly any more...

tests/test_adaptor_strict.cc
```diff
@@ -27,7 +27,9 @@
  * Boston, MA 02111-1307, USA.
  */
 
-#include "tests.hh"
+#include <boost/test/unit_test.hpp>
+#include <boost/mpl/list.hpp>
+
```
mastricker: Should _our_ headers not come before the system headers?

tests/test_calculator.cc
```diff
@@ -25,7 +25,9 @@
  * Boston, MA 02111-1307, USA.
  */
 
-#include "tests.hh"
+#include <boost/test/unit_test.hpp>
+#include <boost/mpl/list.hpp>
+
```
mastricker: again ours before outside?

tests/test_kernels.cc
```diff
@@ -27,6 +27,9 @@
  * Boston, MA 02111-1307, USA.
  */
 
+#include <boost/test/unit_test.hpp>
```
mastricker: Ours before external?

tests/test_key_standardisation.cc
```diff
@@ -25,10 +25,11 @@
  * Boston, MA 02111-1307, USA.
  */
 
-#include "tests.hh"
-#include "utils/key_standardisation.hh"
+#include <boost/test/unit_test.hpp>
```
mastricker: ours before external?

see [here](https://github.com/cosmo-epfl/librascal/blob/master/docs/source/dev_guide/coding-convention.rst#names-and-order-of-includes)

tests/test_lattice.cc
```diff
@@ -25,7 +25,7 @@
  * Boston, MA 02111-1307, USA.
  */
 
-#include "tests.hh"
+#include <boost/test/unit_test.hpp>
```
mastricker: Ok from now on, I'm not bothering any more. Ours headers before external

tests/test_properties.cc
```diff
@@ -27,8 +27,11 @@
  * Boston, MA 02111-1307, USA.
  */
 
+#include <boost/test/unit_test.hpp>
 #include "test_properties.hh"
 
+constexpr double TOLERANCE = 1e-10;
```
mastricker: This tolerance is different than the one in the file above. I think it was 1e-14. Should they be the same?
I mean it is so small that we probably do not bother anyways...

CMakeLists.txt
```diff
@@ -52,71 +52,47 @@ ENDIF()
...
-project(rascal)
+message(STATUS "Build type is: ${CMAKE_BUILD_TYPE}")
+
+project(rascal CXX)
 
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 set(BUILD_SHARED_LIBS ON)
-set(RASCAL_PYTHON_MAJOR_VERSION 3)
```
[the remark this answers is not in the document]
Luthaf: This variable was not used anywhere, I don't know what was its intended use

CMakeLists.txt
```diff
@@ -52,71 +52,47 @@ ENDIF()
 set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS
 "Release" "Debug" "RelWithDebInfo" "MinSizeRel")
 
-project(rascal)
+message(STATUS "Build type is: ${CMAKE_BUILD_TYPE}")
+
+project(rascal CXX)
```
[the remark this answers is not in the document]
Luthaf: setting the required languages in [project](https://cmake.org/cmake/help/latest/command/project.html) means that cmake check the C++ compiler early (name, ABI, default flags, etc.)

bindings/CMakeLists.txt
Luthaf: -Wall is included in CMAKE_CXX_FLAGS in the root CMakeLists.txt, but this only sets -Werror for the _rascal target, not all of them (and in particular not on external libraries)

bindings/CMakeLists.txt
Luthaf: roughly. Here `${CMAKE_CURRENT_SOURCE_DIR} == ${CMAKE_SOURCE_DIR}/bindings`

src/CMakeLists.txt
Luthaf: This is how it was already done in the code.

The alternative is to use `file(GLOB ...)`, but this mean that we have to manually re-run cmake when adding/deleting files.

src/representations/calculator_spherical_invariants.hh
Luthaf: json.hh is part of the repository, hence the external folder. This one is downloaded and compiled by cmake, so we don't control its path.

tests/test_adaptor.hh
Luthaf: manually commenting out #include, and checking what compiles ...

There is a tool to do it systematically: https://github.com/include-what-you-use/include-what-you-use

tests/test_adaptor_strict.cc
Luthaf: [google code style](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes) (on which this project is based, right?) is 

>  Related header, C system headers, C++ standard library headers, other libraries' headers, your project's headers.

Luthaf: I think it makes sense: start from the most generic to the more specific. For examples, `<cstddef>` defines size_t/std::size_t, so I would include it before using these types, even in headers from the same project.

tests/test_properties.cc
Luthaf: the previous global tol was set to 1e-12, and this file used it as tol * 100, making it 1e-10

In other files, tol was used as tol/100 making it 1e-14, or directly, so I followed that. 

I guess we could try to unify the tolerance, but this should be part of a separated PR.

tests/test_key_standardisation.cc
Luthaf: ok, missed this. I would rather change the style guide to follow google's, and enforce headers with `include-what-you-use`, but this should be for a separate PR.

CMakeLists.txt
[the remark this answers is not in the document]
Luthaf: > Another question: I could not find the python major version check. Or is that now related to eh PythonLibNew 3-something and therefore ensured.

Yes, this is how the Python's major version is checked

CMakeLists.txt
Luthaf: I like it better :smiley:!  :man_shrugging:

But this flag is actually not needed: the code in examples already uses //NOLINT for using namespace XXX

tests/test_key_standardisation.cc
Luthaf: Also, this should be part of clang-format or at the very least cpplint, instead of wasting reviewer and implementation time.

Luthaf: Done in a6316c0ae6950ae77c4eb78b93c8629faf374709

tests/test_adaptor_filter.cc
Luthaf: done!

src/utils/key_standardisation.hh
```diff
@@ -102,21 +102,6 @@ namespace rascal {
...
-   */
-  template <typename T, size_t Order>
-  std::ostream & operator<<(std::ostream & os,
-                            const KeyStandardisation<T, Order> & index) {
-    os << "(";
-    for (size_t i = 0; i < Order - 1; ++i) {
-      os << index[i] << ", ";
-    }
-    os << index.back() << ")";
-    return os;
```
mastricker: Did this go somewhere else or is it lost from now on?

tests/test_key_standardisation.cc
```diff
@@ -25,12 +25,25 @@
...
+   */
+  template <typename T, size_t Order>
+  std::ostream & operator<<(std::ostream & os,
+                            const KeyStandardisation<T, Order> & index) {
+    os << "(";
+    for (size_t i = 0; i < Order - 1; ++i) {
+      os << index[i] << ", ";
+    }
+    os << index.back() << ")";
+    return os;
```
mastricker: Well ok, here it is....Resolved I guess.

Luthaf: Yes, I putted it where it was needed. This also removes the `#include <iostream>` from inside rascal main headers.

Issue #162, opened 2019-10-16
Move alanine-X.json and crystal_structure.json from examples folder


rosecers:
These two files are necessary for the build tests but not appropriately placed in the file tree.

mastricker:
Is merged.

PR #166, opened 2019-10-17
Corrected files and wrapped unwrapped atoms

mastricker:
This is in reference to issue #162 changing the placement of files.

On the way I needed to wrap the alanine structures, which was not wrapped yet. This was also done for the center-select version of it, which was used in another example.

Also deleted the copy of the files for only the examples. That is debatable, because one might want to to compile the examples, but not the tests.

What do you think? 
Shall I keep the copy for _only_ example files for that case but copy it out of the `reference_data` folder from the tests?


rosecers:
I'm fine with the json files being included in the examples/data folder as long as they are being used in the tutorials or examples. Which ones are they being loaded within?

mastricker:
The copy mechanism I was writing about went like this:

copy example structures from example folder to build folder
the same structures are also present in the test/reference_data and if one compiles with tests on, the data is copied twice.
Now from a user perspective this means

mastricker:
Ok, I went over it again and added the copy of the alanine-center-select. That would be missing if one would not build the tests.

Should be good now?

examples/alanine_chain.cc
```diff
@@ -66,7 +66,7 @@ int main() {
   // manager.update("alanine-X.json");
   auto pair_manager{rascal::make_adapted_manager<rascal::AdaptorNeighbourList>(
       manager, cutoff)};
-  std::string filename{"alanine-X.json"};
+  std::string filename{"reference_data/alanine-X.json"};
```
Luthaf: The file was moved to `examples/data`, but this refers to `reference_data`. I understand why (the file being referenced in the binary is the one in CMAKE_BINARY_DIR), but should the path be unified?

mastricker:
Good catch.
I created it at some point, but it was apparently not used any more as someone probably changed the tests so a smaller structure. I actually wanted to delete it. Did it now.

examples/alanine_chain.cc
mastricker: The compiled example `alanine_chain` resides in the build folder. I changed the file copy from the examples to the `reference_data` folder _within_ the build folder. So when you run the compiled example, it reads from `librascal/build/reference_data/` and the example structures do not globber up the `build/` folder, which I found weird now that I saw it. All examples (and if tests compile) test files should now be in `build/reference_data/`

Luthaf: yes, I understand that. What I mean is that it can be weird for a reader to see the code mentioning `reference_data/alanine-X.json`, and not finding it in `librascal/examples`. 

To fix this we could either change `librascal/tests/reference_data` to `librascal/tests/data` (and thus `librascal/build/reference_data` to `librascal/build/data`), or `librascal/examples/data` to `librascal/examples/reference_data`.

mastricker: Ok, I agree. I did just that.

mastricker:
@rosecers You are reworking also the examples, right? (I admit: I have not looked at that branch for quite a while)
_And just to get this out of the way: the following is not to blame anybody but a suggestion for improvement_
Further comments, so that they do not get lost:

I just noticed that even though there is a file `examples/tutorial_utils.py`, which seems to be a helper-function file, there is quite some overlap in the actual examples  with helper function _inside_ the jupyter notebooks. Just wanted to write this somewhere, but I guess you noticed this already?

And further - that file `tutorial_utils.py` is in a form which I would not like to show people as an example...long lines, commented out code, etc. pp.
We have agreed on pep8 convention for python and that file is a hacked together amalgam of stuff which in it's current for is ok for being on somebody's computer, but not up to the standard I would like to have in a public library.

Just wanted to put this somewhere.

max-veit:
> I just noticed that even though there is a file `examples/tutorial_utils.py`, which seems to be a helper-function file, there is quite some overlap in the actual examples with helper function _inside_ the jupyter notebooks. Just wanted to write this somewhere, but I guess you noticed this already?

Have you read the tutorials that use this file? The point is to first show a new user what the code can do and let them play with the parameters, without first bothering them with the dirty details. Then they can look into the code in the same notebook if they're interested.

> And further - that file `tutorial_utils.py` is in a form which I would not like to show people as an example...long lines, commented out code, etc. pp.
> We have agreed on pep8 convention for python and that file is a hacked together amalgam of stuff which in it's current for is ok for being on somebody's computer, but not up to the standard I would like to have in a public library.

It was merged in an intermediate state because it was included in the documentation update PR before the work on the tutorials split off into its own branch. @rosecers has been reworking the tutorials significantly, including this file, so rest assured it's been noted.

> Just wanted to put this somewhere.

But why here? What does it have to do with the content of this pull request? If you really need to put these comments somewhere where all the developers can see them, either post it in the developer channel or open an issue.

mastricker:
> > I just noticed that even though there is a file `examples/tutorial_utils.py`, which seems to be a helper-function file, there is quite some overlap in the actual examples with helper function _inside_ the jupyter notebooks. Just wanted to write this somewhere, but I guess you noticed this already?
> 
> Have you read the tutorials that use this file? The point is to first show a new user what the code can do and let them play with the parameters, without first bothering them with the dirty details. Then they can look into the code in the same notebook if they're interested.
> 
I think you misunderstand me: I am not advocating against the use of that file. I am just noting that I am not happy with the current form.

> > And further - that file `tutorial_utils.py` is in a form which I would not like to show people as an example...long lines, commented out code, etc. pp.
> > We have agreed on pep8 convention for python and that file is a hacked together amalgam of stuff which in it's current for is ok for being on somebody's computer, but not up to the standard I would like to have in a public library.
> 
> It was merged in an intermediate state because it was included in the documentation update PR before the work on the tutorials split off into its own branch. @rosecers has been reworking the tutorials significantly, including this file, so rest assured it's been noted.
> 
Yes I can see that but there was no apparent explanation for why it looked likes this.

> > Just wanted to put this somewhere.
> 
> But why here? What does it have to do with the content of this pull request? If you really need to put these comments somewhere where all the developers can see them, either post it in the developer channel or open an issue.

I wanted to point to the state I saw it in, the state my comment refers to. That is why I added @rosecers to my comment. She was the only one I asked. No need for everybody to see it.
And the thing I hope would happen happened: It was confirmed that it has already been noted, although not by her :)
I will make an issue next time...

mastricker:
Another thing: while I'm at it:
We have a duplication of some of the examples structures.
Since this potentially leads to a divergence of the respective files, I suggest one place for them: `tests/reference_data`.
`CMakelists` commands have to change of course and the relative paths also (I would also change the paths in the ipython notebooks.

Any objections @rosecers - since you are the one working on the examples/tutorials. Would that be a good solution for avoiding diver

mastricker:
I think this can be merged?
Maybe should be even merged before the merge of PR #165, since partially similar things have happened and this one is much smaller.

Can I get a approving review - anyone?

I put up a bullet point for discussion for the next meeting for the placement/structuring of reference and example data. Since some structures exist in two places (possible maintenance nightmare).

PR #172, opened 2019-10-21
Feat/tutorials

rosecers:
This is a draft PR to start merging tutorials framework into master branch

Rationale and detailed description of the changes:

- Added splashpage within documentation
- Wrote Learning tutorial
- Outlined tentative tutorials and tutorial maps

Tasks before review:

- [X] Add tests, examples, and complete documentation of any added functionality
    - [ ] Make sure the autogenerated documentation appears in Sphinx and is
          formatted correctly (ask @max-veit if you need help with this task).
    - [ ] Write additional documentation in the Sphinx (not docstrings!) to
          explain the feature and its usage in plain English
    - [X] Make sure the examples run (!) and produce the expected output
    - [ ] For bugfix pull requests, list here which tests have been added or re-enabled
          to verify the fix and catch future regressions as well as similar bugs
          elsewhere in the code.
- [X] Run `make lint` on the project, ensure it passes
    - [X] Run `make pretty-cpp` and `make pretty-python`, check that the
          auto-formatting is sensible
    - [X] Review variable names, make sure they're descriptive (ask a friend)
    - [X] Review all TODOs, convert to issues wherever possible
    - [ ] Make sure draft, in-progress, and commented-out code is moved to its
          own branch
- [X] If committing any code in Jupyter/IPython notebooks, install and run `nbstripout`
- [X] Merge with master and resolve any conflicts
- [X] (anything else that's still in progress)

Remaining tasks, out of scope of this pull request: (optional)

- Continue writing and receiving feedback on tutorials

PR #175, opened 2019-10-22
Fixes #174

### LINKED ISSUE #174: Representation target_type is wrongly documented and caught
QUICKFIX for all: please use target_type "Atom" or "Structure" when using the kernel model from rascal, not as the python doc suggests "atomic" or "structure". This is only essential for you if you want to train with "Atom", because it will switch to Structure automatically.

Representation target_type is wrongly documented on the python side and wrongly caught in the c++ side.
- No error occurs if target_type is wrong
https://github.com/cosmo-epfl/librascal/blob/9106a32cd1a2922354515e294f57b98e7bc80930/src/models/kernels.hh#L180-L188
- documentation is wrong, parameter is wrong https://github.com/cosmo-epfl/librascal/blob/9106a32cd1a2922354515e294f57b98e7bc80930/bindings/rascal/models/kernels.py#L10-L28

There is a switch case at https://github.com/cosmo-epfl/librascal/blob/9106a32cd1a2922354515e294f57b98e7bc80930/src/models/kernels.hh#L223-L237 fails because the empty type is by default the first type which is Structure.

agoscinski:
described here https://github.com/cosmo-epfl/librascal/issues/174

Tasks before review:

- [x] Add tests, have to think how to test this
- [x] Add correct description into the docs


bindings/rascal/models/kernels.py
```diff
@@ -23,7 +23,7 @@ def __init__(self, representation, name='Cosine', target_type='structure',
             (currently) only option
 
         target_type : string
-            Type of target (prediction) properties, either 'atomic' (kernel is
+            Type of target (prediction) properties, either 'Atom' (kernel is
             between atomic environments) or 'structure' (kernel is summed over
```
Luthaf: 'structure' => 'Structure'

felixmusil:
I am good with this fix except for the notebook that shows too many modifications for what this PR is supposed to do right ?

examples/SOAP_example.ipynb
```diff
@@ -4,78 +4,282 @@
...
+    "init_cell": true
+   },
+   "outputs": [],
+   "source": [
+    "%matplotlib notebook\n",
+    "from matplotlib import pylab as plt\n",
+    "\n",
+    "import os, sys\n",
+    "from ase.io import read\n",
+    "sys.path.insert(0,\"../build/\")\n",
```
max-veit: This whole notebook is out of date and will be removed in #172 . Please revert all notebook changes; any example code should go in a small Python file (as per the README and the contribution guidelines).

max-veit:
Oh, and adding documentation to make sure this issue doesn't pop up again is another important point that was included in the PR template; please don't delete these points in the future.

examples/SOAP_example.ipynb
agoscinski: sorry this notebook was not intended to be commited, I ll reset it.

agoscinski:
> Oh, and adding documentation to make sure this issue doesn't pop up again is another important point that was included in the PR template; please don't delete these points in the future.

Where should this be added into the doc? It is in the python file of the, if you type `help(Kernel)` in python you see it and I am working on a test in the c++ and python side.

Sorry, I actually skimmed through all the points and added the one which seemed relevant. I can keep the template next time.

max-veit:
> Where should this be added into the doc? It is in the python file of the, if you type `help(Kernel)` in python you see it and I am working on a test in the c++ and python side.

Sure, but have you made sure it appears in the Sphinx documentation as well? The Python API is being documented in `docs/source/reference/python.rst`; it should be as simple as adding an [`autoclass`](http://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autoclass) in the appropriate section (or a new section if you think that's a better fit) and checking that the docstrings are formatted/rendered properly.

Also, you mentioned in the issue that something was "wrongly caught" on the C++ side; what exactly does that mean? Can you think of an underlying cause for why the error was silently ignored before (silent defaults), and perhaps caution against this pattern in the future? This sort of warning might be better placed in the developer docs, perhaps in one of the how-to documents.

tests/python/python_models_test.py
```diff
@@ -0,0 +1,49 @@
...
+        for target_type in ["Atom", "Structure"]:
+            cosine_kernel = Kernel(rep, name="Cosine", target_type=target_type, zeta=2)
+            cosine_kernel(features)
+
+        # wrong name 
+        with self.assertRaises(RuntimeError):
+            Kernel(rep, name="WrongName", target_type="Structure", zeta=2)
+        # wrong target_type
+        with self.assertRaises(RuntimeError):
+            Kernel(rep, name="Cosine", target_type="WrongType", zeta=2)
```
max-veit: how about adding the wrong case (like "cosine" or "structure") to the list of assertRaises constructors?

tests/test_kernels.cc
```diff
@@ -35,6 +35,20 @@ namespace rascal {
   using multiple_ref_fixtures =
       boost::mpl::list<KernelFixture<DataSphericalInvariantsKernelFixture>>;
 
+  /**
+   * Tests if the wrong target_type is catched correctly.
+   */
+  BOOST_FIXTURE_TEST_CASE(kernel_target_type_test,
+                          KernelFixture<DataSphericalInvariantsKernelFixture>) {
+    auto kernel_hyper = this->ParentA::kernel_hypers.at(0);
+    kernel_hyper["target_type"] = "this_target_type_does_not_exist";
```
max-veit: Again, how about a case with `kernel_hyper["target_type"] = "structure"` to show that it fails?

PR #176, opened 2019-10-23
The wednesday cleanup PR

Luthaf:
More misc cleanup, ~based on top of #165, which should be merged first~.

<sub>I removed all of the points for the PR checklist, since they don't apply here.</sub>


src/math/bessel.hh
```diff
@@ -199,15 +199,6 @@ namespace rascal {
         if (n_up > 0) {
           this->upward_recursion(distance, fac_a, n_up);
         }
-
-        // set small values to 0.
```
Luthaf: this was not doing anything, since unaryExpr does not modify data inplace.

Luthaf:
This is best reviewed on a commit-by-commit basis, since they should all be independent.

src/models/kernels.hh
```diff
@@ -221,19 +220,14 @@ namespace rascal {
...
-        break;
-      }
-      case TargetType::Atom: {
+      case TargetType::Atom:
         return this->compute_helper<Property_t, TargetType::Atom>(
             representation_name, managers_a, managers_b);
-        break;
-      }
       default:
         throw std::logic_error("The combination of parameter is not handdled.");
```
max-veit: Let's fix the spelling while we're at it:
```suggestion
        throw std::logic_error("The combination of parameter is not handled.");
```
Perhaps you could do a global `s/handdled/handled/`, since I think this line has been copy-pasted around quite a bit. But furthermore, it's an error message that should never be reached, so its utility is questionable -- I think this should be more along the lines of https://xkcd.com/2200/ -- implemented e.g. here: https://github.com/cosmo-epfl/librascal/blob/deb7a09bed3e8713267a364cd742687eb28aaa8c/src/representations/calculator_spherical_expansion.hh#L1028-L1039

src/representations/calculator_spherical_covariants.hh
```diff
@@ -53,138 +53,105 @@ namespace rascal {
...
-        // get the number of non zero elements in the w3j
-        // Use the well-known selection rules for the Wigner 3j symbols
-        int n_elements{0};
-        size_t l3{this->lambda};
-        for (size_t l1{0}; l1 < this->max_angular + 1; ++l1) {
-          for (size_t l2{0}; l2 < this->max_angular + 1; ++l2) {
-            // Triangle rule
-            if (l1 < static_cast<size_t>(std::abs<int>(l2 - l3)) ||
-                l1 > l2 + l3) {
+    inline Eigen::ArrayXd precompute_wigner_3js(size_t max_angular,
```
max-veit: Is the `inline` really useful? I know it's been peppered throughout the code, but I thought compilers mostly ignored inlining hints nowadays.

src/math/math_utils.hh
```diff
@@ -49,31 +49,32 @@ namespace rascal {
...
-    // Reminder: C++ floating-point literals are automatically of type double
-    /// Pi to more digits than anyone could possibly need
-    const double PI = 3.14159265358979323846264338327950288419716939937510;
-    const double SQRT_PI =
-        1.7724538509055160272981674833411451827975494561223871282;
-    const double SQRT_TWO = std::sqrt(2.0);
-    const double INV_SQRT_TWO = std::sqrt(0.5);
-    const double SQRT_THREE = std::sqrt(3.0);
+    /// π to more digits than anyone could possibly need
+    constexpr double PI = 3.14159265358979323846264338327950288;
```
max-veit: Why use `constexpr` instead of `const`? Is there a good reason besides making me have to think harder about all the different ways to qualify variables in C++?

Luthaf:
> As for the simplification of the Wigner 3j initialization, you'll have to discuss that with @felixmusil since it's a fairly large change to a quite complex piece of code that he wrote.

The code did not change, I just transformed it from class + std::array<std::shared_ptr> to functions and storing the initialization data inline.

src/math/math_utils.hh
Luthaf: `constexpr` means that the value is (1) const and (2) known at compile time. This means the compiler is able to do constant folding; which it might not do with `std::sqrt(3.0)`, since the value is created at compile-time.

src/representations/calculator_spherical_covariants.hh
Luthaf: Here yes, but not for code inlining. `inline` is a way around the One Definition Rule. Since the file is included and compiled in multiple difference compilation units, the linker then have multiples copies of the function to work with. `inline` tells the linker they are all equivalent and that it should use any of them.

src/models/kernels.hh
```diff
@@ -95,7 +94,7 @@ namespace rascal {
                     representation_name)};
 
             kernel(ii_A, ii_B) =
-                propA.dot(propB).unaryExpr(integer_power).mean();
+                propA.dot(propB).array().pow(this->zeta).mean();
```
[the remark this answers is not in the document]
Luthaf: Eigen pow is also built-in for integer types: https://eigen.tuxfamily.org/dox/group__CoeffwiseMathFunctions.html

src/models/kernels.hh
Luthaf: Looks like it is only written `handdled` in this file. 

> But furthermore, it's an error message that should never be reached, so its utility is questionable

We could have an unreachable macro/function for this, to be used everywhere. But I'd rather leave this for a future PR. Do you want me to open an issue?

src/representations/calculator_spherical_covariants.hh
Luthaf: I removed most of the other inline qualifiers in #150

max-veit: ergh, ok, leave it to C++ to give you a keyword that does something only vaguely related to what it says and requires an advanced degree in compiler design to actually understand.

max-veit:
> The code did not change

Oh yes it did. Maybe the end effects didn't change, but it's still going to be disorienting for the next person who has to modify or use this particular piece of code. Perhaps a more detailed description of that particular change (where the code has been moved, how precomputation works now vs. where it was done before) is needed in the PR summary?

Luthaf:
> Oh yes it did

Ok, my bad. The inside of the function did not change though.

Concerning "how precomputation works now", the answer is "more explicitly" (see calls to `internal::precompute_XXX`):

https://github.com/cosmo-epfl/librascal/blob/15c377caa5757b1eca4e29f79049a1eac48c2d57/src/representations/calculator_spherical_invariants.hh#L210-L222

> Perhaps a more detailed description of that particular change (where the code has been moved, how precomputation works now vs. where it was done before) is needed in the PR summary?

I can do that, but it might be better off as a comment inside the code describing the pre-computation. Although I think the new version is pretty-much self-explanatory.

src/models/kernels.hh
max-veit: Yeah why not

src/math/math_utils.hh
max-veit: But is there a difference if it's just a plain old numeric literal, as it is here?

Luthaf: This should not be different when considering optimization, but shows intent better.

And talking about qualifiers explosions, these should actually be `inline constexpr` (https://stackoverflow.com/a/54467111)

src/models/kernels.hh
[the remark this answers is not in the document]
felixmusil: unfortunately not. it works with integer / integer pow which is not the case here so it cast the int to double. Please put back the intended behavior (integer division).

src/representations/calculator_spherical_expansion.hh
```diff
@@ -1046,21 +1044,14 @@ namespace rascal {
     using internal::AtomicSmearingType;
     using internal::RadialBasisType;
 
-    switch (internal::combineEnums(this->radial_integral_type,
-                                   this->atomic_smearing_type)) {
-    case internal::combineEnums(RadialBasisType::GTO,
-                                AtomicSmearingType::Constant): {
+    assert(this->atomic_smearing_type == AtomicSmearingType::Constant);
+    if (this->radial_integral_type == RadialBasisType::GTO) {
```
felixmusil: How would you deal with a combination of 3, 4 .... template arguments that can't be all mixed together ? Also why go for if statement instead of switch ?

src/representations/calculator_spherical_invariants.hh
```diff
@@ -401,18 +318,19 @@ namespace rascal {
...
+
+    // precomputed wigner symbols for the BiSpectrum
+    Eigen::ArrayXd wigner_w3js{};
   };
 
   template <class StructureManager>
   void CalculatorSphericalInvariants::compute(StructureManager & managers) {
     using internal::SphericalInvariantsType;
-    switch (this->spherical_invariants_type) {
+    switch (this->type) {
```
felixmusil: I prefer the verbose name.

src/atomic_structure.hh
```diff
@@ -122,12 +122,16 @@ namespace rascal {
...
       for (int i_dim{0}; i_dim < Dim; ++i_dim) {
         if (this->pbc[i_dim]) {
           scaled_positions.row(i_dim) =
-              scaled_positions.row(i_dim).unaryExpr(functor);
+              scaled_positions.row(i_dim).unaryExpr([](double d) {
+                // Modulo that follows python standard, i.e. (-3) % 5 == 2
+                // (python) and not (-3) % 5 == -3 (C++).
+                auto m = std::fmod(d, 1.0);
+                return m + (m < 0 ? 1.0 : 0);
+              });
```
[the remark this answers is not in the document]
Luthaf: > This operation is only used once now but we might need it

Then a new functor should be created if there is really a need to abstract this particular operation. But moving the implementation close to the place of use mean there is less jumping around file to understand a particular piece of code.

> Also it is the kind of mind blow (difference in behavior for the mod between cpp and python) that I would like to keep very explicit, like having a dedicated function.

I think having the comment inline is the more explicit this can be.

src/math/bessel.hh
Luthaf: Removing this introduce no changes in behaviour. If you want to make sure the data is modified, this should be a separate PR, to be discussed separately =)

src/models/kernels.hh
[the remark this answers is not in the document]
Luthaf: > Please put back the intended behavior (integer power).

I don't see what behavior change you are talking about here

src/representations/calculator_spherical_expansion.hh
Luthaf: > How would you deal with a combination of 3, 4 .... template arguments that can't be all mixed together ?

```
if (a == TypeA::FOO && b == TypeB::BAR) {
}
```

> Also why go for if statement instead of switch ?

Because they allow to use && and || to combine values. Also, they naturally introduce a new scope, which mean we can define new variables and so on.

src/representations/calculator_spherical_invariants.hh
Luthaf: I don't see the point in repeating the local context: we are in the sperical_invariant class, and checking the value against `SphericalInvariantsType::XXX`, so I find this name clear enough.

src/models/kernels.hh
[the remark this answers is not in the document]
Luthaf: If you are worried about a potential performance loss (which I would like to see demonstrated), we can also use [square](https://eigen.tuxfamily.org/dox/classEigen_1_1ArrayBase.html#a4fe59f7f07a8b445e47b0a6d9f1c15a1) and `cube` functions.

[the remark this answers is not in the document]
felixmusil: There is no question about the efficiency difference between exp + log compared to a few multiplications. 
The previous behaviour is to make sure integer power is used (which is not the case anymore).

src/math/bessel.hh
felixmusil: It is a one line change so please undo the delete at the very least.

src/models/kernels.hh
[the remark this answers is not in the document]
Luthaf: > There is no question about the efficiency difference between exp + log compared to a few multiplications.

std::pow is usually not implemented as exp + log(see the implementation of musl here https://git.musl-libc.org/cgit/musl/tree/src/math/pow.c), and not necessary slower than repeated multiplication: https://martin-ueding.de/articles/efficiency-of-pow-function/index.html.

When talking about performances, better to measure twice. Here, since `unaryExpr` is opaque to Eigen optimisations, this might also be the case, but again should be benchmarked (I'll set something up tomorrow).

And citing Knuth, "premature optimization is the root of all evil": I'd rather optimize for readability and further modifications than gut-feeling performances. This way, when I want to to optimize for performances, I can more easily see what is going on and check that my changes had the intended effects.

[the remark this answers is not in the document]
felixmusil: your link refers to a comparison for power 2 which is often a special case in the pow implementation and the gains are compiler dependents. 
Before removing the usage of std::pow in spherical expansion it used to take up to 20% of runtime (benchmarked with vtune) and after the integer pow function was less than 2% of runtime. 
I agree this is not a polished benchmark but to me this is good enough to warrant not using std::pow when possible.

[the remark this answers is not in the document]
Luthaf: Here is a micro benchmark, where the power is given as a command line argument to make it invisible to the optimizer.: https://gist.github.com/Luthaf/6170745dd6eb6654699dbb033e5ac6a8

The TL;DR is `Eigen::pow == std::pow > functor == explicit for loop > Eigen::square/Eigen::cube`

If performance is really an issue, then the best solution is to explicitly use `square/cube`, and fallback to an explicit for-loop for higher values of zeta.

But again, I am not convinced that this operation is worth optimizing too much: I think the memory accesses to get the properties would take much more time that this (would have to be benchmarked again).

src/math/bessel.hh
Luthaf: I can undo the delete if someone takes care of it in the next days/weeks. Else it will stay like this for god know how much longer, inducing confusion in readers.

Do you want to take care of this change?

felixmusil: yes please

src/models/kernels.hh
Luthaf: see #185 for a general fix to this.

src/math/math_utils.hh
Luthaf: actually, they can't be inline constexpr, since that's a c++17 extension ¯\_(ツ)_/¯

Luthaf:
Should be good now!

src/models/kernels.hh
```diff
@@ -134,22 +133,48 @@ namespace rascal {
...
+        } else if (this->zeta == 2) {
+          kernel = kernel.array().square();
+        } else if (this->zeta == 3) {
+          kernel = kernel.array().cube();
+        } else {
+          kernel = kernel.unaryExpr([zeta = this->zeta](double v) {
+            double res = 1;
+            for (size_t i = 0; i < zeta; i++) {
+              res *= v;
+            }
```
felixmusil: please use the implementation of pow in math_utils since it is there and more efficient.

Luthaf: > and more efficient.

Not what I see in the benchmark, they are roughly equivalent.

I can still use it to diminish code duplication

ceriottm:
Just to chime in - the integer power was really making a very noticeable
difference in the final wall-clock performance of spherical expansion.
I don't know why eigen doesn't implement a decent integer pow, but it
really is much slower in all the tests we ran.
Glad to see otherwise, but pls check what would be the impact on some
real-life tests.

Luthaf:
> Just to chime in - the integer power was really making a very noticeable
difference in the final wall-clock performance of spherical expansion.
I don't know why eigen doesn't implement a decent integer pow, but it
really is much slower in all the tests we ran.
Glad to see otherwise, but pls check what would be the impact on some
real-life tests.

Two things here: 
- I reverted this change, after running a few microbenchmarks (https://gist.github.com/Luthaf/6170745dd6eb6654699dbb033e5ac6a8) showing that Eigen::pow is roughly the same as std::pow.
- Even if pow took a significant part of SphericalHarmonics computation, it does not mean it will take a significant amount of the kernel computation. This should be benchmarked separately.

src/models/kernels.hh
Luthaf: done!

Luthaf:
> How can I commit to this branch ?

add a remote, checkout the branch, commit and push !

src/models/kernels.hh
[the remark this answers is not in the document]
felixmusil: it makes sense since Eigen::square/Eigen::cube have sse2/avx instructions. the functor is to avoid the for loop and if the performance are important it can be written with vectorized instructions.

PR #179, opened 2019-10-23
Fix periodic soap gradients

### LINKED ISSUE #115: Gradient implementation does not correctly handle periodic images
In a periodic system with the environment cutoff larger than one of the cell dimensions (and _not_ using the minimum-image convention), the gradient of a property with respect to the motion of some centre must also account for the motion of its periodic images within that cutoff. This is not currently done, as all atoms in the neighbourhood of a centre are summed over in the gradient computation. As the drawing below illustrates, the correct gradient simply omits the contributions from those neighbours that are periodic images of the central atom, as the distance between those images does not change with motion of the central atom.

IMG_2526

The fix is simple: We just need a test for each atom in the neighbourhood to see if it's an image of the central atom and, if so, skip its "contribution" to the gradient.

max-veit:
Fix #115 so that SOAP gradients are correct for periodic systems

- For each gradient computed with respect to a pair distance (in SphericalExpansion), make sure that pair isn't between two periodic images of a same atom (if it is, then leave it at zero)
- Change the swap_pair_ref() function to handle multiple periodic images of the i-atom (first atom of the input pair), because all of these images contribute to the gradient of j with respect to $r_i$
- Make sure to accumulate gradient contributions only onto real (not ghost) atoms (only for the gradient test)

Tasks before review:

          elsewhere in the code.
      - Gradient test now includes true periodic structures (by leaving the periodic flags on)
      - Explicitly non-periodic structures are also now tested, to protect against regressions
          auto-formatting is sensible
          own branch

Out of scope of the current PR:

- Some gradient tests failing due to #182, will be fixed in a separate PR (with the failing tests to be disabled for now) (Update: fixed)


max-veit:
Oh great, the crashes are back (and only in release mode, not RelWithDebugInfo). How are we supposed to debug this?!

Luthaf:
> Oh great, the crashes are back (and only in release mode, not RelWithDebugInfo). How are we supposed to debug this?!

Are you talking about crashes locally? If so, you can edit `CMAKE_CXX_RELEASE_FLAGS` to append '-g' (or `CMAKE_CXX_RELWITHDEBINFO_FLAGS` to use -O3 instead of -O2)

max-veit:
Oh great, now it's only failing on gcc. What sort of optimization could it be doing to break things so badly?

Luthaf:
Usually, if optimizations break code, that means that the code is relying on undefined behaviour somewhere. Another tool to try to pinpoint what is going on are the sanitizers (asan & family).

I would try compiling and running the tests with `-fsanitize=address`, and then maybe `-fsanitize=memory` & `-fsanitize=undefined`. These flags needs to be added to `CMAKE_CXX_FLAGS` and `CMAKE_EXE_LINKER_FLAGS`. They are all supported by the latest clang, and some of them are also available with recent gcc.

See https://clang.llvm.org/docs/AddressSanitizer.html, https://clang.llvm.org/docs/MemorySanitizer.html, https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html.

Luthaf:
> BTW valgrind shows no memory errors on RelWithDebugInfo...

What about CMAKE_BUILD_TYPE=Release ?

max-veit:
Okay, I can see why not wrapping the atoms after displacement (especially for largish displacements, e.g. at the beginning of the gradient test loop) could be a problem, but why was this only causing an error in release mode (not even RelWithDebugInfo)?

And why are we getting these weird linker errors now? I hope it has nothing to do with the fix proper...

Luthaf:
The linker errors seems related to a01317e14e3a63efe3bb558a44981134e8d9e144

tests/test_calculator.hh
```diff
@@ -588,9 +588,10 @@ namespace rascal {
   template <typename RepManager, class StructureManager>
   class RepresentationManagerGradientCalculator {
    public:
-    using Structure_t = AtomicStructure<3>;
+    static const int Dim = StructureManager::traits::Dim;
```
Luthaf: Maybe this should be a `static constexpr int`

max-veit:
Oh great. Can't we just assume three-dimensional atomic structures for the tests? Seems like 2-D structures are enough of a special case that they would warrant their own testing class (if necessary, we make it a default template parameter of the gradient provider). Otherwise this just seems like another case of unnecessary generalization.

Luthaf:
>  Otherwise this just seems like another case of unnecessary generalization.

I agree with you here, the Dim parameter could be completely removed. Since the code is only ever tested with Dim==3 anyway, and the expressions for spherical harmonics would be completely different for other dimensions, the code is very much unusable for other dimension.

docs/source/reference/cpp.rst
```diff
@@ -67,13 +69,44 @@ below, excerpted from :file:`tests/test_calculator.cc`:
...
+
+where :cpp:var:`representations.back()` is a
+:cpp:class:`RepresentationCalculator`,
+:cpp:var:`manager` is a :cpp:class:`StructureManager`, and
+:cpp:var:`structures.back()` is the :cpp:class:`AtomicStructure` associated with
+that manager.
+
+A more detailed documentation of these two classes follows:
+
+.. doxygenclass:: rascal::RepresentationManagerGradientCalculator
```
Luthaf: small question: should these live in the rascal::test namespace?

max-veit: Didn't know that existed. Maybe, but that would require wrapping basically the whole file (`test_calculator.{cc,hh}`, as well as `test_math.hh` and everything it references) in a `namespace math` declaration.

felixmusil:
> I agree with you here, the Dim parameter could be completely removed. Since the code is only ever tested with Dim==3 anyway, and the expressions for spherical harmonics would be completely different for other dimensions, the code is very much unusable for other dimension.

I agree that you might not want to get Dim from the manager but I would still not remove it to make it clear and explicit in the code that the 3 that are hanging around refer to the dimension and not some other quirks. So I would keep:
`static constexpr int Dim = 3;`

docs/source/reference/cpp.rst
```diff
@@ -48,10 +48,65 @@ Math utilities (namespace :cpp:class:`rascal::math`)
...
+To check the gradient of a new representation calculator, it should suffice to
+use the classes RepresentationCalculatorGradientProvider (to provide the function
+and its gradient) and RepresentationCalculatorGradientFixture (to assist in
+iterating over the atoms of the structure).  An example of its usage is shown
+below, excerpted from :file:`tests/test_calculator.cc`:
+
+.. literalinclude:: ../../../tests/test_calculator.cc
+   :language: cpp
+   :lines: 430-436, 441-444
+   :dedent: 4
```
felixmusil: you might want to use the following keywords instead of lines to avoid shifts in the future ? 
```
:start-after: my-start
:end-before: my-end
```
You can insert my-start into the source code at the proper line  like this and the same for my-end
```
/* -------------------- my-start -------------------- */
````

File: tests/test_calculator.hh [hunk 4]
```diff
@@ -715,72 +748,118 @@ namespace rascal {
...
-      size_t access_index = structure_manager->get_atom_index(pair_ref.back());
+      size_t access_index{structure_manager->get_atom_index(pair_ref.back())};
       auto new_center_it{structure_manager->get_iterator_at(access_index)};
       // Return cluster ref at which the iterator is currently pointing
       auto && new_center{*new_center_it};
+      // Get the position of the original central atom (to find all its periodic
+      // images)
+      size_t i_index{structure_manager->get_atom_index(pair_ref.front())};
+      auto i_position{structure_manager->get_position(i_index)};
+      // std::cout << "cutoff: " << structure_manager->get_cutoff()<<std::endl;
```
felixmusil:
```suggestion
     
```

tests/test_calculator.hh
```diff
@@ -715,72 +748,118 @@ namespace rascal {
...
       // Iterate until (j,i) is found
+      std::vector<PairRefKey_t> new_pairs;
+      using Positions_t = Eigen::Matrix<double, 3, Eigen::Dynamic>;
       for (auto new_pair : new_center) {
-        if (new_pair.back() == pair_ref.front()) {
-          return new_pair;
+        Positions_t i_trial_position = new_pair.get_position();
```
felixmusil:
```suggestion
        // go through the neighbors of atom j and wrap them into the cell 
        // to test if they match atom i
        Positions_t i_trial_position = new_pair.get_position();
```

docs/source/reference/cpp.rst
max-veit: Ok, it clutters the source code a bit but I suppose that's tolerable.

tests/test_calculator.hh
```diff
@@ -637,8 +650,15 @@ namespace rascal {
       // Count all the keys in the sparse gradient structure where the gradient
       // is nonzero (i.e. where the key has an entry in the structure)
       for (auto neigh : center) {
+        // TODO(max,felix) there should be a way to write "if neigh is ghost"
```
max-veit: What do we want to do about this TODO? Convert it to an issue, or is it a simple fix to implement this function (`is_ghost()`) in StructureManager?

tests/test_calculator.hh
```diff
@@ -715,72 +748,120 @@ namespace rascal {
...
+      using Positions_t = Eigen::Matrix<double, 3, Eigen::Dynamic>;
       for (auto new_pair : new_center) {
-        if (new_pair.back() == pair_ref.front()) {
-          return new_pair;
+        // go through the neighbors of atom j and wrap them into the cell 
+        // to test if they match atom i
+        Positions_t i_trial_position = new_pair.get_position();
+        Positions_t i_wrapped_position =
+            atomic_structure.wrap_explicit_positions(i_trial_position);
+        if ((i_wrapped_position - i_position).norm() < image_pos_tol) {
```
max-veit: Is it possible to do this with comparing the `get_index()` of the two atoms (to see if they're images of the same atom) rather than comparing wrapped positions?

docs/source/reference/cpp.rst
Luthaf: > Didn't know that existed.

It does not, the question was about creating it. And yeah on the wrapping, althought we could `using namespace rascal::test; using namespace rascal::math;` in the .cc file.

tests/test_calculator.hh
max-veit: turns out the answer is "yes"

tests/reference_data/methane.json
```diff
@@ -1,11 +1,11 @@
 {"1": {
- "cell": [[0.0, 0.0, 0.0], [0.0, 0.0, 0.0], [0.0, 0.0, 0.0]],
- "ctime": 18.923226180085212,
- "mtime": 18.923226180085212,
+ "cell": [[20.0, 0.0, 0.0], [0.0, 20.0, 0.0], [0.0, 0.0, 20.0]],
+ "ctime": 19.826526159217163,
```
Luthaf: changes here are annoying, as well as in the unique_id & user fields. Anyway to leave these out?

tests/test_calculator.hh
```diff
@@ -416,12 +417,16 @@ namespace rascal {
...
         "reference_data/diamond_2atom_distorted.json",
         "reference_data/diamond_cubic_distorted.json",
-        "reference_data/SiC_moissanite.json",
         "reference_data/SiCGe_wurtzite_like.json",
-        "reference_data/SiC_moissanite_supercell.json"};
+        "reference_data/SiC_moissanite_supercell.json",
+        "reference_data/small_molecule.json",
+        "reference_data/methane.json"
+    };
+    // Simpler structures for debugging:
```
Luthaf: nice 👍

tests/test_calculator.hh
```diff
@@ -574,51 +579,59 @@ namespace rascal {
...
    * Initialize with a Calculator, a StructureManager, and an
    * AtomicStructure representing the original structure (before modifying with
-   * finite-difference displacments).  The gradient of the representation with
+   * finite-difference displacements).  The gradient of the representation with
    * respect to the center position can then be tested, as usual, with
    * test_gradients() (defined in test_math.hh).
    */
-  template <typename RepManager, class StructureManager>
-  class RepresentationManagerGradientCalculator {
+  template <typename RepCalculator, class StructureManager>
```
Luthaf: I think this parameter is called `Calculator` in other places

tests/test_calculator.hh
```diff
@@ -637,8 +650,15 @@ namespace rascal {
       // Count all the keys in the sparse gradient structure where the gradient
       // is nonzero (i.e. where the key has an entry in the structure)
       for (auto neigh : center) {
+        // TODO(max,felix) there should be a way to write "if neigh is ghost"
+        if (static_cast<size_t>(neigh.get_atom_tag()) >=
+            this->structure_manager->get_size()) {
+          // Don't compute gradient contributions onto ghost atoms
+          continue;
+        }
+        auto swapped_ref{std::move(swap_pair_ref(neigh).front())};
```
Luthaf: `std::move` is suspicious here. `swap_pair_ref(neigh)` returns a new vector, but .front() returns a constant reference, on which std::move does nothing. Maybe you can have a different function to get the first value, or use `std::move(swap_pair_ref(neigh)[0])` which can do something, or ditch the std::move together if `PairRefKey_t` is trivial enough.

tests/test_calculator.hh
```diff
@@ -653,14 +673,20 @@ namespace rascal {
         result_idx += n_entries_per_key;
       }
       for (auto neigh : center) {
+        if (static_cast<size_t>(neigh.get_atom_tag()) >=
```
Luthaf: can this be made into a function?  `this->structure_manager.is_ghost(neigh.get_atom_tag())` or something. It seems to rely on the underlying representation [real atoms ..., ghost atoms ...] which might change at some point.

tests/test_calculator.hh
```diff
@@ -653,14 +673,20 @@ namespace rascal {
...
+        }
         auto & data_neigh{data_sparse[neigh]};
         // The neighbour gradient (i =/= j) only contributes to certain species
         // channels (keys), in the case of SOAP and SphExpn those keys
         // containing the species of the center (the atom wrt the derivative is
         // being taken)
         // The nonzero gradient keys are already indicated in the sparse
         // gradient structure
-        auto keys_neigh{gradients_sparse[swap_pair_ref(neigh)].get_keys()};
+        auto swapped_ref{std::move(swap_pair_ref(neigh).front())};
```
Luthaf: same with std::move

tests/test_calculator.hh
```diff
@@ -715,72 +748,114 @@ namespace rascal {
         result_idx += n_entries_per_key;
       }
       for (auto neigh : center) {
+        if (static_cast<size_t>(neigh.get_atom_tag()) >=
```
Luthaf: same with is_ghost

tests/test_calculator.hh
```diff
@@ -715,72 +748,114 @@ namespace rascal {
...
+  class RepresentationCalculatorGradientFixture : public GradientTestFixture {
    public:
     using StdVector2Dim_t = std::vector<std::vector<double>>;
     using Calculator_t =
-        RepresentationManagerGradientCalculator<RepManager_t,
-                                                StructureManager_t>;
+        RepresentationCalculatorGradientProvider<RepCalculator_t,
+                                                 StructureManager_t>;
 
     static const size_t n_arguments = 3;
```
Luthaf: not in this PR, but I would rather make static values `ALL_CAPS`

tests/reference_data/methane.json
max-veit: No, the previous molecule crashed because it wasn't wrapped into the unit cell (which was defined as zeros, causing a bad error). This change fixes that.

tests/test_calculator.hh
max-veit: It might no longer be necessary with PairRefKey_t; I haven't tried.

tests/test_calculator.hh
max-veit: I literally left a comment on this a moment ago.

tests/test_calculator.hh
max-veit: yeah yeah, it's in progress (see the TODO earlier)

src/atomic_structure.hh
```diff
@@ -133,6 +133,27 @@ namespace rascal {
...
 
+    /**
+     * Wrap a set of explicitly provided positions into this structure's cell
+     *
+     * Like wrap(), but uses a set of user-specified positions instead of the
+     * atomic positions of this structure.j
+     *
+     * @param input_positions The positions to wrap
+     */
+    Positions_t wrap_explicit_positions(Positions_t input_positions) {
```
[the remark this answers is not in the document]
Luthaf: Is there a test for this function then? I did not see one.

tests/reference_data/methane.json
Luthaf: no, I mean leave out the ctime/user/unique_id fields, keeping the cell & position changes

tests/test_calculator.hh
Luthaf: I've seen the comment, I am just asking to implement this now rather than later :smiley:

tests/test_calculator.hh
```diff
@@ -695,12 +721,19 @@ namespace rascal {
...
+        }
+        auto swapped_ref{std::move(swap_pair_ref(neigh).front())};
         n_entries_neighbours +=
-            (gradients_sparse[swap_pair_ref(neigh)].get_keys().size() *
+            (gradients_sparse[swapped_ref].get_keys().size() *
              n_entries_per_key);
       }
       Eigen::Matrix<double, 3, Eigen::Dynamic, Eigen::RowMajor>
           grad_coeffs_pairs(3, n_entries_center + n_entries_neighbours);
+      grad_coeffs_pairs.setZero();
```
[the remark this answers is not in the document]
Luthaf: ok, fair enough!

src/atomic_structure.hh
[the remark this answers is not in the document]
max-veit: If it comes to that, I'd rather delete it than write a bunch of test cases for this thing. It's exactly the same thing as `wrap()` only using user-supplied positions instead of the atomic ones.

tests/test_calculator.hh
max-veit: Coding convention treats static and dynamic variables the same. If you think this is bad, raise an issue at the next group meeting.

src/atomic_structure.hh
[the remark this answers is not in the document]
Luthaf: if it is not tested and not used, I would rather delete it yes

tests/test_calculator.hh
max-veit: Okay, removing the `std::move` does actually cause a compiler error because it's looking for the (deleted) copy constructor. So what exactly do you mean when you say "std::move does nothing" on this `vector.front()` reference?

max-veit: Hang on, `std::vector` can return either a const or a non-const reference: https://en.cppreference.com/w/cpp/container/vector/front. So how does it decide which one to return? Does it pick the non-const version because `move` only takes a non-const reference?

docs/source/reference/cpp.rst
```diff
@@ -48,10 +48,72 @@ Math utilities (namespace :cpp:class:`rascal::math`)
     :project: rascal
     :members:
 
+Gradient tests (developer)
```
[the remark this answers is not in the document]
max-veit: done

tests/test_calculator.hh
Luthaf: If both are available it should pick one depending on the `const`ness of the vector itself, so I guess it picks the non-const one here.

Looks like a `&&` reference can be created from a `&` reference, so my bad here.

PR #183, opened 2019-10-28
Bugfix/radial scaling

### LINKED ISSUE #182: Incorrect gradients when using radial scaling
The gradient tests for the SphericalExpansion and SphericalInvariants representations are failing when using radial scaling cutoff functions. The problem wasn't caught before because of a brace error in the tests, fixed in 681e64e, meant that radial scaling functions weren't actually being tested - though I'm frankly surprised the test code still worked, given the error. Furthermore, there are no tests for the derivatives of the cutoff functions on their own (which would have also caught the problem much sooner), though it would be straightforward to write one based off of the model of e.g. the hyp1f1 derivative test: https://github.com/cosmo-epfl/librascal/blob/9e72bab303fb4d7b4c91897efd04b1a83b42bde8/tests/test_math_hyp1f1.cc#L131-L142

From a quick look at the code, I suspect the problem is that the radial scaling function is actually a _product_ of the "radial-scaling function" itself and a cosine-type smoothing function near the cutoff. This is not clear from the documentation, which says only this after giving a description of the plain "radial-scaling function": https://github.com/cosmo-epfl/librascal/blob/9e72bab303fb4d7b4c91897efd04b1a83b42bde8/src/representations/cutoff_functions.hh#L129

Since the function is a product, its derivative follows the Leibniz rule - though that's not what seems to be implemented: https://github.com/cosmo-epfl/librascal/blob/9e72bab303fb4d7b4c91897efd04b1a83b42bde8/src/representations/cutoff_functions.hh#L180-L181

So the fix is simple, but we need to write some tests to prevent this sort of thing from happening again. Furthermore, I suggest we look into ways of simplifying how we select hyperparameter sets to test; the current Cartesian-product procedure generates a very large number of tests when we don't necessarily need to check all interactions of all features.

felixmusil:
Fix bad gradient for the radial scaling cutoff function and add tests for the cutoff functions.

Fix #182

Tasks before review:

          elsewhere in the code.
          auto-formatting is sensible
          own branch

### DIFF
--- a/src/representations/cutoff_functions.hh
+++ b/src/representations/cutoff_functions.hh
@@ -151,7 +151,7 @@ namespace rascal {
         this->scale = hypers.at("scale").at("value").get<double>();
       }
 
-      double f_c(double distance) {
+      double value(double distance) {
         double factor{0.};
         if (this->rate > math::dbl_ftol) {
           factor = this->rate / (this->rate + math::pow(distance / this->scale,
@@ -161,11 +161,10 @@ namespace rascal {
         } else {
           factor = math::pow(distance / this->scale, -this->exponent);
         }
-        return factor * math::switching_function_cosine(distance, this->cutoff,
-                                                        this->smooth_width);
+        return factor;
       }
 
-      double df_c(double distance) {
+      double grad(double distance) {
         double factor{0.};
         if (this->rate < math::dbl_ftol) {
           factor = -this->exponent / distance *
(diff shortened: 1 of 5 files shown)


max-veit:
Did you merge in the periodic SOAP gradients stuff to here? This branch was meant to be an independent fix...

max-veit:
Rebased to sort out changes for the periodic gradients fix; @felixmusil please check and make sure I didn't break anything

PR #187, opened 2019-10-30
fix 184

### LINKED ISSUE #184: Fix code without effects in src/math/bessel.hh
These lines do nothing, since unaryExpr does not modify the data inplace.

https://github.com/cosmo-epfl/librascal/blob/4feef3147c9784771602a6e893603def00df2525/src/math/bessel.hh#L203-L210

felixmusil:
Fix #184


Luthaf:
What is the rational for setting small values to 0? Can it be documented?

Tests seems to be passing, but they also where passing before. Could you add a small test to ensure this behaviour is guaranteed in case of future changes (cf PR template)?

src/math/bessel.hh
```diff
@@ -200,7 +200,9 @@ namespace rascal {
           this->upward_recursion(distance, fac_a, n_up);
         }
 
-        // set small values to 0.
+        // set small values to 0. On the python side it avoids some unexpected
+        // interpretation of values that are strictly speaking outside of the
+        // range of double precision
```
Luthaf: I don't understand this comment. 1e-100 is still withing the range of double precision, and pretty far from the limit (1e-308). What do you mean by `unexpected interpretation`?

Luthaf:
Except for my question above, this is good with me!

src/math/bessel.hh
felixmusil: 1e-100 is because we anyway don't care for smaller stuff and you can get values ~1e-312 which on the python side are not well handled by json serialization for instance.

Luthaf:
good for me, this branch should be rebased since there is a small conflict with master. You can also squash the commits while rebasing if you want.

bindings/rascal/neighbourlist/structure_manager.py
```diff
@@ -44,15 +44,17 @@ def __init__(self, frames, nl_options, start=None, length=None):
             managers = StructureCollectionFactory(nl_options)
             try:
                 managers.add_structures(structures)
-            except:
-                print("""Neighbourlist of structures failed. trying
-                one at a time.""")
+            except Exception as e:
```
Luthaf: git error? this should not be in this PR =)

PR #196, opened 2019-10-30
Bugfix/nl failure message

felixmusil:
I though this was already taken care of by a previous PR but it seems not so here it is. The potential failures in the c++ neighbor list are now caught and displayed in python.


mastricker:
It is part of the neighbourlist repair branch `feat/delete_ghost_neighbours`. So I am not gonna review it?

Luthaf:
> It is part of the neighbourlist repair branch feat/delete_ghost_neighbours. So I am not gonna review it?

Does it need to wait until `feat/delete_ghost_neighbours` is ready? It feels independant

felixmusil:
I thought it was already in master since we discussed this a while ago already and it is a quick fix. I would put it on master now so that I can avoid having more users being bugged by this.

bindings/rascal/neighbourlist/structure_manager.py
```diff
@@ -44,15 +44,18 @@ def __init__(self, frames, nl_options, start=None, length=None):
...
                 for structure, manager in zip(structures, managers):
                     try:
                         manager.update(structure)
-                    except:
-                        print("Structure Rep computation {} failed".format(ii))
+                    except Exception as e:
+                        print(
+                        "Neighbourlist of structure"+
+                        " {} failed because: ".format(ii) + str(e))
+                    ii += 1
```
mastricker: I do not see why we need the 'one at a time'. If one structures is failing half way through, it is not checked if the other would have gotten through?
I would get rid of the one-at-a-time and just fail. In my opinion it is very likely that other structures will fail, too. So why not clean up the whole data and retry...

felixmusil: I keep it because I think it is useful when wrapping the atoms is not enough to avoid failures. You can single out the structure and try to understand better what happens. In principle there should not be failures if the atoms are wrapped but who knows.

mastricker: I don't see why this version helps you more when wrapping atoms is not enough - that should be a different error message anyways. And if you really have _bad_ structures, you could try them one by one outside by not supplying a list but checking them in a python loop like

Pseudo-example
```
for structure in structures:
    rep = compute(structure)
```

If the output of the error message would list _all_ structures, which have failed, I can see that as a valuable information, but if the looping is done only until it fails and possible structures in line behind that also fail, I don't see the point. It is basically hoping that there is only one offending structure. If you have multiple offending ones and you fix them one after another, it feels to me that it is like banging your head on the until the wall gives in. I myself at least would go the example route above if I encountered that issue twice.

So my suggestion is: either getting rid of the looping or make it useful by going through all the input structures and list them as a list of _failed_ structures with indices afterwards. Something like "structures [12,34,312,2422] were not computed because of errors 'blabla bla'. Then you can also see the different errors.

max-veit:
Not sure if this is related, but I get a crash with no sensible error message if the cell vectors are set to all 0 (which is somewhat common if you're using molecules with no PBCs). This was the reason for updating `methane.json`, besides it not being wrapped properly (because the cell vectors were all zero).

mastricker:
I would say this is only related in that it concerns the neighbour list because it relies on cell vectors being nonzero and larger than the extend of the atomic positions. They are used in building the grid for the linked lists.

Zero-cell vectors are only valid for non-periodic cells, right?
So that could be caught independently and the _internal_ cell could just be the minimum/maximum coordinates. PBCs (on/off-true/false per direction) are handled correctly, as far as I can say and the tests go.
I could add that special case to the Neighbourlist, but I would say it is a different issue and it does not result in a comprehensible error message. Except that the atoms are not inside the box, which they can not be...right?

max-veit:
Yes, I think zero cells are fine as a special case. It usually means that there is no periodicity for the system in question, and the software/user either doesn't know about PBC flags or doesn't support them.

mastricker:
That can be implemented, I will create an issue and self-assign.

Just to specify:
If such a molecule does not define the cell , i.e. cell vectors are zero:
1. Are they usually all zero or can it be that only one being zero translates to no periodicity in one direction?
2. Are the coordinates of the atomic positions > 0 in all directions?

I.e.: what is the _worst_ case we need/want to catch, I would say:
- Atomic positions > 0 in all dimensions
- All cell vectors 0
-> make a shrink-wrap cell with some margin around the atomic positions, assume no periodicity and build the neighbourlist.

Other cases where only one or two cell vectors are zero should probably raise an error, like the failed wrapping?

And: do you have an example molecule which I can use for test-driven implementation?

felixmusil:
> Yes, I think zero cells are fine as a special case. It usually means that there is no periodicity for the system in question, and the software/user either doesn't know about PBC flags or doesn't support them.

I added a routine to handle non periodic structures that have cell == 0 . For the moment we expect to have a cell filled with zeros when it is not provided: default behavior of ase the only supported external structure input atm.
On the c++ side we read ase formated json structures so I expect the field to exist and it might be filled with zeros so we need to catch that.

max-veit:
> And: do you have an example molecule which I can use for test-driven implementation?

Too late now, but I was going to suggest the old `methane.json` file that made me stumble into this error in the first place before I fixed the cell in 5374ec3. That would have been the simplest test case; I guess `small_molecule_no_cell.json` works too but it's a bit too complicated to see what's going on if we do break this feature in the future. (It may be a "small" molecule if you ask a biochemist, but not if you ask e.g. a quantum chemist...)

PR #194, opened 2019-10-30
Move some code out of headers

Luthaf:
This should improve compile time for rascal. I find locally a 10% decrease in compile time for a clean build with this PR, from 10 min to 9 min in release and from 12min to 11min in debug.

This mainly touches files in src/math and bindings. Other headers are template classes, so splitting the code between header and implementation will be harder. I can see two possibility: using extern template to declare explicitly instantiated templates, or using `.tpp` files with template code separated from the declarations.

---

Tasks before review:

- [x] benchmark overall performance to ensure the compiler do not miss optimization opportunities

TBD in a future PR:
- [ ] evaluate the use of `extern template` in the remaining of the code
- [ ] split test (they should not need header), and move tests utils to a separate place.

PR #208, opened 2019-11-07
Use a common folder for the main rascal headers

### LINKED ISSUE #189: #include namespace for rascal headers
Currently, rascal headers are included directly, without any filesystem namespace. i.e. use `#include "math/math_utils.hh"` vs `#include "rascal/math/math_utils.hh"`.

This is not ideal for **users** of the C++ library who need to know which headers are from rascal, and which are not.

I think this needs to be updated. A nice way to do it is to separate headers and sources to look like:
```bash
librascal/
    include/
        rascal/
            math/
            representation/
    src/
        math/
        representation/
```

Luthaf:
Fix #189 

----

Tasks before review:

- [X] Add tests, examples, and complete documentation of any added functionality
    - [X] Make sure the autogenerated documentation appears in Sphinx and is
- [X] Run `make lint` on the project, ensure it passes
    - [X] Run `make pretty-cpp` and `make pretty-python`, check that the
          auto-formatting is sensible


examples/test_sph_harm_poles.cc
```diff
@@ -1,5 +1,5 @@
 /**
- * file   test_sph_harm_poles.cc
+ * @file   examples/test_sph_harm_poles.cc
```
max-veit: I thought we'd agreed that this file can be deleted. When did it get revived?

Luthaf: It is deleted in #194

tests/test_adaptor_half_full_neighbour_list.cc
```diff
@@ -25,10 +25,11 @@
...
-#include "structure_managers/adaptor_full_neighbour_list.hh"
-#include "structure_managers/adaptor_half_neighbour_list.hh"
 #include "test_structure.hh"
 
+#include "rascal/structure_managers/adaptor_full_neighbour_list.hh"
+#include "rascal/structure_managers/adaptor_half_neighbour_list.hh"
+
 #include <boost/test/unit_test.hpp>
 
 constexpr double TOLERANCE = 1e-14;
```
mastricker: This should probably go to the math tolerance, right? 
I'll make an issue.

tests/test_adaptor_increase_maxorder.cc
```diff
@@ -26,9 +26,10 @@
...
  */
 
-#include "structure_managers/adaptor_half_neighbour_list.hh"
 #include "test_structure.hh"
 
+#include "rascal/structure_managers/adaptor_half_neighbour_list.hh"
+
 #include <boost/test/unit_test.hpp>
 
 constexpr double TOLERANCE = 1e-12;
```
mastricker: This also.

PR #216, opened 2019-11-12
Disable debug symbols on circle CI

Luthaf:
As discussed in last meeting, this might help with OOM errors on circle CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,7 +46,7 @@ jobs:
     environment:
       CC: gcc-9
       CXX: g++-9
-      CMAKE_EXTRA: -DCMAKE_BUILD_TYPE=Debug
+      CMAKE_EXTRA: -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_FLAGS_DEBUG=
     <<: *default_job
   clang-4:
     environment:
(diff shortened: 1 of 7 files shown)


.circleci/config.yml
```diff
@@ -46,7 +45,7 @@ jobs:
     environment:
       CC: gcc-9
       CXX: g++-9
-      CMAKE_EXTRA: -DCMAKE_BUILD_TYPE=Debug
+      CMAKE_EXTRA: -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_FLAGS_DEBUG=
```
agoscinski: why is this needed `DCMAKE_CXX_FLAGS_DEBUG=`?

Luthaf: This set `CMAKE_CXX_FLAGS_DEBUG` to the empty string, instead of `-g`

`-DCMAKE_CXX_FLAGS_DEBUG=""` sets it to the string containing two double quotes ...

performance/benchmarks/benchmark_spherical_representations.py
```diff
@@ -1,104 +1,109 @@
...
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this software; see the file LICENSE. If not, write to the
+# Free Software Foundation, Inc., 59 Temple Place - Suite 330,
+# Boston, MA 02111-1307, USA.
 
 # This file should be executed from the building directory
-# quick solution for loading 
-import sys, os
-sys.path.insert(0, os.getcwd())
```
agoscinski: Hey you mentioned this today I think, but I did not get why this is not needed anymore?

Luthaf: We set PYTHONPATH instead: https://github.com/cosmo-epfl/librascal/blob/e151317f6461fb440739d4a664818cf654c249ff/performance/benchmarks/CMakeLists.txt#L28-L30

Issue #217, opened 2019-11-13
homogenize .clang-format file or find out different version


mastricker:
base our style on an existing style or find out if formatting depends on the clang-format version

PR #219, opened 2019-11-14
Rebase onto master

mastricker:
As per discussion yesterday, I checked `clang-format` versions 6, 7, 8, 9
With this updated configuration including a base-style it does not change anything for me, regardless of the `clang-format` version.

Please do check if that is also the case for whatever local configuration you have.

This PR should fix issue #217, i.e. commit noise due to clang-format
- Added base style
- Reordered alphabetically in .clang-format
- Reordered folders in `CMakeLists.txt` alphabetically

Tasks before review:

Remaining tasks, out of scope of this pull request: (optional)
- I ask someone else with a different configuration to check for possible _noise_ and get back to me with the clang-format version

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,14 +1,9 @@
+# Default style
+BasedOnStyle: LLVM
+# Customization
 AccessModifierOffset: -1
-SpacesBeforeTrailingComments: 2
-NamespaceIndentation: All
-SortIncludes: false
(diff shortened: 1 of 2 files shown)


Luthaf:
Concerning the diff noise, we still have to ensure `make pretty` was ran with CI, but this will ensure no false positive can go through

PR #225, opened 2019-11-18
Feat/dvr grad

felixmusil:
Implement the radial integral derivatives for DVR in Spherical expansion.

- implement derivative of modified spherical bessel function
- improve a bit the overflow checks in hyp1f1.hh and bessel.hh
- increase the range of tests for hyp1f1 

Tasks before review:

          auto-formatting is sensible


src/rascal/math/bessel.hh
```diff
@@ -199,6 +252,7 @@ namespace rascal {
         if (n_up > 0) {
           this->upward_recursion(distance, fac_a, n_up);
         }
+        assert(this->bessel_values.isFinite().all());
```
Luthaf: CI is failing on this assert, so it looks like it was useful!

src/rascal/math/bessel.hh
```diff
@@ -222,19 +282,29 @@ namespace rascal {
        *         Note that a reference is returned to avoid unnecessary
        *         copies.
        */
-      auto get_values() { return ref(bessel_values); }
+      auto get_values() {
+        if (this->compute_gradients) {
+          // when the gradient are computed bessel_values has one additional
+          // column that should not be returned
```
Luthaf: Could the extra column be stored separately?

src/rascal/math/bessel.hh
felixmusil: yes I used them a lot too when making the thing work. The problem on the CI is still mysterious though.

src/rascal/math/bessel.hh
felixmusil: it could but the implementation of gradient_recursion, upward_recursion and downward_recursion would not be as straight forward.

Luthaf: ok, fair enough.

src/rascal/representations/calculator_spherical_expansion.hh
```diff
@@ -775,7 +782,7 @@ namespace rascal {
         auto smearing{downcast_atomic_smearing<AST>(this->atomic_smearing)};
 
         // a = 1 / (2*\sigma^2)
-        double fac_a{0.5 * pow(smearing->get_gaussian_sigma(center), -2)};
+        double fac_a{0.5 / pow(smearing->get_gaussian_sigma(center), 2_size_t)};
```
Luthaf: why was this changed? To ensure `math::pow` over `std::pow`?

tests/test_calculator.hh
```diff
@@ -323,7 +323,8 @@ namespace rascal {
     ~SimpleStructureManagerNLCCStrictFixture() = default;
 
     const std::string filename{
-        "reference_data/CaCrP2O7_mvc-11955_symmetrized.json"};
+        "reference_data/CaCrP2O7_mvc-11955_symmetrized.json",
```
Luthaf: wait, what is this doing? We don't have a `std::vector<std::string>` at the end, only a single `std::string` initialized with two `const char*`. And AFAIK, there is [no constructor for `std::string` taking two `const char*`](https://en.cppreference.com/w/cpp/string/basic_string/basic_string).

Is `"reference_data/small_molecule.json"` being converted to `size_type` here?

tests/test_calculator.hh
```diff
@@ -570,6 +573,84 @@ namespace rascal {
...
+  struct RadialIntegralHandlerFixture : MultipleStructureFixture<BaseFixture> {
+    using Parent = MultipleStructureFixture<BaseFixture>;
+    using Manager_t = typename Parent::Manager_t;
+    using RadialIntegral_t =
+        internal::RadialContributionHandler<RadialType, SmearingType, OptType>;
+
+    RadialIntegralHandlerFixture() : Parent{} {
+      // filter out the hypers that don't correspond to the current Type
+      std::vector<json> hypers_temp;
+      // here it directly sets GTO DVR and only
```
Luthaf: I don't understand this comment. Is the end of the sentence missing?

tests/test_math_hyp1f1.cc
```diff
@@ -83,7 +83,7 @@ namespace rascal {
         std::cout << func.get_z_switch() << std::endl;
       }
 
-      BOOST_CHECK_LE(der_consistency_rel_error, 3e5 * math::DBL_FTOL);
+      BOOST_CHECK_LE(der_consistency_rel_error, 2e6 * math::DBL_FTOL);
```
Luthaf: Why did the precision decreased here?

src/rascal/representations/calculator_spherical_expansion.hh
felixmusil: yes it is

tests/test_calculator.hh
felixmusil: I am not sure why it did not crash...

tests/test_math_hyp1f1.cc
felixmusil: I increase the range of tested values and that is the minimal precision that we get on the whole range.

src/rascal/representations/calculator_spherical_expansion.hh
Luthaf: Should we rename `math::pow` to `math::fast_pow` to be sure what we are using at a given point in code?

felixmusil: seems too verbose for no reason to me. Function overloading does a good job here.

tests/test_calculator.hh
felixmusil: neither do I. I guess @agoscinski should look at it.

src/rascal/representations/calculator_spherical_expansion.hh
Luthaf: One reason would be knowing immediately which function is called when reading the code

tests/test_calculator.hh
felixmusil: I reworded the block of comment

src/rascal/representations/calculator_spherical_expansion.hh
felixmusil: this change goes beyond this PR so let's discuss this elsewhere. I personally prefer function overloading.

Luthaf: #227

src/rascal/math/bessel.hh
```diff
@@ -100,7 +100,7 @@ namespace rascal {
        *         Note that a reference is returned to avoid unnecessary
        *         copies.
        */
-      Eigen::Ref<Eigen::ArrayXXd> get_values() {
+      auto get_values() {
```
Luthaf: Please don't do this. I want to see what is going on just by looking at the function signature, not the implementation.

felixmusil: This is not the type of the returned value I and actually don't really know what it is since `ref(this->bessel_values.leftCols(this->l_max + 1));` is returned. This is a perfect case to use auto I think

Luthaf: I would rather take the time to figure the type once, so that future readers of the code don't have to do it in the future. You can get the compiler to print you the type with

```
auto a = ref(this->bessel_values.leftCols(this->l_max + 1));
decltype(a)::do_not_exists = 44;
``` 

This will generate an error message with the compiler-deduced type.

felixmusil: That is the problem and your code snippet is not applicable here. 
In principle this function returns 2 different types since in the else branch it returns `Eigen::Ref<Eigen::ArrayXXd>`. So Eigen here is doing its magic finding a common ground which is equivalent to `Eigen::Ref<Eigen::ArrayXXd>`making sure it compiles properly.

Luthaf: This is confusing me. The magic Eigen type must still be nameable, right? You say it is not `Eigen::Ref<Eigen::ArrayXXd>`, is it closer to `Eigen::Ref<Eigen::Map<Eigen::ArrayXXd>>`?

Luthaf: Related, if the return type of the new branch is convertible to `Eigen::Ref<Eigen::ArrayXXd>`, why not keep the interface as returning Eigen::Ref<Eigen::ArrayXXd>?

Luthaf: For reference, the actual type was `Eigen::Ref<const Eigen::ArrayXXd>`, fiexed in a4c433cba32d7c8e804fe0729bb9dcc767e79177

Issue #107, opened 2019-08-02
SOAP gradients computes superfluous zero entries


max-veit:
Exactly what it says on the tin: The current SOAP gradients implementation computes entries that are known to be zero from the species keys (e.g. if taking the gradient of a SOAP entry with species (C, Si) wrt a neighbour of type Ge, the gradient is _known_ to be zero, but it's iterated over and explicitly zeroed anyway). Our first attempt at skipping the zero-gradient entries (a98605efd9) was unsuccessful, so we're going to look into this more closely and fix it soon.

Luthaf:
This code seems to be related (see https://github.com/cosmo-epfl/librascal/pull/150#discussion_r333900165), saving it here for future usage:

```cpp
// Neighbour gradients need a separate pair list because if the species
// of j is not the same as either of the species for that SOAP entry,
// the gradient is zero.
for (auto neigh : center) {
  std::vector<internal::SortedKey<Key_t>> grad_pair_list{};
  for (const auto & el1 : coefficients) {
    auto && neigh_1_type{el1.first[0]};
    for (const auto & el2 : coefficients) {
      auto && neigh_2_type{el2.first[0]};
      auto neigh_type = neigh.get_atom_type();
      if (neigh_1_type <= neigh_2_type) {
        pair_type[0] = neigh_1_type;
        pair_type[1] = neigh_2_type;
        if ((neigh_type == pair_type[0]) or
            (neigh_type == pair_type[1])) {
          grad_pair_list.emplace_back(is_sorted, pair_type);
        }
      }
    }
  }
}
```

Luthaf:
Fixed in #228 as far as I know

PR #228, opened 2019-11-24
Feat/improve soap grad

felixmusil:
Make sure the contribution from neighbors to the soap gradients are as sparse as possible with respect ot the species key (see #107) and add a test for this feature.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,6 +73,7 @@ option(BUILD_PROFILES "Build the files for profiling" OFF)
 option(BUILD_TESTS "Build the unit tests" OFF)
 option(BUILD_DOC "Build documentation" OFF)
 option(USER "Switch installation folder" OFF)
+option(BUILD_SANDBOX "If on, builds the sandbox" OFF)
 
 set(INSTALL_PATH "" CACHE STRING "Path to install the libraries")
 
@@ -208,6 +209,10 @@ endif()
 if(BUILD_BINDINGS)
     add_subdirectory(bindings)
 endif()
+
+if (${BUILD_SANDBOX})
+    add_subdirectory(sandbox)
+endif (${BUILD_SANDBOX})
 
 if(BUILD_DOC)
     add_subdirectory(docs)
--- a/performance/profiles/profile_spherical_invariants.cc
+++ b/performance/profiles/profile_spherical_invariants.cc
@@ -48,13 +48,9 @@
 // using namespace std;
 using namespace rascal;  // NOLINT
 
-const int N_ITERATIONS = 1000;
-
-using Manager_t = AdaptorStrict<AdaptorNeighbourList<StructureManagerCenters>>;
+const int N_ITERATIONS = 10;
 
 using Representation_t = CalculatorSphericalInvariants;
-
-using Property_t = typename Representation_t::template Property_t<Manager_t>;
 
 int main(int argc, char * argv[]) {
   if (argc < 2) {
@@ -133,7 +129,4 @@ int main(int argc, char * argv[]) {
   std::cout << "Compute representation with gradients"
             << " elapsed: " << elapsed_grad.count() / N_ITERATIONS << " seconds"
             << std::endl;
-  // auto property_name{representation.get_name()};
-  // auto&& property{manager->template
-  // get_validated_property_ref<Property_t>(property_name)};
 }
--- a/src/rascal/representations/calculator_spherical_invariants.hh
+++ b/src/rascal/representations/calculator_spherical_invariants.hh
@@ -1227,7 +1227,7 @@ namespace rascal {
           }
           soap_vector_gradients[neigh].resize(
               grad_pair_list, n_spatial_dimensions * n_row, n_col, 0.);
-        } // auto neigh : center
+        }  // auto neigh : center
       }  // if compute gradients
     }    // for center : manager
   }
--- a/src/rascal/representations/calculator_spherical_invariants.hh
+++ b/src/rascal/representations/calculator_spherical_invariants.hh
@@ -538,8 +538,6 @@ namespace rascal {
                   expansions_coefficients_gradient[neigh]};
               auto & soap_neigh_gradient{soap_vector_gradients[neigh]};
 
-
-
               auto neigh_type = neigh.get_atom_type();
               if ((neigh_type != spair_type[0]) and
                   (neigh_type != spair_type[1])) {
@@ -1203,8 +1201,6 @@ namespace rascal {
         soap_vector_gradients[ii_pair].resize(
             pair_list, n_spatial_dimensions * n_row, n_col, 0.);
 
-
-
         // Neighbour gradients need a separate pair list because if the species
         // of j is not the same as either of the species for that SOAP entry,
         // the gradient is zero.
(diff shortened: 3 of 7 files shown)


CMakeLists.txt
```diff
@@ -73,6 +73,7 @@ option(BUILD_PROFILES "Build the files for profiling" OFF)
 option(BUILD_TESTS "Build the unit tests" OFF)
 option(BUILD_DOC "Build documentation" OFF)
 option(USER "Switch installation folder" OFF)
+option(BUILD_SANDBOX "If on, builds the sandbox" OFF)
```
Luthaf: That's nicer than the previous solution! Why do we need the sandbox to be part of the repository though? I.e. what does this bring that could not be done with this?

```bash
vim example.cpp
g++ example.cpp -Ilibrascal/src -Ilibrascal/build/external/Eigen3 -Llibrascal/build -lrascal
```

performance/profiles/profile_spherical_expansion.cc
```diff
@@ -48,7 +48,7 @@
 // using namespace std;
 using namespace rascal;  // NOLINT
 
-const int N_ITERATIONS = 1000;
+const int N_ITERATIONS = 10;
```
Luthaf: Does this need to be changed or was it changed for development and should be reset?

performance/profiles/profile_spherical_invariants.cc
```diff
@@ -48,14 +48,10 @@
 // using namespace std;
 using namespace rascal;  // NOLINT
 
-const int N_ITERATIONS = 1000;
-
-using Manager_t = AdaptorStrict<AdaptorNeighbourList<StructureManagerCenters>>;
+const int N_ITERATIONS = 10;
```
Luthaf: Same as the other profile file

src/rascal/representations/calculator_spherical_invariants.hh
```diff
@@ -550,6 +547,10 @@ namespace rascal {
                 continue;
               }
 
+              auto && soap_neigh_gradient_by_species_pair{
```
Luthaf: Why the `auto &&`here? Could `const auto&` do the job?

src/rascal/representations/calculator_spherical_invariants.hh
```diff
@@ -550,6 +547,10 @@ namespace rascal {
...
+                  soap_neigh_gradient[spair_type]};
+              // soap_neigh_gradient_by_species_pair.setZero();
```
Luthaf: Why is this commented out?

src/rascal/representations/calculator_spherical_invariants.hh
```diff
@@ -550,6 +547,10 @@ namespace rascal {
...
+
               // TODO(max) is there a symmetry here we can exploit?
```
Luthaf: ping @max-veit, is this still relevant?

tests/test_calculator.cc
```diff
@@ -24,6 +24,7 @@
  * Free Software Foundation, Inc., 59 Temple Place - Suite 330,
  * Boston, MA 02111-1307, USA.
  */
+#define BOOST_TEST_MODULE boost_test_sequence
```
Luthaf: Could you remind me why this is needed? We don't use it in other files, right?

tests/test_calculator.cc
```diff
@@ -32,6 +33,8 @@
 #include <boost/mpl/list.hpp>
 #include <boost/test/unit_test.hpp>
 
+#include <set>
```
Luthaf: I don't see any use of `std::set` in this file

tests/test_calculator.cc
```diff
@@ -311,6 +314,102 @@ namespace rascal {
...
+    auto & hypers = Fix::representation_hypers;
+    const bool verbose{false};
+    // assumes the structure is CaCrP2O7_mvc-11955_symmetrized.json
+    // the keys of the center contribution (1st center)
+    std::vector<std::vector<int>> all_keys{{8, 8},   {8, 15},  {8, 20},
+                                           {15, 15}, {15, 20}, {20, 20}};
+    // the list of keys for each neighbor (1st center)
+    std::vector<std::vector<std::vector<int>>> neigh_keys{
+        {{8, 15}, {15, 15}, {15, 20}}, {{8, 8}, {8, 15}, {8, 20}},
+        {{8, 8}, {8, 15}, {8, 20}},    {{8, 8}, {8, 15}, {8, 20}},
```
Luthaf: Formatting is weird here. Would be clearer like this:

```
        {{8, 15}, {15, 15}, {15, 20}}, 
        {{8, 8}, {8, 15}, {8, 20}},
        {{8, 8},  {8, 15},  {8, 20}},  
        {{8, 8}, {8, 15}, {8, 20}},
```

Or without the spaces inside. Or just let clang-format do its thing =)

src/rascal/representations/calculator_spherical_invariants.hh
max-veit: Yup. I think the gradients code can be simplified quite a bit, but I'm still not sure how.

max-veit:
Looks good. Besides the comments from @Luthaf, I noticed for the tests that the structure you added (the CaCr six-element crystal) has some overlapping purposes with the Si-Ge-C structure I originally used to test SOAP gradient sparsity. The latter is a much more minimal example that allows manual inspection of the SOAP and gradient vectors (and alerted me to #107 in the first place), while the one you added seems to be better for benchmarking the asymptotics of the gradient performance in the limit of a large number of species. What are your thoughts?

CMakeLists.txt
felixmusil: I like to avoid starting from scratch every time I develop something. Also I would prefer to stay within the cmake environment so that debuging this file is the same as other files of the library. This folder is only part of the early development phase and will be removed at some point.

performance/profiles/profile_spherical_expansion.cc
felixmusil: it was too high for the gradient computation with a not so large structure.

src/rascal/representations/calculator_spherical_invariants.hh
felixmusil: I think nothing forbids us to use the auto && in the coding convention so I use it instead of something else. Why would you not use this language feature ?

tests/test_calculator.cc
felixmusil: I use boost::test_tools::per_element() to simplify test over collection collections so [boost needs this define to work](https://www.boost.org/doc/libs/1_65_0/libs/test/doc/html/boost_test/testing_tools/extended_comparison/collections.html)

tests/test_calculator.cc
felixmusil: this is the auto format

felixmusil:
I used this structure because it is the one I have been using for the last year. Also I feel that the gain for the gradient sparsity start to be visible only with a few species like here: I am not sure the structure you suggest will allow neigh_keys to have some variations like in the example I am using to test. Anyhow I would stick with this structure at this point.

src/rascal/representations/calculator_spherical_invariants.hh
Luthaf: As always, because I like to see the types, and `auto &&` breaks that, since it can bind by value or by reference.

Luthaf:
Good for me, I'll let @max-veit max finish the discussion on which file to use for tests.

Luthaf:
Don't forget to delete your branches!

PR #237, opened 2019-12-03
Run make pretty-cpp in CI

Luthaf:
As discussed recently-ish =)

I found that clang-format <= 7 has a different output than clang-format >=8, so I added a version check in cmake to ensure everyone has the same output.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ version: 2
 # indicated by the CC/CXX environment variables.
 default_job: &default_job
   docker:
-    - image: cosmoepfl/rascal-ci:2
+    - image: cosmoepfl/rascal-ci:3
   steps:
     - checkout
     - run:
@@ -73,10 +73,11 @@ jobs:
       CXX: g++-9
       CMAKE_EXTRA: -DRASCAL_TESTS_USE_VALGRIND=ON -DTYPE_ARCHITECTURE=core2
     <<: *default_job
+
   # Special job only building the documentation
   docs:
     docker:
-    - image: cosmoepfl/rascal-ci:2
+    - image: cosmoepfl/rascal-ci:3
     steps:
       - checkout
       - run:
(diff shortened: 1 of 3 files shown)


mastricker:
Really, even with the config file....damn it. I could not find any difference on my machine.

.circleci/config.yml
```diff
@@ -5,7 +5,7 @@ version: 2
 # indicated by the CC/CXX environment variables.
 default_job: &default_job
   docker:
-    - image: cosmoepfl/rascal-ci:2
+    - image: cosmoepfl/rascal-ci:3
```
mastricker: What is the difference between 2/3 exactly?

mastricker: Apparently 2 existed while 3 does not according to the CI.

Luthaf: The difference is https://github.com/cosmo-epfl/librascal/pull/237/files#diff-bebe822d715123ec31b03d977212fe2a> And I forgot to update the dockerhub image, I'll restart the build

Luthaf: Everything is here: https://hub.docker.com/repository/docker/cosmoepfl/rascal-ci

Luthaf:
> Still puzzled that it does make a difference even with the config file.

Did you update the value of `CLANG_FORMAT` using ccmake? This is how I switched from one to another locally.

mastricker:
To make sure using the right executable, I did it manually in the command line without any make/cmake.

On Dec 3, 2019 17:44, Guillaume Fraux <notifications@github.com> wrote:

Still puzzled that it does make a difference even with the config file.

Did you update the value of CLANG_FORMAT using ccmake? This is how I switched from one to another locally.

-

.circleci/config.yml
mastricker: Hmm, I need a docker account to access it. I'll check that.

Luthaf:
The error was that the cmake file tried to test clang-format version even if clang-format was not found.

PR #239, opened 2019-12-04
Bugfix/tidy gen scripts

### LINKED ISSUE #238: Rename ShiftedCosine to CosineShifted (esp. ref data generation scripts)
Is done in tests and code, but has to be done in the scripts for reference data.

I am lost with this - the file with the DFT smiles is missing....

Please help me sorting dated/outdated stuff. @max-veit  or @felixmusil ?

felixmusil:
I deleted dft-smiles_500.json in a previous merge but forgot to remove dft-smiles_500.ubjson. This is done now and I updated the generation scripts and explicit paths. I use instead the first 20 structures of small_molecules.xyz.
`make pretty-python` was messing the scripts so I tried to ignore a specific rule and realized that the python config style for autopep8 was not used. Now this is fixed and all python files have been properly auto formatted.

Fix #238


felixmusil:
yes I think so.

PR #241, opened 2019-12-06
Feat/half nl soap

felixmusil:
Adapt the spherical expansion and invariants so that it can work with half neighbor lists.


src/rascal/structure_managers/structure_manager.hh
```diff
@@ -294,7 +294,7 @@ namespace rascal {
     size_t size() const { return this->implementation().get_size(); }
 
     template <size_t Order, size_t Layer>
-    size_t is_center_atom(const ClusterRefKey<Order, Layer> & cluster) const {
+    bool is_center_atom(const ClusterRefKey<Order, Layer> & cluster) const {
       return (cluster.get_atom_tag() < static_cast<int>(this->size()));
```
mastricker: I can't really comment on the implementation, that should be someone else more familiar with SOAP.

But just quickly here: This only works if you use the `StructureManagerCenters`, where you assign you own tags. If lammps supplies you with its tags , there is not guarantee. Rather use the cluster index of lowest layer and compare it to the size, that should be bullet prove. E.g. you could have atoms tags which are [223, 230002, 4224422]. And size is 3.

felixmusil: you are right. I made the modification but now it is only valid for clusters of order 1 and 2.

src/rascal/math/utils.hh
```diff
@@ -33,6 +33,7 @@
 
 #include <cmath>
 #include <cstdint>
+#include <iostream>
```
Luthaf: I would rather not have <iostream> in a public header. Can you replace the writing to `std::cout` by an exception where needed?

We could also add a small logging/warning framework for this.

src/rascal/representations/calculator_spherical_expansion.hh
```diff
@@ -1571,43 +1573,75 @@ namespace rascal {
       expansions_coefficients_gradient.resize();
     }
 
-    /* @TODO(felix,max) use the parity of the spherical harmonics to use half
-     * neighbourlist, i.e. C^{ij}_{nlm} = (-1)^l C^{ji}_{nlm}.
-     */
+    // TODO(felix) think about an option to have "global" species,
```
Luthaf: Could this be an issue instead? This way it can have more context, and other can help.

src/rascal/representations/calculator_spherical_expansion.hh
```diff
@@ -1617,42 +1651,65 @@ namespace rascal {
             radial_integral->template compute_neighbour_contribution(dist,
                                                                      neigh);
         double f_c{cutoff_function->f_c(dist)};
-        auto && coefficients_center_by_type{coefficients_center[neigh_type]};
+        auto coefficients_center_by_type{coefficients_center[neigh_type]};
+        // auto coefficients_neigh_by_type{coefficients_neigh[center_type]};
```
Luthaf: commented out code

src/rascal/structure_managers/adaptor_neighbour_list.hh
```diff
@@ -633,10 +633,10 @@ namespace rascal {
     }
 
     //! Returns atom type given an atom tag, also works for ghost atoms
-    int & get_atom_type(int atom_tag) {
-      // return this->atom_types[this->get_atom_index(atom_tag)];
-      return this->atom_types[atom_tag];
-    }
+    int & get_atom_type(int atom_tag) { return this->atom_types[atom_tag]; }
```
Luthaf: Why is this function returning a non-const reference? Are we writing to it anywhere?

src/rascal/math/utils.hh
felixmusil: I want a warning not a termination of the program. The simplest way to do that is cout. What would you propose ?

src/rascal/representations/calculator_spherical_expansion.hh
felixmusil: I don't think it is need. I will eventually do this task but I still need to think how to do it precisely.

src/rascal/structure_managers/adaptor_neighbour_list.hh
felixmusil: I am not quite sure. That's just how it was.

src/rascal/math/utils.hh
Luthaf: #245 

Let's keep the code as-is for now, and replace it later

src/rascal/structure_managers/adaptor_neighbour_list.hh
Luthaf: Can you remove it and only provide the other one then?

src/rascal/representations/calculator_spherical_expansion.hh
Luthaf: This is what we decided in the last meeting this was brought up.

Luthaf:
This looks good to me. I would still appreciate if @max-veit could give this a look for the SOAP calculation part, which I am not yet very familiar with.

tests/test_calculator.hh
```diff
@@ -92,6 +92,52 @@ namespace rascal {
     json factory_args{};
   };
 
+  struct TestDataHalfNL {
```
max-veit: Let's mark this for deletion, since full vs half are already compared above

tests/test_calculator.cc
```diff
@@ -569,6 +537,197 @@ namespace rascal {
     }
   }
 
+  /* ---------------------------------------------------------------------- */
+  /**
+   * Utility fixture used to compare representations computed with full and
+   * half neighbor lists.
+   */
+  template <class CaculatorFixtureFull, class CaculatorFixtureHalf>
```
max-veit: spelling (Ca**l**culator)

tests/test_calculator.cc
```diff
@@ -542,9 +542,9 @@ namespace rascal {
    * Utility fixture used to compare representations computed with full and
    * half neighbor lists.
    */
-  template <class CaculatorFixtureFull, class CaculatorFixtureHalf>
-  struct MergeHalfAndFull : CaculatorFixtureFull, CaculatorFixtureHalf {
-    using ParentFull = CaculatorFixtureFull;
+  template <class CalculatorFixtureFull, class CaculatorFixtureHalf>
+  struct MergeHalfAndFull : CalculatorFixtureFull, CaculatorFixtureHalf {
+    using ParentFull = CalculatorFixtureFull;
     using ParentHalf = CaculatorFixtureHalf;
```
max-veit: Missed one! (and in the line above)

max-veit: (and two lines above)

PR #249, opened 2019-12-13
Feat/improve block sparse dot

felixmusil:
Improve the performance of the kernel computation using the coefficients of the spherical invariants through the dot functions of block_sparse_property. The data of a structure is now stored in one array for `block_sparse_property` and `InternallySortedKeyMap` acts only as a view. To allow for the optimization to be triggered 2 new settings have been added to the spherical expansion: `expansion_by_species_method` and  `global_species`. It changes how the representation generates keys and hence the level of sparsity of the block_sparse_properties. 

Here are some preliminary benchmarks using the structures in ../reference_data/inputs/small_molecules-1000.json using r_c=3.5, max_radial=8, max_angular=6, target_type='Structure', zeta=2:

Kernel with user defined species elapsed: 70.0698 seconds
Kernel with structure wise species elapsed: 49.5647 seconds
Kernel with environment wise species elapsed: 226.643 seconds (old)

Getting the features with python and computing something similar (but it should be faster than the exact equivalent) with `np.power(np.dot(X, X.T), zeta).sum()` where X is the dense feature matrix: 28.3 seconds

Direct use of BLAS is still faster but we can't really beat it since we don't want to and can't build the environmental kernel in one Eigen dot product.


src/rascal/representations/calculator_spherical_expansion.hh
```diff
@@ -1165,24 +1165,56 @@ namespace rascal {
...
-      }
-      if (hypers.find("compute_gradients") != hypers.end()) {
+      if (hypers.count("compute_gradients")) {
         this->compute_gradients = hypers.at("compute_gradients").get<bool>();
       } else {  // Default false (don't compute gradients)
         this->compute_gradients = false;
       }
 
+      if (hypers.count("expansion_by_species_method")) {
+        std::vector<std::string> possible_expansion_by_species{
```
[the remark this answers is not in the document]
Luthaf: or even just an `expansion_by_species_tmp == "user defined" || expansion_by_species_tmp == "structure wise" || ...`?

src/rascal/representations/calculator_spherical_expansion.hh
```diff
@@ -1405,13 +1437,32 @@ namespace rascal {
...
     double interaction_cutoff{};
+    //! size of the transition region r_t spanning [r_c-r_t, r_c] in which the
+    //! contributions to the environment expansion go to zero smoothly
     double cutoff_smooth_width{};
+    //! defines the maximal mean error allowed to the interpolator when fitting
+    //! the reference
     double interpolator_accuracy{};
+    //! number of radial basis function to use in the expansion
     size_t max_radial{};
+    /**
```
Luthaf: Not really important, but why use different doc comment styles ? This line uses `/** doc */` and below uses `//! Doc`. I am fine with either, but maybe we should be consistent and stick to one style.

src/rascal/structure_managers/property_block_sparse.hh
```diff
@@ -738,29 +932,126 @@ namespace rascal {
...
 
+    /**
+     * Get a dense feature matrix Ncenter x Nfeatures only if
+     * `this->are_keys_uniform == true`. It only reinterprets the flat data
+     * storage object since all cluster entries have the same number of keys,
+     * i.e. each rows has the same size and has the same key ordering.
+     */
+    MatrixMapConst_Ref_t get_raw_data_view() const {
+      // should not be called if the keys are not uniform
+      assert(this->are_keys_uniform() == true);
```
Luthaf: Should this be an exception instead of an assert? `assert` are usually removed in 'release' mode

src/rascal/structure_managers/property_block_sparse.hh
```diff
@@ -738,29 +932,126 @@ namespace rascal {
...
+    }
+
+    std::array<int, 4> get_block_info_by_key(const Key_t & key) const {
+      SortedKey_t skey{key};
+      return this->get_block_info_by_key(skey);
+    }
+
+    std::array<int, 4> get_block_info_by_key(const SortedKey_t & skey) const {
+      // should not be called if the keys are not uniform
+      assert(this->are_keys_uniform() == true);
```
Luthaf: same as above

bindings/rascal/representations/spherical_invariants.py
```diff
@@ -49,6 +49,32 @@ class SphericalInvariants(object):
         Whether to normalize so that the kernel between identical environments
         is 1.  Default and highly recommended: True.
 
+    expansion_by_species_method : string
+        Specifies the how the species key of the invariant are set-up.
+        Possible values: 'environment wise', 'user defined', 'structure wise'.
```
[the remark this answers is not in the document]
felixmusil: do you think it would not be a little too short and less clear ?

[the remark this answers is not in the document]
Luthaf: It is slightly less clear, but it should be consistent. 

I think `Atom/Structure/User` is fine here, but we could change everything to use `'environment wise', 'user defined', 'structure wise'`

src/rascal/models/kernels.hh
```diff
@@ -148,6 +175,46 @@ namespace rascal {
         return kernel;
       }
 
+      template <class Property_t, internal::TargetType Type,
+                std::enable_if_t<Type == internal::TargetType::Atom, int> = 0,
+                class StructureManagers>
+      math::Matrix_t compute(const StructureManagers & managers_a,
```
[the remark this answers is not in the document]
felixmusil: I added some documentation to the compute functions, is it good now ?

src/rascal/representations/calculator_spherical_expansion.hh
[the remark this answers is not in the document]
felixmusil: Did not thought of the set. I prefer using a list compared to if else because you don't have to update the error statement if you add options.

src/rascal/representations/calculator_spherical_expansion.hh
felixmusil: I use `//!` when its a one liner and `/**` for longer comments. Then when you already have a one liner and you add a bit of text I keep the `//!` when its still 2 lines. I am not sure there is a need to stick to one in particular is there ?

src/rascal/structure_managers/property_block_sparse.hh
felixmusil: Do you think its not sufficient to check only in debug ? If it is not true the results will be messed up so probably seg fault and to one will go to the debug compilation and get the assert. This is supposed to be used only in few places inside the class not anywhere else.

src/rascal/representations/calculator_spherical_expansion.hh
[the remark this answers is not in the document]
felixmusil: did not think of the set.

bindings/rascal/representations/spherical_invariants.py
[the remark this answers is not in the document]
felixmusil: note that in the kernel context User does not make sense.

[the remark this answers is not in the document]
felixmusil: and the meaning behind Atom and Structure is different too for the species definition.

src/rascal/representations/calculator_spherical_expansion.hh
Luthaf: I prefer when the style is consistent, it makes it easier to navigate the codebase. But this is a minor point w.r.t this PR

bindings/rascal/representations/spherical_invariants.py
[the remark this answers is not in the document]
Luthaf: Sure, should we switch everything to your new nomenclature?

src/rascal/structure_managers/property_block_sparse.hh
Luthaf: The problem is that you don't have the guarantee of a segfault, and this means that one can get messed up results, think they are reasonable and go on publishing them. So yes, I would rather check more than less.

felixmusil: I understand your concern but I feel that this is a bit superfluous since wherever it is used, the flag are_keys_uniform() has already been checked so the scenario you describe is not possible at the moment.

Luthaf: >  at the moment.

that's the point: assertions/checks are here to prevent future changes from breaking everything.

bindings/rascal/representations/spherical_invariants.py
```diff
@@ -52,19 +52,24 @@ class SphericalInvariants(object):
...
-        species key associated with atoms within the environement.
-        'structure wise' means that within a structure the keys species key
+        The descriptor is computed for each atomic enviroment and it is indexed
+        using tuples of atomic species that are present within the environment.
+        This index is by definition sparse since a species tuple will be non
+        zero only if the atomic species are present inside the environment.
+        'environment wise' means that each environemental representation
+        will only contain the minimal set of species tuples needed by each
+        atomic environement.
+        'structure wise' means that within a structure the species tuples
```
Luthaf: this is much better!

PR #253, opened 2020-01-27
boost include

felixmusil:
For custom boost path, typically set with BOOST_ROOT and different from /usr of /usr/local, make sure the include path are properly set when compiling the tests.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -36,8 +36,6 @@ target_include_directories(rascal_tests_main SYSTEM PRIVATE ${Boost_INCLUDE_DIR}
 target_compile_definitions(rascal_tests_main PUBLIC -DBOOST_TEST_DYN_LINK)
 target_link_libraries(rascal_tests_main ${Boost_LIBRARIES} "${LIBRASCAL_NAME}")
 
-message(STATUS "+++++++++++ ${Boost_INCLUDE_DIR}")
-
 file(GLOB tests "${CMAKE_CURRENT_SOURCE_DIR}/test_*.cc")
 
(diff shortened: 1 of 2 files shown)


tests/CMakeLists.txt
```diff
@@ -29,18 +29,26 @@ else()
     set(TEST_RUNNER "")
 endif()
 
+
 add_library(rascal_tests_main STATIC helpers/tests_main.cc)
+# make sure all tests include the boost that has been found
+target_include_directories(rascal_tests_main SYSTEM PRIVATE ${Boost_INCLUDE_DIR})
```
Luthaf: If you set this as PUBLIC instead of PRIVATE, then you don't have to set it again below

felixmusil: I feel it is more explicit to add the includes to all test binaries so that it is more explicit.

tests/CMakeLists.txt
```diff
@@ -29,18 +29,26 @@ else()
...
 target_compile_definitions(rascal_tests_main PUBLIC -DBOOST_TEST_DYN_LINK)
 target_link_libraries(rascal_tests_main ${Boost_LIBRARIES} "${LIBRASCAL_NAME}")
 
+message(STATUS "+++++++++++ ${Boost_INCLUDE_DIR}")
```
Luthaf: Can you remove this line?

tests/CMakeLists.txt
Luthaf: It is more explicit, but it also goes against the usual "modern" cmake way of doing this. The whole point of targets & targets properties is to remove the repetition of include/compiler flags required.

PR #254, opened 2020-01-27
change and refactor invariants gradients computation

felixmusil:
Gradients in SphericalInvariants and SphericalExpension where stored like this for a center i:
+ \grad_i p^i
+ \grad_j p^i for any j neighbors of center i

This form is not convenient for computing forces with a kernel model so this PR modifies the storage into:
+ \grad_i p^i
+ \grad_i p^j for any j neighbors of center i

Also the gradients  \grad_j p^i where not properly computed before because the contributions from periodic images of j that could be inside the environment of i where not taken into account. The gradient tests where passing before because this issue was corrected there so now gradient tests have been simplified and the terms \grad_i p^j are properly computed.


reference_data/tests_only/spherical_expansion_gradient_test.json
```diff
@@ -2,9 +2,9 @@
     "direction_mode": "Cartesian",
     "comment": "Symmetry directions below for diamond to separate radial and angular components (set direction_mode to Provided to use)",
     "displacement_directions": [
-      [1.0, 1.0, 1.0],
-      [1.0, 0.0, -1.0],
-      [-1.0, 2.0, -1.0]
+      [1.0, 0, 0],
+      [0, 1, 0],
+      [0, 0, 1]
```
max-veit: This is unnecessary; the `"direction_mode": "Cartesian"` above already takes care of this

src/rascal/representations/calculator_spherical_expansion.hh
```diff
@@ -1636,8 +1636,15 @@ namespace rascal {
         AdaptorTraits::NeighbourListType::half};
     using math::PI;
     using math::pow;
-
     constexpr bool ExcludeGhosts{true};
+    constexpr static size_t ClusterLayer{
+        StructureManager::template cluster_layer_from_order<2>()};
+    const bool are_some_centers_masked{manager->are_some_centers_masked()};
```
max-veit: It would make more sense to name this `are_any_centers_masked()`, since that corresponds more closely to the underlying logical operation

src/rascal/representations/calculator_spherical_expansion.hh
```diff
@@ -1659,6 +1666,7 @@ namespace rascal {
             this->radial_integral)};
 
     auto n_row{this->max_radial};
+    // to store linearly all l,m components with -l<=m<=l needs l**2 elements
```
max-veit: Not l**2, (l+1)**2. I think this is documented in a doc-comment somewhere.

src/rascal/representations/calculator_spherical_expansion.hh
```diff
@@ -1716,35 +1729,34 @@ namespace rascal {
...
-            std::stringstream err_str{};
-            err_str << "Half neighbor list should only be used when all the "
-                    << "atoms inside the unit cell are centers, i.e. "
-                    << "center_atoms_mask should not mask atoms.";
-            throw std::runtime_error(err_str.str());
+        // compute the coefficients
+        // complications here because the infrastructure for swaping ij is not
+        // here yet.
+        // TODO(felix) make swaping ij possible and implement the
+        // proper half for gradients so the whole computation can be skipped.
```
max-veit: So is half neighbourlist for the gradients turned off now? I remember discussing such a swap-ij function, but I thought it'd be part of this PR. Did you not end up needing it?

In any case, let's turn this into an issue so as not to forget about it.

src/rascal/representations/calculator_spherical_expansion.hh
```diff
@@ -1716,35 +1729,34 @@ namespace rascal {
...
+        if (not are_some_centers_masked and atom_i_tag > atom_j_tag) {
+          // continue;
+        } else {
```
max-veit: Is the `// continue` a placeholder for anything you want to implement later? Otherwise replacing the if-else with just `if (are_any_centers_masked or (atom_i_tag <= atom_j_tag)) {` (the negation of the current statement) would be clearer.

Also, please use more parentheses around the negation and the comparison; it helps readability and avoids confusion due to having to remember C++'s order of operations.

src/rascal/representations/calculator_spherical_expansion.hh
max-veit: Btw, if this is half-neighbourlist code, where did the `IsHalfNL` test go? Is it now in `are_any_centers_masked`?

src/rascal/representations/calculator_spherical_expansion.hh
```diff
@@ -1716,35 +1729,34 @@ namespace rascal {
...
+          // when j == i there is nothing to add to ji
+          if (not are_some_centers_masked and
+              (atom_j_tag != atom_i_tag)) {  // NOLINT
+            // half list branch for c^{ji} terms using
+            // c^{ij}_{nlm} = (-1)^l c^{ji}_{nlm}.
             auto & coefficients_neigh{expansions_coefficients[atom_j]};
             auto coefficients_neigh_by_type{coefficients_neigh[center_type]};
             l_block_idx = 0;
-            double parity{1.};
+            double parity{1.};  // account for (-1)^l
```
max-veit: The comment isn't very helpful, since it just repeats the code. Either take it out or explain that it comes from swapping i and j (and thereby negating `\hat{r}` as the argument to the spherical harmonics)

src/rascal/representations/calculator_spherical_expansion.hh
```diff
@@ -1764,35 +1776,43 @@ namespace rascal {
...
         if (this->compute_gradients and (atom_j_tag != atom_i_tag)) {  // NOLINT
+          // \grad_i c^j
           auto & coefficients_neigh_gradient =
               expansions_coefficients_gradient[neigh];
 
           auto && neighbour_derivative =
               radial_integral->compute_neighbour_derivative(dist, neigh);
           double df_c{cutoff_function->df_c(dist)};
           // The gradients only contribute to the type of the neighbour
-          // (the atom that's moving)
```
max-veit: Perhaps rephrase comment -- in any case, the gradients only contribute to the type of atom that's moving (i.e. w.r.t. which the derivative is being taken), right? This mixed-type notation is confusing me, but I'm assuming the multi-species tests pass...?

src/rascal/representations/calculator_spherical_expansion.hh
```diff
@@ -1805,17 +1825,23 @@ namespace rascal {
...
+                  max_radial, l_block_size) = parity *
+                    pair_gradient_contribution;
               l_block_idx += l_block_size;
+              parity *= -1.;
               // clang-format on
             }  // for (angular_l)
           }    // for cartesian_idx
 
+          // TODO(felix): This block should be refactored after swapping ij is
+          // possible
```
max-veit: Again reference to issue (same one as comment above) so we can keep track of it

src/rascal/representations/calculator_spherical_expansion.hh
```diff
@@ -1842,10 +1868,75 @@ namespace rascal {
                   l_block_idx += l_block_size;
                 }  // for (angular_l)
               }    // for cartesian_idx
-            }
-          }  // if (IsHalfNL)
-        }    // if (this->compute_gradients)
-      }      // for (neigh : center)
+            }      // if (is_center_atom)
+          }        // if (IsHalfNL)
```
max-veit: The closing if-statement comments no longer match the opening conditions, please update

src/rascal/representations/calculator_spherical_expansion.hh
```diff
@@ -1842,10 +1868,75 @@ namespace rascal {
...
+        // periodic images.
+        std::map<int, std::vector<ClusterRefKey<2, ClusterLayer>>>
+            periodic_images_of_center{};
+
+        // find all center atoms within the environment of i
+        for (auto pair : center.pairs()) {
+          auto atom_j = pair.get_atom_j();
+          int atom_tag_j = atom_j.get_atom_tag();
+          if (not manager->is_ghost_atom(pair)) {
+            periodic_images_of_center[atom_tag_j].emplace_back(
```
max-veit: Can the finding of periodic images of the center be encapsulated into its own function? That way it can be used again for other descriptors, since this seems like a fairly general-purpose operation.

src/rascal/representations/calculator_spherical_expansion.hh
felixmusil: For spherical invarients gradients we don't want to have a halfNL since \grad_i p^j =/= \grad_j p^i. So I removed the IsHalfNL test and made a static assert about it not being compatible in spherical invarients. Then we don't have HalfNL but we still want to do half of the work in spherical expansion, the problem is that for gradients we need a swap ij functionality to make it work. This is not part of this PR but I am currently working on such implementation in `feat/swap_pair`.

src/rascal/representations/calculator_spherical_expansion.hh
felixmusil: when the swap ij functionality is there (which is soon) there will be a 
```
        if (not are_some_centers_masked and atom_i_tag > atom_j_tag) {
          continue;
        }
```
at the very top of the iteration over the pairs so writing it like this make the future change simpler.

src/rascal/representations/calculator_spherical_expansion.hh
felixmusil: it does not repeat the code to me since parity only holds for (-1)^l because there is a 
```parity *= -1.;``` 
later in the loop.

File: src/rascal/representations/calculator_spherical_expansion.hh (hunk 7 above)
Comment (felixmusil) [reply src/rascal/representations/calculator_spherical_expansion.hh]: it passes the tests yes.

File: src/rascal/representations/calculator_spherical_expansion.hh (hunk 9 above)
Comment (felixmusil) [reply src/rascal/representations/calculator_spherical_expansion.hh]: it does to me. Could you check again or be more precise ?

File: src/rascal/representations/calculator_spherical_expansion.hh (hunk 10 above)
Comment (felixmusil) [reply src/rascal/representations/calculator_spherical_expansion.hh]: you mean by own function, a function external to the class ?

File: src/rascal/representations/calculator_spherical_expansion.hh (hunk 9 above)
Comment (max-veit) [reply src/rascal/representations/calculator_spherical_expansion.hh]: Oh, the `IsHalfNL` check didn't actually disappear, it just got moved. Sorry, it didn't show up properly in the diff view. But why is the `IsHalfNL` test still here, when you said that the option now marked as incompatible with this representation?

File: src/rascal/representations/calculator_spherical_expansion.hh (hunk 10 above)
Comment (max-veit) [reply src/rascal/representations/calculator_spherical_expansion.hh]: Yes, so that other representations can use such functionality if they need it.

File: src/rascal/structure_managers/structure_manager_centers.hh [hunk 11]
```diff
@@ -304,6 +306,11 @@ namespace rascal {
       return this->atoms_object;
     }
 
+    bool are_any_centers_masked() const {
+      // center_atoms_mask is true is no center are masked
+      return (not this->get_center_atoms_mask().all());
```
mastricker: Yes, here it is instead of the evaluation - which is possibly a very big number set it at construction.
I am not sure if it is necessary and how often it is evaluated, just a thought.

src/rascal/structure_managers/structure_manager_lammps.hh
```diff
@@ -180,6 +182,8 @@ namespace rascal {
      */
     size_t get_nb_clusters(int order) const;
 
+    bool are_any_centers_masked() const { return false; }
```
mastricker: As a side note: is the masking only for training?

tests/test_calculator.cc
```diff
@@ -491,6 +497,10 @@ namespace rascal {
       CalculatorFixture<
           SingleHypersSphericalInvariants<SimplePeriodicNLCCStrictFixture>>>;
 
+  // using gradient_fixtures = boost::mpl::list<
```
mastricker: Can this go away?

tests/test_calculator.cc
```diff
@@ -530,210 +540,12 @@ namespace rascal {
...
       }
       ++filename_it;
     }
   }
 
-  /* ---------------------------------------------------------------------- */
-  /**
-   * Utility fixture used to compare representations computed with full and
-   * half neighbor lists.
-   */
```
mastricker: Where did this go? Or is it just not used anymore?

tests/test_calculator.hh
```diff
@@ -569,9 +517,12 @@ namespace rascal {
         {{"type", "Constant"},
          {"gaussian_sigma", {{"value", 0.4}, {"unit", "AA"}}}}};
     std::vector<json> radial_contribution_hypers{{{"type", "GTO"}}};
+    // std::vector<json> rep_hypers{
+    //     {{"max_radial", 2}, {"max_angular", 2}, {"compute_gradients", true}},
+    //     {{"max_radial", 3}, {"max_angular", 0}, {"compute_gradients",
+    //     true}}};
     std::vector<json> rep_hypers{
-        {{"max_radial", 2}, {"max_angular", 2}, {"compute_gradients", true}},
-        {{"max_radial", 3}, {"max_angular", 0}, {"compute_gradients", true}}};
```
mastricker: Is this the reason the test got faster? Less computations?

src/rascal/representations/calculator_spherical_expansion.hh
max-veit: Perhaps writing out the sum would make it less confusing, something like `grad_i c^{ib} = sum_j (-1)^l grad_j c^{jb}`?

src/rascal/representations/calculator_spherical_expansion.hh
max-veit: Ok, then maybe just call the `(-1)^l` term in the comment 4 lines up a "parity term" explicitly, so that there's the link to the variable name below and you don't have to explain it with an extra comment.

src/rascal/representations/calculator_spherical_expansion.hh
```diff
@@ -1683,28 +1752,33 @@ namespace rascal {
...
 
-      auto atom_i_tag = center.get_atom_tag();
-      // coeff C^{ij}_{nlm}
-      auto c_ij_nlm = math::Matrix_t(n_row, n_col);
-
       for (auto neigh : center.pairs()) {
-        auto && atom_j = neigh.get_atom_j();
-        auto atom_j_tag = atom_j.get_atom_tag();
-        bool is_center_atom{manager->is_center_atom(neigh)};
+        auto atom_j = neigh.get_atom_j();
```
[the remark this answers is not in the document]
felixmusil: `atom_j` is created so `&&` was effectively just an `auto`.

src/rascal/representations/calculator_spherical_expansion.hh
```diff
@@ -1764,35 +1837,45 @@ namespace rascal {
...
           auto && gradient_center_by_type{
               coefficients_center_gradient[neigh_type]};
-          // grad_j c^{i}
+          // grad_i c^{ja}
           auto && gradient_neigh_by_type{
-              coefficients_neigh_gradient[neigh_type]};
+              coefficients_neigh_gradient[center_type]};
 
-          // Radial component: d/dr_{ij} (c_{ij} f_c{r_{ij}}) \hat{r_{ij}}
           // clang-format off
```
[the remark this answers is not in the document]
felixmusil: @max-veit did this off. I think it is readable like it is and clang-format can only make worse or as good.

src/rascal/representations/calculator_spherical_expansion.hh
```diff
@@ -1764,35 +1837,45 @@ namespace rascal {
...
-          // grad_i c^{i}
+          // The type of the contribution c^{ij} to the coefficient c^{i}
+          // depends on the type of j (and it is the same for the gradients)
+          // In the following atom i is of type a and atom j is of type b
+
+          // grad_i c^{ib}
           auto && gradient_center_by_type{
               coefficients_center_gradient[neigh_type]};
-          // grad_j c^{i}
+          // grad_i c^{ja}
```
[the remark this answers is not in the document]
felixmusil: type of atom i. It is mentioned a few lines above.

src/rascal/representations/calculator_spherical_expansion.hh
max-veit: Ok. And I really do prefer putting parentheses around the operands of a boolean operator, both because it avoids anyone reading this as `if not (are any centers masked and atom_i_tag > atom_j_tag)` (that's how I first read this) and having to remember that both `not` and `>` are higher in this table: https://en.cppreference.com/w/cpp/language/operator_precedence than the `and` operator.

From the coding convention, section https://cosmo-epfl.github.io/librascal/dev_guide/coding-convention.html#boolean-expressions:

> Feel free to insert extra parentheses judiciously because they can be very helpful in increasing readability when used appropriately.

src/rascal/representations/calculator_spherical_expansion.hh
```diff
@@ -1764,35 +1837,45 @@ namespace rascal {
...
-          // Radial component: d/dr_{ij} (c_{ij} f_c{r_{ij}}) \hat{r_{ij}}
           // clang-format off
+          // d/dr_{ij} (c_{ij} f_c{r_{ij}})
           Matrix_t pair_gradient_contribution_p1 =
                 ((neighbour_derivative * f_c)
                  + (neighbour_contribution * df_c));
+          // grad_j c^{ij}
           Matrix_t pair_gradient_contribution{this->max_radial,
                                               this->max_angular + 1};
           for (int cartesian_idx{0}; cartesian_idx < n_spatial_dimensions;
```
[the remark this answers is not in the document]
felixmusil: changed to `NDims`

src/rascal/representations/calculator_spherical_expansion.hh
```diff
@@ -1764,35 +1837,45 @@ namespace rascal {
         // (the periodic images move with the center, so their contribution to
         // the center gradient is zero)
         if (this->compute_gradients and (atom_j_tag != atom_i_tag)) {  // NOLINT
+          // \grad_i c^j
           auto & coefficients_neigh_gradient =
               expansions_coefficients_gradient[neigh];
 
           auto && neighbour_derivative =
               radial_integral->compute_neighbour_derivative(dist, neigh);
           double df_c{cutoff_function->df_c(dist)};
```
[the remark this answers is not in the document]
felixmusil: its evaluated only once and potentially used several times

src/rascal/representations/calculator_spherical_expansion.hh
```diff
@@ -1842,10 +1931,39 @@ namespace rascal {
...
+
+      // In the case of having several periodic images of other centers in
+      // the environement of center i, we need to sum up
+      // of all these contributions, e.g. ij, ij', ij'' ... where
+      // j primes are periodic images of j. And assign this sum back to all
+      // of these terms.
+      if (this->compute_gradients) {
+        std::map<int, std::vector<ClusterRefKey<2, ClusterLayer>>>
+            periodic_images_of_center =
+                internal::find_periodic_images_pairs_in_environment(manager,
```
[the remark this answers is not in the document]
felixmusil: I feel it is overkill because `periodic_images_of_center` (the result of this adaptor) would be used only once so there are no advantages.

src/rascal/representations/calculator_spherical_invariants.hh
```diff
@@ -308,6 +311,85 @@ namespace rascal {
...
+            soap_vector_dot_gradient += (soap_gradient_dim_N * soap_vector_N);
+          }  // for (const auto& key : keys_intersect)
+
+          // Now update each species-pair-block using the dot-product just
+          // computed
+          for (const auto & key : keys_intersect) {
+            auto soap_gradient_by_species_pair =
+                soap_vector_gradients_by_neigh[key];
+            const auto & soap_vector_by_species_pair = soap_vector[key];
+            // reshape for easy dot prod
```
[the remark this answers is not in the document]
felixmusil: its a view.

src/rascal/representations/calculator_spherical_invariants.hh
```diff
@@ -395,6 +476,12 @@ namespace rascal {
     // using operator[] of soap_vector
     internal::SortedKey<Key_t> spair_type{pair_type};
 
+    const size_t n_n1n2{math::pow(this->max_radial, 2_size_t)};
+    // to store the norm of the soap vectors
+    SpectrumNorm_t<StructureManager> soap_vector_norm_inv{
+        *manager, "power spectrums inverse norms", true};
```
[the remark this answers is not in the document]
felixmusil: I am not sure we need to formalize the metadata of the Properties (here soap_vector_norm_inv is not stored in the manager).

src/rascal/representations/calculator_spherical_invariants.hh
```diff
@@ -440,122 +527,77 @@ namespace rascal {
...
+      if (this->compute_gradients) {
+        const int atom_i_tag{center.get_atom_tag()};
+
+        // Sum the gradients wrt the neighbour atom position
+        // compute the \grad_i p^{k} coeffs where k is either i or j
+        for (auto neigh : center.pairs_with_self_pair()) {
+          auto && atom_j = neigh.get_atom_j();
+          const int atom_j_tag{atom_j.get_atom_tag()};
+          const bool is_center_atom{manager->is_center_atom(neigh)};
+          // compute grad contribution only if the neighbour is _not_ an
```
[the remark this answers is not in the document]
felixmusil: The filtering `(atom_j_tag != atom_i_tag)` is only useful for gradients so you actually want them when computing the coefficients. If it is implemented inside an adaptor then we would have to come up with a get_center_periodic_images_pairs() function that keeps the distances and direction vectors for every center. That seems a bit too much complications compared to the gains.

src/rascal/representations/calculator_spherical_invariants.hh
```diff
@@ -440,122 +527,77 @@ namespace rascal {
...
+            const auto & grad_neigh_coefficients_1{
+                grad_neigh_coefficients[coef_key_1]};
+
+            for (const auto & coef_key_2 : keys_coef_j) {
+              // c^{k b}
+              const auto & expansion_coefficients_j_2{
+                  coefficients_j[coef_key_2]};
+              bool sorted{true}, equal{false};
+              // make sure spair_type has sorted entries
+              if (coef_key_1[0] > coef_key_2[0]) {
```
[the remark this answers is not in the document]
felixmusil: there is a `(a,b)` index to the power spectrum gradients with `a` and `b` atomic types but the coefs in `(a,b)` and `(b,a)` are the same so we only compute and keep  `(a,b)` with `a < b`. 
There are 3 cases in this loop resulting in different computation load:
+  `a < b` the first block
+  `a > b` the second block
+ `a == b` both 1st and 2nd block

so the `if` here are saving quite a bit of computation.  I don't think there is a point to precomputing the branching.

src/rascal/representations/calculator_spherical_invariants.hh
```diff
@@ -440,122 +527,77 @@ namespace rascal {
...
+                equal = true;
+                spair_type[0] = coef_key_1[0];
+                spair_type[1] = coef_key_2[0];
+              } else {
+                spair_type[0] = coef_key_1[0];
+                spair_type[1] = coef_key_2[0];
               }
 
+              // \grad_i p^{k ab}
               auto soap_neigh_gradient_by_species_pair{
```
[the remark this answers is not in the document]
felixmusil: here the auto stands for an Eigen::Map<Matrix> so there is no copy.

src/rascal/structure_managers/adaptor_filter.hh
```diff
@@ -126,6 +126,8 @@ namespace rascal {
     using ParentBase = FilterBase;
     using ManagerImplementation_t = ManagerImplementation;
     using ImplementationPtr_t = std::shared_ptr<ManagerImplementation>;
+    using ConstImplementationPtr_t =
```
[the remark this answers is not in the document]
felixmusil: it is useful to keep the const correctness when using the get_previous_manager() function in StructureManager.

src/rascal/structure_managers/adaptor_strict.hh
```diff
@@ -225,12 +227,15 @@ namespace rascal {
...
-    // if this was wrong, please explain
+    //! Get the manager used to build the instance
+    ConstImplementationPtr_t get_previous_manager_impl() const {
+      return this->manager->get_shared_ptr();
+    }
+
     //! returns the distance between atoms in a given pair
     template <size_t Order, size_t Layer>
-    double get_distance(const ClusterRefKey<Order, Layer> & pair) const {
+    const double &
```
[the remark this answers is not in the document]
felixmusil: > Btw: we changed distance property a little bit on our side: it is now registered with the manager. Does not change the interface to the distances here. Just saying...

This is not a problem to me since it is already the case: a shared pointer in adaptor_strict and another one in the manager's map.

src/rascal/structure_managers/property_block_sparse.hh
```diff
@@ -430,12 +430,13 @@ namespace rascal {
       /**
        * squared l^2 norm of the entire vector (sum of squared elements)
        */
-      void normalize() {
+      double normalize() {
```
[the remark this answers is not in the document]
felixmusil: `normalize_and_get_norm` seems a bit of a mouthful, you sure you want this name ?

src/rascal/structure_managers/property_block_sparse.hh
```diff
@@ -430,12 +430,13 @@ namespace rascal {
...
         double norm{this->norm()};
         auto block{this->get_full_vector()};
```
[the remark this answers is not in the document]
felixmusil: its an Eigen::Map so no copy.

src/rascal/structure_managers/property_block_sparse.hh
```diff
@@ -969,6 +970,12 @@ namespace rascal {
       return arr;
     }
 
+    double sum() const { return this->values.sum(); }
+
+    double l1_norm() const {
```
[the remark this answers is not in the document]
felixmusil: in Eigen most reductions like `sum` return the value so I am following their example. Also we use `pairs()` instead of `get_pairs()` in our code so it seems consistent. What do you think ?

src/rascal/structure_managers/structure_manager.hh
```diff
@@ -624,13 +627,17 @@ namespace rascal {
...
-    inline const Vector_ref
-    get_direction_vector(const ClusterRefKey<Order, Layer> & pair) {
+    const Vector_ref
+    get_direction_vector(const ClusterRefKey<Order, Layer> & pair) const {
       static_assert(HasDirectionVectors == traits::HasDirectionVectors,
                     "The manager does not have direction vectors.");
       return this->get_previous_manager()->get_direction_vector(pair);
     }
 
+    bool are_any_centers_masked() const {
```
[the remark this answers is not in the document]
felixmusil: what do you mean by 'by construction' ?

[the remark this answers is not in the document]
max-veit: I don't know, I think the name is as descriptive as it needs to be, given how specific the condition is that it needs to test. And a function name being literal is not a disadvantage: It tells you exactly what it does, which is the best kind of function name.

src/rascal/structure_managers/structure_manager_lammps.hh
felixmusil: the masking is for properties that are different from the energy like transfert integral between dimer molecules. So both for training and predictions.

src/rascal/representations/calculator_spherical_expansion.hh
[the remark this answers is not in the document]
mastricker: Ok, I tried it, it definitely looks less clear. Let's hope for the best. We can change it if it leads to noise.

src/rascal/representations/calculator_spherical_expansion.hh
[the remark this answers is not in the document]
mastricker: Hmm. Well, ok. But if you evaluate on the same structure, it is computed again.
Let's file it under _possible premature optimization_.

src/rascal/representations/calculator_spherical_invariants.hh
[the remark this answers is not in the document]
mastricker: Ok - is there (from heuristics) an order of those conditions which leads to the most common as the first, etc. Of course, it depends on the structure and the different atom types.
But given the experience, is there a more efficient order?

In the behler-implementation we do precompute those combinations.
It plays into my concern that the calculator does indeed adapt the structure further - but that also changes (possibly) how the data is stored. I'll make a longer issue out of it for future reference later.

src/rascal/representations/calculator_spherical_invariants.hh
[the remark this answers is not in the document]
mastricker: But does it hurt to put the `&` - as a guide so when you look at it, it already suggests that it is a reference?

src/rascal/structure_managers/property_block_sparse.hh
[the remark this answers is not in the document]
mastricker: I am not passionate - just suggesting a different name.
And I am a little confused: in `calculator_spherical_expansions.hh` is a call
```
      // normalize the soap vector
      if (this->normalize) {
        soap_vector.normalize();
      }
```
where the return is not assigned?
Is that a part that is never compiled?

src/rascal/structure_managers/property_block_sparse.hh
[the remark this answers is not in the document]
mastricker: Again, I would argue for visual aid with `&`

src/rascal/structure_managers/structure_manager.hh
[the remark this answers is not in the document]
mastricker: Ok, agreed on the name.
But for construction: when the manager is constructed - I see this has been changed.

mastricker:
Ok, most of the stuff is cleared for me.
On thing I would add as an issue to keep in mind - not change it right away - is the fact that the `calculator_spherical_*` does some sorting/adapting on its own during computation. And I think that brakes our logic a little bit of the calculator being a calculator and an adaptor/manager being a filter.
This probably does play into the logic of the storage - though I'm not sure. Properties supplied to the calculator do should be writeable at their respective layer, regardless.

Would you be ok with me making an issue for this though for future reference?

tests/test_calculator.hh
felixmusil: I am not sure actually

src/rascal/representations/calculator_spherical_expansion.hh
```diff
@@ -59,6 +59,66 @@ namespace rascal {
 
   namespace internal {
 
+    /**
+     * Utility to find the periodic images of the center atoms that are within
```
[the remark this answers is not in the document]
felixmusil: I agree with you that it could be an improvement but I think this would make the computation procedure quite redundant or fuzzy to understand. Also there are a few things that will have to be thought through very well because it might just not be worth it.

src/rascal/representations/calculator_spherical_invariants.hh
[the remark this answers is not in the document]
felixmusil: you are right we could do this shuffling ahead of time but I am not convinced it matters so much since the computation inside the ifs are 'heavy' and I am sure that it is a lot of work to do properly.

src/rascal/representations/calculator_spherical_invariants.hh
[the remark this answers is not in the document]
felixmusil: it's not a reference to an Eigen::Map but an Eigen::Map which is by definition a reference. It's the annoying thing yet essential feature of Eigen.

src/rascal/structure_managers/property_block_sparse.hh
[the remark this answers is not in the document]
felixmusil: when you call a function that returns something you don't have to assign the returned value to a variable and yes this compiles and runs.

felixmusil:
> And I think that brakes our logic a little bit of the calculator being a calculator and an adaptor/manager being a filter.
This probably does play into the logic of the storage - though I'm not sure.
Would you be ok with me making an issue for this though for future reference?

I agree that it goes against the loop design but the implementation of adaptors dedicated (and super specialized so no reuse) to this is a lot of work and will make changing how this computation is done much harder to maintain, change and understand I think. So feel free to make an issue just I am not sure how feasible/desirable such changes would be.

ceriottm:
Seems like something that needs discussing together. My gut feeling is that
this actually relates deeply to the "cluster SOAP vs Fourier SOAP" issue.
Are we computing these as a sum over neighbors (a "species exploder" is
clearly the right way to go) or as an outer product of components (perhaps
a specialization is needed). Please do open an issue and add as a prominent
discussion point for a plenary rascal get-together!

src/rascal/representations/calculator_spherical_expansion.hh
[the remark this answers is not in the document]
max-veit: Yeah, clang-format is terrible for equations in general. I'd never trust an autoformatter to do better than a human at clarifying the structure of some dense mathematical code.

src/rascal/representations/calculator_spherical_invariants.hh
```diff
@@ -308,6 +311,84 @@ namespace rascal {
...
+     * resulting in \grad_i \tilde{p}^{k}.
+     * We have:
+     * \grad_i \tilde{p}^{k} = \grad_i p^{k} / N_k
+             - \tilde{p}^{k} [\tilde{p}^{k} \cdot \grad_i p^{k} / N_k],
+     * where $\cdot$ is a dot product between vectors.
+     * Note that this expects the soap vectors to be normalized already, and
+     * the norm stored separately
+     */
+    template <class StructureManager, class SpectrumNorm>
+    void normalize_gradients(
```
max-veit: This naming isn't great -- it implies that the gradients are being divided by their norms, when the actual operation (written above) is more complicated. I'd prefer something like `update_gradients_for_normalization`.

src/rascal/representations/calculator_spherical_invariants.hh
```diff
@@ -1237,7 +1109,7 @@ namespace rascal {
           InvariantsDerivative & soap_vector_gradients,
           ExpansionCoeff & expansions_coefficients,
           std::shared_ptr<StructureManager> manager) {
-    constexpr static int n_spatial_dimensions = StructureManager::dim();
+    constexpr static int NDims = StructureManager::dim();
```
max-veit: I'm honestly not a fan of this renaming back to NDims, because that could refer to any kind of dimension -- the descriptor dimension, for instance -- and the previous name made it explicit that this was _spatial_ dimensions (something that's not likely to change). Just a thought.

src/rascal/structure_managers/property_block_sparse.hh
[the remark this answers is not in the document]
max-veit: See, this is the type of confusion that happens when you have a function whose name implies it _does_ something (`normalize`) and in fact it also _gets_ something -- usually, these two purposes are separated into two different functions (compare Python `sort()`, which _does_ something -- sorts a list, in-place if possible -- and the builtin `sorted()`, which _returns_ a sorted copy of the list).

So even if `normalize_and_return_norm` is long, it's an accurate name for a double-barreled function such as this. (And really, it's returning the norm of the previous, unnormalized vector, so calling this function twice will give different results. Something to keep in mind...)

src/rascal/representations/calculator_spherical_invariants.hh
felixmusil: renamed NDims to ThreeDims

felixmusil:
> Just one question concerning the gradient-of-normalized-SOAP code: Was the previous implementation wrong, or just unsuited to the current ordering of the gradient storage? Because if it was wrong, I would be very interested to know how the tests were passing. But otherwise, the new implementation seems quite a bit simpler and easier to understand.

I had to reformulate the gradient-norm computation for it to work with the new ordering so the previous implementation was working fine.

src/rascal/representations/calculator_spherical_expansion.hh
```diff
@@ -1805,54 +1909,78 @@ namespace rascal {
...
 
-          // half list branch for computing grad_j c^{j} using
-          // grad_j c^{ji} = (-1)^{l} grad_j c^{ij}
+          // half list branch for accumulating parts of grad_j c^{j} using
+          // grad_j c^{ji a} = - grad_i c^{ji a}
           if (IsHalfNL) {
             if (is_center_atom) {
+              // grad_j c^{j}
               auto & coefficients_neigh_center_gradient =
                   expansions_coefficients_gradient[neigh.get_atom_jj()];
```
max-veit: Is this all that was needed to get the half NL accumulation to work?

src/rascal/representations/calculator_spherical_invariants.hh
max-veit: That wouldn't have been my first choice, but I'll take it.

src/rascal/representations/calculator_spherical_expansion.hh
```diff
@@ -552,14 +613,13 @@ namespace rascal {
         coefficients.lhs_dot(this->ortho_norm_matrix);
       }
 
-      template <int n_spatial_dimensions, typename Coeffs, typename Center>
+      template <int ThreeDims, typename Coeffs, typename Center>
```
mastricker: I like this - but would rename it to `ThreeD`?

src/rascal/representations/calculator_spherical_expansion.hh
felixmusil: yes because I throw an error when `IsHalfNL == true` and `(2*cutoff > cell_lengths).any()` which ensures that no center can have in its neighborhood another center and a periodic image associated to it.

PR #261, opened 2020-02-14
Add a CI builder to collect coverage

Luthaf:
Code coverage measure how much of the source code is covered by the tests. While it is not a perfect way to ensure that the code is bug-free, tracking coverage allows to know if some part of the code are not tested at all.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -151,20 +151,25 @@ jobs:
           name: Install Python dependencies
           command: |
             python3.6 -m pip install -r requirements.txt
-            python3.6 -m pip install codecov
+            python3.6 -m pip install codecov coverage
       - run:
           name: Configure
           command: |
             mkdir build && cd build
-            cmake -DBUILD_BINDINGS=OFF -DBUILD_TESTS=ON \
+            cmake -DBUILD_BINDINGS=ON -DBUILD_TESTS=ON \
                   -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_FLAGS="-coverage" \
                   ..
       - run:
           name: Build
           command: cd build && make
       - run:
-          name: Run tests
+          name: Run C++ tests
           command: cd build && ctest
+      - run:
+          name: Run Python tests
+          command: coverage run tests/python/python_binding_tests.py
+          environment:
+            PYTHONPATH: build/
       - run:
           name: Collect and upload coverage
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -141,6 +141,37 @@ jobs:
           command: |
             cd build && make pretty-cpp && git diff --exit-code
 
+  # Special job collecting code coverage
+  coverage:
+    docker:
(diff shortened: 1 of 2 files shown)


Luthaf:
Looks like current coverage is 82% of the lines: https://codecov.io/gh/cosmo-epfl/librascal/tree/e00ccb5c2977292af79a5e122129b53ae9c18901/src/rascal. I set the target coverage to this to prevent coverage from going down in the future.

felixmusil:
I looked at the report on some of our files and I am not sure how to interpret the results of this tool. 
For instance many of the miss I see are std::error* that are not raised in the tests. So does that mean we need to test for every error that we raise ?

Luthaf:
> For instance many of the miss I see are std::error* that are not raised in the tests. So does that mean we need to test for every error that we raise ?

We only need to if we want 100% coverage, but that is not a good metric to use. So I would say no, we don't have to make tests for each error.

On the other hand, having tests that check that the right error is thrown when expected is always a good idea, especially from an user interface perspective.

> Sometimes we opted for covering tests through the python side rather than directly testing c++ code. It is not clear to me from the output if that is also being checked? My understanding is 'no'. 

You are right, and I should add the Python tests to the coverage! I'll check to see how we can combine C++ and Python coverage.

> Partial code coverage (yellow) is often indicated for the for (...) statement. Although all cases inside are covered. I am not sure what that means.

I don't know specially about for loops, but partial coverage is usually indicating that some implicit branches/code path are not taken, in particular branches coming from exceptions. Any specific examples?

mastricker:
Regarding the `for (...)` constructs. I looked over several files, one specific example would be 
[adaptor_neighbor_list](https://codecov.io/gh/cosmo-epfl/librascal/src/e00ccb5c2977292af79a5e122129b53ae9c18901/src/rascal/structure_managers/adaptor_neighbour_list.hh).
At first glance _all_ the `for (...)` constructs have the yellow, with or without additional `if (...)` statements. And all of the `if` cases seem to be covered, because the actual lines which do something after the `if` are green, but the `if` statement lines themselves not.

Luthaf:
Ok, this is slightly strange. Maybe this means that the "don't loop" case is not executed (i.e. `for(int i=0; i<n; i++)` with n = 0)?

Or we would have to dig a bit more into GCC coverage instrumentation to understand what is going on.

Luthaf:
I added Python code coverage. The total coverage is slightly lower (https://codecov.io/gh/cosmo-epfl/librascal/tree/95b04e42da1a1c099f6a6d6b3c0bd691f53aeaa3), so I also updated the minimal coverage requirement.

mastricker:
Ok, thanks. I would say it is good then?
We can also postpone the merge until after the meeting - maybe there are further ideas?

I'd say merge and make a new PR if necessary? Fine with me either way.

Luthaf:
i think we can merge this and update as needed later =)

mastricker:
Go ahead...

PR #262, opened 2020-02-18
implementation of a sparse kernel

felixmusil:
This PR brings a few new features to the code to enable sparse GPR training with energy and forces and predictions of  energy and forces using rascal and Spherical Invariants:
1. a Sparse Kernel class that mimics GAP
2. a Pseudo points class tailored for models 
3. python krr model with GAP trainer

An example of building an interatomic potential for a small molecule is in MLIP_example.ipynb

1) The class is meant to compute kernel matrices that are used in the sparse GPR methods like Subset of Regressors. It is pretty straight forward and most details are in the doc of the class and maybe a more comprehensive descriptions in the docstring of the train_gap_model() function.

2) Pseudo points are key to many sparse GPR models, they basically corresponds to interpolation points in the regression methods. It aims at reducing the cost of training and prediction from N^3/N^2 to ~MN^2/M^2 where N and M are respectively the number of training and pseudo points. We typically choose pseudo points from the training set though.

3) the GAP training is a particular procedure defined in QUIP. It uses sparse kernels and a particular kernel to train with energy/forces and predicts energy/forces. I am replicating this model because it works quite well and 'many' people are familiar with it.


mastricker:
How is this PR related to #256 ?
Is seems to be a subset?

I am a little confused :)

felixmusil:
@max-veit required #256 to be split into two. Here is the 1st part.

mastricker:
> @max-veit required #256 to be split into two. Here is the 1st part.
Ok, I remember the discussion, of course.

Are the already requested changes on #256 incorporated here or where is it split?

(at some point you have to show me how you split it...)

bindings/bind_py_structure_manager.cc
```diff
@@ -642,7 +646,7 @@ namespace rascal {
         "__getitem__",
         [](ManagerCollection_t & v, int index) { return v[index]; },
         py::keep_alive<0, 1>());
-
+    manager_collection.def("__len__", &ManagerCollection_t::size);
```
max-veit: hmm, I'm confused as to what this "len" actually corresponds to -- is it the number of centres or the number of structures in the collection?

bindings/rascal/models/kernels.py
```diff
@@ -24,47 +32,94 @@ class Kernel(object):
         is between atomic environments) or 'Structure' (the kernel is summed over
         atoms in a structure), which is the default
 
+    kernel_type : string
+        Type of kernel method, either 'Full' (computing exact covariance matrix)
+        or 'Sparse' (computing GAP like kernel for sparse kernel methods like
+        Subset of Regressors)
```
max-veit: cite Quiñonero-Candela and Rasmussen if you're going to invoke Gaussian-process-specific jargon

bindings/rascal/models/krr.py
```diff
@@ -0,0 +1,195 @@
+import numpy as np
+
+
+class KRR(object):
+    """Kernel Ridge Regression model. Only compatible fully with sparse GPR
```
max-veit:
```suggestion
    """Kernel Ridge Regression model. Only supports sparse GPR
```

bindings/rascal/models/krr.py
```diff
@@ -0,0 +1,195 @@
...
+        np.array
+            predictions
+        """
+        KNM, Y0 = self._preprocess_input(managers, compute_gradients)
+        if compute_gradients is False:
+            return Y0 + np.dot(KNM, self.weights).reshape((-1))
+        else:
+            return np.dot(KNM, self.weights).reshape((-1, 3))
+
+    def get_weigths(self):
```
max-veit:
```suggestion
    def get_weights(self):
```

bindings/rascal/models/pseudo_points.py
```diff
@@ -0,0 +1,58 @@
...
+    def __init__(self, representation):
+        self.representation = representation
+        if 'SphericalInvariants' in str(representation):
+            self._pseudo_points = _pseudo_points['PseudoPointsBlockSparse_SphericalInvariants'](
+            )
+        else:
+            raise ValueError(
+                'No pseudo point is appropiate for ' + str(representation))
+
+    def extend(self, atoms_list, selected_indices):
```
max-veit: It's still ambiguous how `selected_indices` is interpreted in the context of a ManagerCollection -- does it index all centres in the collection, regardless of which structure it belongs to? How would one obtain such a list of indices?

bindings/rascal/utils/cur.py
```diff
@@ -0,0 +1,179 @@
...
+
+    is_deterministic: bool
+        flag to switch between selction criteria
+
+    seed: int
+        if is_deterministic==False, seed for the random selection
+
+    """
+
+    def __init__(self, representation, Nselect, act_on='sample per specie', is_deterministic=True, seed=10):
```
max-veit: `species` is the correct singular

bindings/rascal/utils/cur.py
```diff
@@ -0,0 +1,179 @@
...
+            Selected samples
+
+        Raises
+        ------
+        ValueError
+            [description]
+        NotImplementedError
+            [description]
+        """
+        if self.act_on in ['sample per specie']:
```
max-veit: `species`

src/rascal/models/kernels.hh
```diff
@@ -346,7 +357,8 @@ namespace rascal {
         return this->compute_helper<Property_t, TargetType::Atom>(
             representation_name, managers_a, managers_b);
       default:
-        throw std::logic_error("The combination of parameter is not handled.");
+        throw std::logic_error(
```
max-veit: Seriously, is this condition ever reached? Even if it is, it's not a useful error message; I would just leave out the default case entirely.

bindings/bind_py_structure_manager.cc
felixmusil: added documentation

bindings/rascal/models/pseudo_points.py
felixmusil: added some doc. Is it better now ?

src/rascal/models/kernels.hh
felixmusil: the compiler does not allow to not have a default in this case so I improved the message

max-veit: You shouldn't have bothered with the message; in fact, I think it's doing more harm than good by making a user think it's something _they_ did with some weird parameter combination, rather than being caused by a programmer error. If C++ allows throwing exceptions without a message, I would have just done that. Or at least write a message that makes it clear the programmer, not the user, is at fault.

src/rascal/models/kernels.hh
```diff
@@ -376,8 +376,10 @@ namespace rascal {
         return kernel->template compute<Property_t, Type>(
             managers_a, managers_b, representation_name);
       } else {
-        throw std::logic_error(
-            "The desired combination of parameters can not be handled.");
+        throw std::logic_error("Given kernel_type " +
+                            this->parameters["kernel_type"].get<std::string>() +
+                              " is not known."
+                              " It is 'Cosine'");
```
max-veit: I would have said `It must be one of: 'Cosine'`. But it's merged already, so keep that in mind for next time...

bindings/rascal/models/pseudo_points.py
max-veit: Yes, much better, thank you.

Luthaf:
@felixmusil, you forgot to delete your branch 😃

PR #263, opened 2020-02-22
Feat/update doc

felixmusil:
I realized the how to add a representation has been broken by some of our recent changes and also it needed some additional clarifications (@max-veit). 

I would keep this branch alive and uptodate with master so that we can use it to work on the documentation when needed. Sounds good ?

### DIFF
--- a/bindings/bind_py_structure_manager.cc
+++ b/bindings/bind_py_structure_manager.cc
@@ -880,32 +880,35 @@ namespace rascal {
     py::class_<ClusterRefBase>(m_internal, "ClusterRefBase");
     bind_cluster_refs(m_internal);
 
+    bind_atomic_structure(m_nl);
+
     py::module m_strc_mng = m_nl.def_submodule("StructureManager");
     m_strc_mng.doc() = "Structure Manager Classes";
     py::module m_adp = m_nl.def_submodule("Adaptor");
     m_adp.doc() = "Adaptor Classes";
 
-    using Manager_t = StructureManagerCenters;
-    bind_structure_manager<Manager_t>(m_strc_mng, m_internal);
-
-    // bind the factory function
-    bind_make_structure_manager<Manager_t>(m_nl);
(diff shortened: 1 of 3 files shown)


bindings/bind_py_structure_manager.cc
```diff
@@ -880,32 +880,35 @@ namespace rascal {
...
-    bind_make_structure_manager<Manager_t>(m_nl);
+    // list of already binded manager stacks to avoid double binding
     std::set<std::string> name_list{};
-    BindAdaptorStack<Manager_t, AdaptorNeighbourList, AdaptorStrict>
-        adaptor_stack_1{m_nl, m_adp, m_internal, name_list};
-
-    BindAdaptorStack<Manager_t, AdaptorNeighbourList, AdaptorCenterContribution,
+    /*-------------------- struc-man-bind-start --------------------*/
+    // bind the root structure manager
+    bind_structure_manager<StructureManagerCenters>(m_strc_mng, m_internal);
```
Luthaf: Why is the code updated too? Is this required for the doc changes ?

Luthaf:
> I would keep this branch alive and uptodate with master so that we can use it to work on the documentation when needed. Sounds good ?

I am not sure what you mean by that. Could you expand on it?

bindings/bind_py_structure_manager.cc
felixmusil: I reshuffled a bit the code so that it becomes more meaningful when displayed in the doc.

Luthaf: :+1:

felixmusil:
> I am not sure what you mean by that. Could you expand on it?

I would not delete this branch and keep it around for when one feels like updating the doc.

Luthaf:
What is the advantage of keeping the branch around versus creating a new one from master when one wants to change the docs? I can see at least one inconvenient with keeping a long-lived branch with a multiplication of merge commits.

PR #264, opened 2020-03-22
Bugfix/gaussian sigma

felixmusil:
Improve user friendliness of the the Spherical* family by setting a more sensible default to gaussian_sigma_constant on the python side and throwing an error if it is too small on the c++ side.


felixmusil:
> Seems sensible. Do we want to had a test so that the error do not come back later?

I would rather not since it would have to be generalized to other hypers and I would wait for the code to be more mature to start implementing important but not essential features like this. Also I would like to study properly the range of parameter we can safely use with GTO and DVR so that the test would become more meaningful.

max-veit:
If it's a user-friendliness improvement, why is this labelled a "bugfix"? Where's the bug?

felixmusil:
The default was 0 which leads to undefined behaviour, likely Nan everywhere.

PR #266, opened 2020-03-24
Feat/pip install

felixmusil:
Setup the infrastructure for pip installation of rascal. To avoid writing boilerplate code I use the scikit-build package which simplifies a lot driving cmake compilation. It is not strictly speaking needed but still quite useful to avoid superfluous work.
The demo package is available at `pip install -i https://test.pypi.org/simple/ rascal==0.3.8 ` (beware you need to install `pip install ase==3.19.0 scikit-build` beforehand because the demo mode of pip does not install missing dependencies). It will download, compile the code and install it in your python installation (I tested on linux and macos but only with conda virtual envs so please tell me if it works with different python installations). The only requirement that is not covered here is gcc/clang and cmake.

It works the following:
+ a minimal setup.py file takes care of the installation
+ uses  MANIFEST.in to copy the relevant files (now pip installation does not allow for anything but the bare minimum so no tests, doc, examples ...). The main point of this selection is to limit the package size.

I had to make a few modifications to the CMAKE files. Basically we need a package that can be copied over freely so when compiling for pip I set up the `RPATH` to be relative making sure that `_rascal.so` will find `librascal.so` within the python folder architecture. This also means that the installation procedure needs to already follow the python folder architecture, e.g. `PYTHON_ROOT/lib/python3.6/site-packages/`, hence the modification of  `bindings/CMakeLists.txt`.
This leaves one outstanding question: do we want to have relative RPATH by default too ?

We should be able to provide binaries too for several architectures / python versions but I am not quite clear on how to cross compile and it is not essential for the pip deployment to work.


Luthaf:
Very nice!

> I had to make a few modifications to the CMAKE files. Basically we need a package that can be copied over freely so when compiling for pip I set up the RPATH to be relative making sure that _rascal.so will find librascal.so within the python folder architecture. This also means that the installation procedure needs to already follow the python folder architecture, e.g. PYTHON_ROOT/lib/python3.6/site-packages/, hence the modification of bindings/CMakeLists.txt.

Another solution is to build librascal as a static lib, and link that inside `_rascal.so` instead of using a separate shared library. This would get rid of the RPATH issue completely.

> We should be able to provide binaries too for several architectures / python versions but I am not quite clear on how to cross compile and it is essential for the pip deployment to work.

I use docker on macOS and the scripts [here](https://github.com/chemfiles/chemfiles.py/tree/master/scripts) (start with build-all-wheels.sh) to build wheels (i.e. python binary distributions) for chemfiles, it could be a starting point to build rascal this way too. Bear in mind that since rascal links to libpython, we would need to build separate wheels for all python version.

Luthaf:
Currently, binding code can be built either with cmake -DBUILD_BINDINGS=ON, or with `python setup.py install`, right? So either `python setup.py install => scikit-build => cmake(rascal) -DBUILD_BINDINGS=ON`, or `cmake(rascal) -DBUILD_BINDINGS=ON; make install` 

I would suggest to change it so that the only way to build bindings is though `python setup.py install`, to simplify documentation and help to future users. This could work by creating a separate cmake project for the bindings, using the code in `bindings` and doing a `add_subdirectory(..)` there to get access to the librascal target. Then the build chain for the Python bindings would be `python setup.py install => scikit-build => cmake (rascal-python) => cmake(rascal)`.

Let me know if that's unclear.

felixmusil:
> Another solution is to build librascal as a static lib, and link that inside _rascal.so instead of using a separate shared library. This would get rid of the RPATH issue completely.

I don't know what difference it would make in the long run. The RPATH thing is under control at the moment so I would bear with it for a while until we discuss properly this kind of change.

> I use docker on macOS and the scripts here (start with build-all-wheels.sh) to build wheels (i.e. python binary distributions) for chemfiles, it could be a starting point to build rascal this way too. Bear in mind that since rascal links to libpython, we would need to build separate wheels for all python version.

Thanks for the reference it will come in handy later. Nevertheless, I think it is even worse than that because we need the standard library so we would also need to compile for gcc and clang at least (and hopefully not for several version of it). All in all I feel this is too much of a hassle considering that (1) a conda-forge distribution solves this 'reproducible and transferable builds' issue entirely by providing all the tool chain that we would need, making sure we don't rely on the user's installation for any dependency (2) the compilation time is quite short. 

> Currently, binding code can be built either with cmake -DBUILD_BINDINGS=ON, or with python setup.py install, right? So either python setup.py install => scikit-build => cmake(rascal) -DBUILD_BINDINGS=ON, or cmake(rascal) -DBUILD_BINDINGS=ON; make install

That is true and I would consider the `python setup.py install` route to not be documented for users. If one wants to play he can but we would keep the `cmake(rascal) ; make install` route the preferred one. Note that `python setup.py install` only sets up a `.egg` in site packages (so still can't `import rascal`) so it would be more like `pip install git+https://github.com/cosmo-epfl/librascal.git` or something like that. 
I feel it is easier (at the moment) to keep the python bindings on the same footing as other compilation modes. When we leave the development phase we might want to reconsider though.

Luthaf:
> Nevertheless, I think it is even worse than that because we need the standard library so we would also need to compile for gcc and clang at least (and hopefully not for several version of it). 

No, the point of the docker images used in these scripts is to deal with this. On Linux, this means building for the manylinux1 image (i.e. CentOS 6) which have old enough glibc and libstdc++ to be compatible with all modern distro. On macOS, this means setting MACOSX_DEPLOYMENT_TARGET to something old like 10.9 (from 2013). On Windows, this means cross-compiling with mingw64, and statically linking libstdc++. 

So no need to deal with multiple version of lib/compilers/different compilers. The resulting binary might not be compatible for linking with C++ software, but it will be compatible with the official Python builds/system Python builds on Linux.

> All in all I feel this is too much of a hassle considering that (1) a conda-forge distribution solves this 'reproducible and transferable builds' issue entirely by providing all the tool chain that we would need, making sure we don't rely on the user's installation for any dependency (2) the compilation time is quite short.

I agree with (1), but this is also a question of conda vs pip. I think the best is to have the code available on both package managers, meaning we need to build wheels for conda. Not everyone is using conda for everything, for different reasons.

Concerning (2), the compilation time is not the issue, the problem is mostly with installing dependencies (compiler + cmake), which can be hard to do on some systems (Windows, specially since librascal requires GCC for compilation; and macOS where most users do not have a compiler installed).

> Note that python setup.py install only sets up a .egg in site packages (so still can't import rascal)

Not sure I understand what you mean here. If I clone librascal and only run `python setup.py install` I don't get a working install? If so we should modify setup.py/cmake for this to work.

> That is true and I would consider the python setup.py install route to not be documented for users.

I think we should do the reverse, since trying to deal with python install directoryis hard. For example, the current code will by default take the first version of Python from the user's PATH, and install there. This causes issue with "user" site-package (pip/python setup.py install --user), with [workaround](https://github.com/cosmo-epfl/librascal/blob/45d4e53dbff34b9362c3bddd321b8ff8e3acfd8a/CMakeLists.txt#L145-L162) like the USER cmake variable we have. This will also cause issues if the user want to install with a different version of Python. 

All of that can be fixed by adding more workaround and cmake settings, or as I am proposing by not even trying to deal with this ourself in cmake and instead use the standard Python tools.

> When we leave the development phase we might want to reconsider though.

I generally disagree with this point of view. If we always wait for "end of dev phase" to provide user-focussed improvements, we will never do it since (1) development of any software project is never finished and (2) there will be more and more work to do, making the task too big to be done. 

This is the same as with documentation, I think it should be done in an incremental fashion, all the time.

felixmusil:
I found the problem that lead me to modify the cmake files: we were not installing rascal in bindings which changes the folder architecture between the build and install folder. After making the proper changes we don't need these special cases. I also took the liberty to move to relative RPATH always, is that ok ?. 


> So no need to deal with multiple version of lib/compilers/different compilers. The resulting binary might not be compatible for linking with C++ software, but it will be compatible with the official Python builds/system Python builds on Linux.

I might be missing something ? The builds that are provided are linked explicitly to the stl library, e.g. `librascal.so` is linked to `libstdc++.so.6` so if `libstdc++.so.6` is not around it will crash, hence the need for a compiler providing  `libstdc++.so.6` on the user side and the dependency requirement is not solved. 

> Windows 

we don't care about Windows.

> Not sure I understand what you mean here. If I clone librascal and only run python setup.py install I don't get a working install? If so we should modify setup.py/cmake for this to work.

I am a bit at a loss considering the not so clear standards and instructions for pip build/deployment... My two cents after a few hours of reading: using `python setup.py install` triggers the `egg` packaging of setuptools which is not recommended to use [anymore](https://packaging.python.org/discussions/wheel-vs-egg/) it seems. So we don't want to do that in any cases. One should use `pip install .`, which uses `wheel`, in the librascal directory. It works as it should, is there still a problem ?

> I think we should do the reverse, since trying to deal with python install directoryis hard. For example, the current code will by default take the first version of Python from the user's PATH, and install there. This causes issue with "user" site-package (pip/python setup.py install --user), with workaround like the USER cmake variable we have.

How about the 'new' procedure for python installation is `pip install .` ? It builds the library and installs it in the proper place with a minimal installation. I would be able to remove the USER keyword and some python path related stuff from the cmake files too. Nevertheless for development purposes we keep the same functionality, just we would not be able to install the bindings coming from this directory. 

> This will also cause issues if the user want to install with a different version of Python.

what do you mean by installing with a different python ?

Luthaf:
> I also took the liberty to move to relative RPATH always, is that ok ?.

What are the implications of using a RPATH, and in particular of a relative RPATH? I am not totally clear on the subject.

> hence the need for a compiler providing libstdc++.so.6 on the user side and the dependency requirement is not solved.

`libstdc++.so.6` will always be available on Linux, as soon as the user have any software written in C++ installed. The .so file is a separated dependency that will be provided even without a compiler. `libstdc++.so.6` can have different versions, as indicated by the `GLIBCXX_x.y.z` symbol. libstdc++.so.6 have been [backward compatible since GCC 3.4.0](https://gcc.gnu.org/onlinedocs/libstdc++/manual/abi.html), hence the compilation of the code using an old version of `libstdc++.so.6` to be compatible with all newer linux distro.

On macOs, the same is true of `libSystemB.dylib`.

> I am a bit at a loss considering the not so clear standards and instructions for pip build/deployment...

Yeah, Python packaging is a mess =/ 

I am happy with `pip install`, either as `pip install .` or `pip install git+https:// ...` =).

Re development purposes, `python setup.py develop` is a pretty nice solution too. It basically add a symlink to the current folder in site-package (an entry in a .pth file in practice), meaning that you can change the files and immediately see your changes in other projects.

Concerning `python setup.py install` I'll have to do more searching, but I think you can have it  build wheels too.

felixmusil:
> What are the implications of using a RPATH, and in particular of a relative RPATH? I am not totally clear on the subject.

I don't think there is any. On UNIX systems the RPATH are the first places to be looked at for libraries and then the LD_LIB... so if it is not useful then the linker will look somewhere else. I our case we use it just for one thing, _rascal.so should be able to find librascal.so in the python installation. The limitation of a relative RPATH is that if the python distro changes its folder architecture then we will have to adapt this bit of the cmake files which is quite unlikely to happen so it should not be an issue.

> libstdc++.so.6 will always be available on Linux, as soon as the user have any software written in C++ installed. The .so file is a separated dependency that will be provided even without a compiler. libstdc++.so.6 can have different versions, as indicated by the GLIBCXX_x.y.z symbol. libstdc++.so.6 have been backward compatible since GCC 3.4.0, hence the compilation of the code using an old version of libstdc++.so.6 to be compatible with all newer linux distro.

I see.

requirements_pip.txt
```diff
@@ -0,0 +1,2 @@
+# python library that contain numpy and scipy already
+ase==3.19.0
```
Luthaf: We should be more explicit on the dependencies. If we import numpy directly ourself, it should be in the requirements.txt file. Same for scipy, and any other dependency.

setup.py
```diff
@@ -0,0 +1,35 @@
...
+from os.path import join
+from distutils.sysconfig import get_python_lib
+from setuptools import find_packages
+
+try:
+    from skbuild import setup
+    from skbuild.constants import CMAKE_INSTALL_DIR
+
+except ImportError:
+    print('scikit-build is required to build from source.', file=sys.stderr)
```
Luthaf: why catch the ImportError? We don't do this for other dependencies

src/pyproject.toml
```diff
@@ -0,0 +1,2 @@
+[build-system]
```
Luthaf: this file seems to be duplicated

setup.py
felixmusil: yeah this dependency should be handled by pip. I forgot once or twice to install it so I put a warning..

Luthaf:
After doing `python setup.py install` in a docker container, I get the following error

```
>>> from rascal import representations
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.6/site-packages/rascal-0.3.9-py3.6-linux-x86_64.egg/rascal/representations/__init__.py", line 1, in <module>
    from .coulomb_matrix import SortedCoulombMatrix
  File "/usr/local/lib/python3.6/site-packages/rascal-0.3.9-py3.6-linux-x86_64.egg/rascal/representations/coulomb_matrix.py", line 4, in <module>
    from ..neighbourlist import AtomsList
  File "/usr/local/lib/python3.6/site-packages/rascal-0.3.9-py3.6-linux-x86_64.egg/rascal/neighbourlist/__init__.py", line 1, in <module>
    from .structure_manager import AtomsList, get_neighbourlist, convert_to_structure_list
  File "/usr/local/lib/python3.6/site-packages/rascal-0.3.9-py3.6-linux-x86_64.egg/rascal/neighbourlist/structure_manager.py", line 7, in <module>
    from ..lib import neighbour_list
  File "/usr/local/lib/python3.6/site-packages/rascal-0.3.9-py3.6-linux-x86_64.egg/rascal/lib/__init__.py", line 1, in <module>
    from ._rascal import (neighbour_list, models,
ImportError: librascal.so: cannot open shared object file: No such file or directory
```

`pip install .` does work, so we should either detect that the user is running `python setup.py install` directly and error telling them to use pip; or make it so that the .so file is put inside the .egg (maybe using the compiled dependency concept of setuptools).

----

rascal is seen by pip at version 0.3.9 for some reason (`Successfully installed rascal-0.3.9` messages), why is it so? There is no `__version__` attribute on the module. 

This should be set to an explicit version which we can control so that update behave properly.

Luthaf:
> After doing python setup.py install in a docker container, I get the following error

This is still not working, we should either make it work or send an error when a user tries to `python setup.py install`.

felixmusil:
What do you mean ? It works on my machine so please elaborate.

Luthaf:
If I run `python setup.py install` and then `python -c "from rascal import representations"`, I get no error during installation, and the following during import:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.6/site-packages/rascal-0.0.0-py3.6-linux-x86_64.egg/rascal/representations/__init__.py", line 1, in <module>
    from .coulomb_matrix import SortedCoulombMatrix
  File "/usr/local/lib/python3.6/site-packages/rascal-0.0.0-py3.6-linux-x86_64.egg/rascal/representations/coulomb_matrix.py", line 4, in <module>
    from ..neighbourlist import AtomsList
  File "/usr/local/lib/python3.6/site-packages/rascal-0.0.0-py3.6-linux-x86_64.egg/rascal/neighbourlist/__init__.py", line 1, in <module>
    from .structure_manager import AtomsList, get_neighbourlist, convert_to_structure_list
  File "/usr/local/lib/python3.6/site-packages/rascal-0.0.0-py3.6-linux-x86_64.egg/rascal/neighbourlist/structure_manager.py", line 7, in <module>
    from ..lib import neighbour_list
  File "/usr/local/lib/python3.6/site-packages/rascal-0.0.0-py3.6-linux-x86_64.egg/rascal/lib/__init__.py", line 1, in <module>
    from ._rascal import (neighbour_list, models,
ImportError: librascal.so: cannot open shared object file: No such file or directory
```

felixmusil:
When you run the install command there is a full traceback of the actions taken but the installation procedure. Please send this traceback since the import error is useless for understanding the problem.

Luthaf:
There is no installation error, here is the install log

<details>

```
# python3 setup.py install

--------------------------------------------------------------------------------
-- Trying "Ninja" generator
--------------------------------
---------------------------
----------------------
-----------------
------------
-------
--
Not searching for unused variables given on the command line.
CMake Error: CMake was unable to find a build program corresponding to "Ninja".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.
-- Configuring incomplete, errors occurred!
See also "/librascal/_cmake_test_compile/build/CMakeFiles/CMakeOutput.log".
--
-------
------------
-----------------
----------------------
---------------------------
--------------------------------
-- Trying "Ninja" generator - failure
--------------------------------------------------------------------------------



--------------------------------------------------------------------------------
-- Trying "Unix Makefiles" generator
--------------------------------
---------------------------
----------------------
-----------------
------------
-------
--
Not searching for unused variables given on the command line.
-- The C compiler identification is GNU 7.3.1
-- Check for working C compiler: /opt/rh/devtoolset-7/root/usr/bin/cc
-- Check for working C compiler: /opt/rh/devtoolset-7/root/usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- The CXX compiler identification is GNU 7.3.1
-- Check for working CXX compiler: /opt/rh/devtoolset-7/root/usr/bin/c++
-- Check for working CXX compiler: /opt/rh/devtoolset-7/root/usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Configuring done
-- Generating done
-- Build files have been written to: /librascal/_cmake_test_compile/build
--
-------
------------
-----------------
----------------------
---------------------------
--------------------------------
-- Trying "Unix Makefiles" generator - success
--------------------------------------------------------------------------------

-- Build type is: Release
-- The CXX compiler identification is GNU 7.3.1
-- Check for working CXX compiler: /opt/rh/devtoolset-7/root/usr/bin/c++
-- Check for working CXX compiler: /opt/rh/devtoolset-7/root/usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Downloading wigxjpf
-- Downloading Eigen3
-- Eigen3 3.3.4
-- Found PythonInterp: /usr/bin/python3 (found suitable version "3.6.8", minimum required is "3")
-- Found PythonLibs: libpython3.6m.a
-- Downloading pybind11
-- pybind11 v2.3.0
-- Installation ROOT: /librascal/_skbuild/linux-x86_64-3.6/cmake-install
-- The project is built using scikit-build
-- The optional cpplint parser has not been found. For more information see https://pypi.python.org/pypi/cpplint
-- The optional clang-format auto formatter has not been found. For more information see https://clang.llvm.org/docs/ClangFormat.html
-- The optional autopep8 parser has not been found. For more information see https://github.com/hhatto/autopep8
-- Configuring done
-- Generating done
-- Build files have been written to: /librascal/_skbuild/linux-x86_64-3.6/cmake-build
Scanning dependencies of target rascal
[  5%] Building CXX object src/CMakeFiles/rascal.dir/rascal/utils/json_io.cc.o
[ 10%] Building CXX object src/CMakeFiles/rascal.dir/rascal/utils/units.cc.o
[ 15%] Building CXX object src/CMakeFiles/rascal.dir/rascal/utils/utils.cc.o
[ 20%] Building CXX object src/CMakeFiles/rascal.dir/rascal/utils/sparsify_fps.cc.o
[ 25%] Building CXX object src/CMakeFiles/rascal.dir/rascal/math/bessel.cc.o
[ 30%] Building CXX object src/CMakeFiles/rascal.dir/rascal/math/hyp1f1.cc.o
[ 35%] Building CXX object src/CMakeFiles/rascal.dir/rascal/math/interpolator.cc.o
[ 40%] Building CXX object src/CMakeFiles/rascal.dir/rascal/math/gauss_legendre.cc.o
[ 45%] Building CXX object src/CMakeFiles/rascal.dir/rascal/math/spherical_harmonics.cc.o
[ 50%] Building CXX object src/CMakeFiles/rascal.dir/rascal/structure_managers/structure_manager_lammps.cc.o
[ 55%] Building CXX object src/CMakeFiles/rascal.dir/rascal/structure_managers/structure_manager_centers.cc.o
[ 60%] Building CXX object src/CMakeFiles/rascal.dir/rascal/representations/calculator_base.cc.o
[ 65%] Linking CXX shared library librascal.so
[ 65%] Built target rascal
[ 65%] Built target rascal
[ 70%] Building CXX object bindings/CMakeFiles/_rascal.dir/bind_py_module.cc.o
[ 75%] Building CXX object bindings/CMakeFiles/_rascal.dir/bind_py_structure_manager.cc.o
[ 80%] Building CXX object bindings/CMakeFiles/_rascal.dir/bind_py_utils.cc.o
[ 85%] Building CXX object bindings/CMakeFiles/_rascal.dir/bind_py_representation_calculator.cc.o
[ 90%] Building CXX object bindings/CMakeFiles/_rascal.dir/bind_py_models.cc.o
[ 95%] Linking CXX shared module _rascal.cpython-36m-x86_64-linux-gnu.so
[ 95%] Built target _rascal
Scanning dependencies of target copy_python_bindings
[100%] Copying python bindings and libraries
[100%] Built target copy_python_bindings
Install the project...
-- Install configuration: "Release"
-- Installing: /librascal/_skbuild/linux-x86_64-3.6/cmake-install/lib/librascal.so
-- Set runtime path of "/librascal/_skbuild/linux-x86_64-3.6/cmake-install/lib/librascal.so" to "$ORIGIN/../../../../"
-- Installing: /librascal/_skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal
-- Installing: /librascal/_skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/models
-- Installing: /librascal/_skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/models/__init__.py
-- Installing: /librascal/_skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/models/kernels.py
-- Installing: /librascal/_skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/neighbourlist
-- Installing: /librascal/_skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/neighbourlist/base.py
-- Installing: /librascal/_skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/neighbourlist/__init__.py
-- Installing: /librascal/_skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/neighbourlist/structure_manager.py
-- Installing: /librascal/_skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/utils
-- Installing: /librascal/_skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/utils/fps.py
-- Installing: /librascal/_skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/utils/__init__.py
-- Installing: /librascal/_skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/utils/pool_worker.py
-- Installing: /librascal/_skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/__init__.py
-- Installing: /librascal/_skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/lib
-- Installing: /librascal/_skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/lib/__init__.py
-- Installing: /librascal/_skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/representations
-- Installing: /librascal/_skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/representations/coulomb_matrix.py
-- Installing: /librascal/_skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/representations/base.py
-- Installing: /librascal/_skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/representations/spherical_invariants.py
-- Installing: /librascal/_skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/representations/__init__.py
-- Installing: /librascal/_skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/representations/spherical_covariants.py
-- Installing: /librascal/_skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/representations/spherical_expansion.py
-- Installing: /librascal/_skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/lib/_rascal.cpython-36m-x86_64-linux-gnu.so
-- Set runtime path of "/librascal/_skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/lib/_rascal.cpython-36m-x86_64-linux-gnu.so" to "$ORIGIN/../../../../"

copying bindings/rascal/__init__.py -> _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/__init__.py
copying bindings/rascal/models/__init__.py -> _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/models/__init__.py
copying bindings/rascal/models/kernels.py -> _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/models/kernels.py
copying bindings/rascal/neighbourlist/base.py -> _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/neighbourlist/base.py
copying bindings/rascal/neighbourlist/__init__.py -> _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/neighbourlist/__init__.py
copying bindings/rascal/neighbourlist/structure_manager.py -> _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/neighbourlist/structure_manager.py
copying bindings/rascal/utils/fps.py -> _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/utils/fps.py
copying bindings/rascal/utils/__init__.py -> _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/utils/__init__.py
copying bindings/rascal/utils/pool_worker.py -> _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/utils/pool_worker.py
copying bindings/rascal/lib/__init__.py -> _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/lib/__init__.py
copying bindings/rascal/representations/coulomb_matrix.py -> _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/representations/coulomb_matrix.py
copying bindings/rascal/representations/base.py -> _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/representations/base.py
copying bindings/rascal/representations/spherical_invariants.py -> _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/representations/spherical_invariants.py
copying bindings/rascal/representations/__init__.py -> _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/representations/__init__.py
copying bindings/rascal/representations/spherical_covariants.py -> _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/representations/spherical_covariants.py
copying bindings/rascal/representations/spherical_expansion.py -> _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/representations/spherical_expansion.py

running install
running bdist_egg
running egg_info
creating rascal.egg-info
writing rascal.egg-info/PKG-INFO
writing dependency_links to rascal.egg-info/dependency_links.txt
writing requirements to rascal.egg-info/requires.txt
writing top-level names to rascal.egg-info/top_level.txt
writing manifest file 'rascal.egg-info/SOURCES.txt'
reading manifest file 'rascal.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
warning: no previously-included files matching '__pycache__' found under directory '*'
warning: no previously-included files matching '*.pyc' found under directory '*'
writing manifest file 'rascal.egg-info/SOURCES.txt'
installing library code to _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg
running install_lib
running build_py
creating _skbuild/linux-x86_64-3.6/setuptools/lib
creating _skbuild/linux-x86_64-3.6/setuptools/lib/rascal
copying _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/__init__.py -> _skbuild/linux-x86_64-3.6/setuptools/lib/rascal
creating _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/models
copying _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/models/__init__.py -> _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/models
copying _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/models/kernels.py -> _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/models
creating _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/neighbourlist
copying _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/neighbourlist/base.py -> _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/neighbourlist
copying _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/neighbourlist/__init__.py -> _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/neighbourlist
copying _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/neighbourlist/structure_manager.py -> _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/neighbourlist
creating _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/utils
copying _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/utils/fps.py -> _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/utils
copying _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/utils/__init__.py -> _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/utils
copying _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/utils/pool_worker.py -> _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/utils
creating _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/lib
copying _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/lib/__init__.py -> _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/lib
creating _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/representations
copying _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/representations/coulomb_matrix.py -> _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/representations
copying _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/representations/base.py -> _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/representations
copying _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/representations/spherical_invariants.py -> _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/representations
copying _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/representations/__init__.py -> _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/representations
copying _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/representations/spherical_covariants.py -> _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/representations
copying _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/representations/spherical_expansion.py -> _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/representations
copying _skbuild/linux-x86_64-3.6/cmake-install/bindings/rascal/lib/_rascal.cpython-36m-x86_64-linux-gnu.so -> _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/lib
copied 16 files
running build_ext
creating _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64
creating _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg
creating _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/rascal
creating _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/rascal/models
copying _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/models/__init__.py -> _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/rascal/models
copying _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/models/kernels.py -> _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/rascal/models
creating _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/rascal/neighbourlist
copying _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/neighbourlist/base.py -> _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/rascal/neighbourlist
copying _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/neighbourlist/__init__.py -> _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/rascal/neighbourlist
copying _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/neighbourlist/structure_manager.py -> _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/rascal/neighbourlist
creating _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/rascal/utils
copying _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/utils/fps.py -> _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/rascal/utils
copying _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/utils/__init__.py -> _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/rascal/utils
copying _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/utils/pool_worker.py -> _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/rascal/utils
copying _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/__init__.py -> _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/rascal
creating _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/rascal/lib
copying _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/lib/_rascal.cpython-36m-x86_64-linux-gnu.so -> _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/rascal/lib
copying _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/lib/__init__.py -> _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/rascal/lib
creating _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/rascal/representations
copying _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/representations/coulomb_matrix.py -> _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/rascal/representations
copying _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/representations/base.py -> _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/rascal/representations
copying _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/representations/spherical_invariants.py -> _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/rascal/representations
copying _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/representations/__init__.py -> _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/rascal/representations
copying _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/representations/spherical_covariants.py -> _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/rascal/representations
copying _skbuild/linux-x86_64-3.6/setuptools/lib/rascal/representations/spherical_expansion.py -> _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/rascal/representations
copied 17 files
installing package data to _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg
running install_data
creating _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/lib
copying _skbuild/linux-x86_64-3.6/cmake-install/lib/librascal.so -> _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/lib
creating _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/EGG-INFO
copying rascal.egg-info/PKG-INFO -> _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/EGG-INFO
copying rascal.egg-info/SOURCES.txt -> _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/EGG-INFO
copying rascal.egg-info/dependency_links.txt -> _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/EGG-INFO
copying rascal.egg-info/not-zip-safe -> _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/EGG-INFO
copying rascal.egg-info/requires.txt -> _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/EGG-INFO
copying rascal.egg-info/top_level.txt -> _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/EGG-INFO
writing _skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg/EGG-INFO/native_libs.txt
creating dist
creating 'dist/rascal-0.0.0-py3.6-linux-x86_64.egg' and adding '_skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg' to it
removing '_skbuild/linux-x86_64-3.6/setuptools/bdist.linux-x86_64/egg' (and everything under it)
Processing rascal-0.0.0-py3.6-linux-x86_64.egg
creating /usr/local/lib/python3.6/site-packages/rascal-0.0.0-py3.6-linux-x86_64.egg
Extracting rascal-0.0.0-py3.6-linux-x86_64.egg to /usr/local/lib/python3.6/site-packages
Adding rascal 0.0.0 to easy-install.pth file

Installed /usr/local/lib/python3.6/site-packages/rascal-0.0.0-py3.6-linux-x86_64.egg
Processing dependencies for rascal==0.0.0
Searching for ase==3.19.0
Best match: ase 3.19.0
Adding ase 3.19.0 to easy-install.pth file
Installing ase script to /usr/local/bin
Installing ase-build script to /usr/local/bin
Installing ase-db script to /usr/local/bin
Installing ase-gui script to /usr/local/bin
Installing ase-info script to /usr/local/bin
Installing ase-run script to /usr/local/bin

Using /usr/local/lib/python3.6/site-packages
Searching for scipy==1.4.1
Best match: scipy 1.4.1
Adding scipy 1.4.1 to easy-install.pth file

Using /usr/local/lib64/python3.6/site-packages
Searching for numpy==1.18.3
Best match: numpy 1.18.3
Adding numpy 1.18.3 to easy-install.pth file
Installing f2py script to /usr/local/bin
Installing f2py3 script to /usr/local/bin
Installing f2py3.6 script to /usr/local/bin

Using /usr/local/lib64/python3.6/site-packages
Searching for matplotlib==3.2.1
Best match: matplotlib 3.2.1
Adding matplotlib 3.2.1 to easy-install.pth file

Using /usr/local/lib64/python3.6/site-packages
Searching for pyparsing==2.4.7
Best match: pyparsing 2.4.7
Adding pyparsing 2.4.7 to easy-install.pth file

Using /usr/local/lib/python3.6/site-packages
Searching for kiwisolver==1.2.0
Best match: kiwisolver 1.2.0
Adding kiwisolver 1.2.0 to easy-install.pth file

Using /usr/local/lib64/python3.6/site-packages
Searching for cycler==0.10.0
Best match: cycler 0.10.0
Adding cycler 0.10.0 to easy-install.pth file

Using /usr/local/lib/python3.6/site-packages
Searching for python-dateutil==2.8.1
Best match: python-dateutil 2.8.1
Adding python-dateutil 2.8.1 to easy-install.pth file

Using /usr/local/lib/python3.6/site-packages
Searching for six==1.14.0
Best match: six 1.14.0
Adding six 1.14.0 to easy-install.pth file

Using /usr/local/lib/python3.6/site-packages
Finished processing dependencies for rascal==0.0.0
```

</details>

Bear in mind this is inside a container, while you may have `librascal.so` already available somewhere on your machine.

Luthaf:
Here is a docker file for reproduction

```dockerfile
FROM ubuntu:latest

RUN apt update
RUN apt install wget python3 python3-dev python3-pip git cmake g++

RUN git clone https://github.com/cosmo-epfl/librascal --branch feat/pip_install
RUN pip3 install -r librascal/requirements_pip.txt scikit-build

RUN cd librascal && python3 setup.py install
RUN python3 -c "from rascal import representations"
```

Put it somewhere, and then run `docker build`, or manually run the commands inside `ubuntu:latest`

felixmusil:
it works now. I had to take into account the fact that pip and setuptool don’t install in the same way.

for future reference here is the working docker scritp I used:
`
FROM ubuntu:latest
USER root
RUN  apt-get update && \
     apt-get install -y wget python3 python3-dev python3-pip git cmake g++

RUN pip3 install  scikit-build wheel setuptools ninja
RUN git clone https://github.com/cosmo-epfl/librascal --branch feat/pip_install
RUN pip3 install -r librascal/requirements_pip.txt
RUN cd librascal && python3 setup.py install
`

Luthaf:
> it works now. I had to take into account the fact that pip and setuptool don’t install in the same way.

I don't see any new commit, did you push?

felixmusil:
> I don't see any new commit, did you push?

Yes, it's in commit 405799d.

setup.py
```diff
@@ -0,0 +1,41 @@
...
+
+with open('./requirements_pip.txt', 'r') as fp:
+    requirements = list(filter(lambda x: '#' not in x, (line.strip() for line in fp)))
+
+setup(
+    name="rascal",
+    version=version,
+    description="""A versatile and scalable computation of representations of
+atomic structures for machine learning.""",
+    author='Félix Musil',
```
Luthaf, line 29: Should this be `librascal developers`, or just list everyone?

setup.py
```diff
@@ -0,0 +1,41 @@
...
+    license="LGPL-3",
```
Luthaf, line 30: I have never seen the license specified as an argument to setup, instead it is usually in `classifiers`, like `"License :: OSI Approved :: Python Software Foundation License"`. I could also not find documentation for this specific keyword.

CMakeLists.txt
```diff
@@ -142,47 +141,54 @@ if(BUILD_BINDINGS)
...
+# relative RPATH so that _rascal.so can find librascal.so when installed.
+# see link for more details on the topic
 # https://gitlab.kitware.com/cmake/community/wikis/doc/cmake/RPATH-handling
 set(CMAKE_SKIP_BUILD_RPATH  FALSE)
 set(CMAKE_BUILD_WITH_INSTALL_RPATH FALSE)
-set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib")
+set(CMAKE_BUILD_RPATH_USE_ORIGIN TRUE)
+if(NOT APPLE AND UNIX)
+  # this link path accommodates most installations
+  set(CMAKE_INSTALL_RPATH "\\$ORIGIN/../../../../")
```
Luthaf, line 182: I don't like this. It is assuming that python packages are always installed at `/usr/local/lib/pythonX.Y/site-packages/`, which is not true. For example, I sometimes use homebrew on linux to install software, in particular modern Python, which then install packages to a different location.

What is wrong with `set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib")`?

I think the proper solution here would be to remove the separation between librascal.so and _rascal.cpython-XXX.so, but as a middle ground how about always put `librascal.so` in `site-packages/rascal-egg/lib/librascal.so`, and use the second RPATH only?

setup.py
felixmusil: I just put it there as a placeholder for the moment.

setup.py
felixmusil: this is a proper keyword as you can see from [here](https://setuptools.readthedocs.io/en/latest/setuptools.html?highlight=setup#configuring-setup-using-setup-cfg-files). classifiers is another metadata keyword that serves another purpose.

Luthaf: Ok, good! Should it be using SPDX license identifier, i.e. `LGPL-3.0-or-later` here?

Luthaf: https://spdx.org/licenses/

CMakeLists.txt
felixmusil: > I don't like this. It is assuming that python packages are always installed at /usr/local/lib/pythonX.Y/site-packages/, which is not true. For example, I sometimes use homebrew on linux to install software, in particular modern Python, which then install packages to a different location.

I am not sure to follow. You mean to say that one might install python packages not in `somewhere/lib/pythonX.Y/site-packages/` or `~/.local/lib/pythonX.Y/site-packages/` ? Is this actually following the general gidelines of python's package managing ?

> What is wrong with set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib")?

if you ran the command `python setup.py install` you can see that the package is not installed from the build package but installed locally and then copied over twice. It seems that it is one of the PEPXXX requirements. So `CMAKE_INSTALL_PREFIX` is not proper.

> I think the proper solution here would be to remove the separation between librascal.so and _rascal.cpython-XXX.so, but as a middle ground how about always put librascal.so in site-packages/rascal-egg/lib/librascal.so, and use the second RPATH only?

having `librascal.so` in the lib folder of the python installation was decided upon a while ago as the **proper** way to do it. Note that it is how it is done with many packages that rely on dynamic c/c++ libraries.
Installing with pip (meaning `wheel`) or with `python setup.py install` (meaning `egg`) does not lead to the same behaviour. In on case you have this egg folder that contains lib while in the other librascal.so is in the main lib folder. I think that both should be handled separately as I could not find a way to block using `egg`. For sure one **should not** use `egg` for installation as this package is not registered within pip.

setup.py
felixmusil: we use `LGPL-3.0` so I would use `LGPL-3.0-only` rather than what you suggest. Or am I missing something ?

Luthaf: `LGPL-3.0-only` works, I did not check if we had the `or any later version` text in the license

CMakeLists.txt
Luthaf: > I am not sure to follow. You mean to say that one might install python packages not in somewhere/lib/pythonX.Y/site-packages/ or ~/.local/lib/pythonX.Y/site-packages/ ? Is this actually following the general gidelines of python's package managing ?

Sorry, what I mean is that assuming the existence of a file outside of `$PREFIX/lib/pythonX.Y/site-packages/rascal/` may backfire, and indeed the file only exist if the whole package was installed using pip.

I also somehow misunderstood what `/../../../../` meant in the RPATH, could you add a comment that this is to find a lib in `$PREFIX/lib` when the bindings are installed in `$PREFIX/lib/pythonX.Y/site-packages/rascal/`?

Thinking a bit more about it, it should indeed work in most settings, so I would be happy with a comment explaining it. I would still prefer not polluting `$PREFIX/lib` when installing a Python package though.

> having librascal.so in the lib folder of the python installation was decided upon a while ago as the proper way to do it. Note that it is how it is done with many packages that rely on dynamic c/c++ libraries.

Do you mean having `$PREFIX/lib/librascal.so` and `$PREFIX/lib/pythonX.Y/site-packages/rascal/lib/_rascal.cpython.X.Y.so`? If so I agree this is the right way to do it when one want to install both the C++ lib and the Python package. But here we are talking about user interested in installing the python lib only, and the C++ lib is only an implementation detail.

> For sure one should not use egg for installation as this package is not registered within pip.

I fail to see the link between using egg and registering with PyPI.

I use `python setup.py install`  and `python setup.py develop` often for Python packages, so I expect other would do the same, hence why I think we should make it work.

setup.py
Luthaf: Should actually be the `-or-later` version

```
# Rascal is free software; you can redistribute it and/or
# modify it under the terms of the GNU Lesser General Public License as
# published by the Free Software Foundation, either version 3, or (at
# your option) any later version.
```

CMakeLists.txt
felixmusil: > Do you mean having $PREFIX/lib/librascal.so and $PREFIX/lib/pythonX.Y/site-packages/rascal/lib/_rascal.cpython.X.Y.so? If so I agree this is the right way to do it when one want to install both the C++ lib and the Python package. But here we are talking about user interested in installing the python lib only, and the C++ lib is only an implementation detail.

the end goal here is to potentially have librascal.so to be used by other executable, e.g. lammps for instance, so that the when building a model with python and running it in c++ uses the same underlying library avoiding incompatibilities and such silly things. So ultimately it would not be only the python lib that we are interested in. 
Of course we can discuss this but that would be outside of this PR.

> I fail to see the link between using egg and registering with PyPI.

PyPI is different from pip since the later is a local command line utility for package managing while the former is an online python package repository that can be accessed with pip. 
What I mean is that a package installed using `python setup.py install` is not managed by any package manager while `pip install .` results in such a package. Hence on should prefer `pip install .` for simplicity of management. Also egg does not follow any PEP standard probably resulting in the difference in behaviour that we observe and the workarounds needed to make it work for our purpose.

> I use python setup.py install and python setup.py develop often for Python packages, so I expect other would do the same, hence why I think we should make it work.

and it does work now, it just needed some additional workaround.

PR #267, opened 2020-03-25
serialization

felixmusil:
This PR brings:

1. some c++ class serialization to the json type
2. a mechanism to serialize python class 

After training a model we need to save it whether it exists in python or c++. At the moment I only implemented a python model so it need to be saved/loaded through python. I choose python dictionaries as the serialization format as it can then be written to disk in most cross platform format like JSON, msgpack... so potentially a model saved from python could be read and ran from c++
To do that one needs to serialize python objects to dictionaries (done thought the BaseIO class and some modifications of the class that should be saved). 
And also the underlying c++ objects which is done in a 2 step procedure:  (a) convert the c++ class to the json class (see the end of the SparseKernel file for instance) in c++ (b) then convert the json c++ object to a python dictionary in the to_dict/from_dict binding functions.
Note that the main object that needs serialization is the PseudoPoints class as it holds a lot of the training information stored in a sparse format when dealing with SphericalInvariants features.

Note that most of the diff comes from the update of the json.hpp to a newer version.


felixmusil:
> why is serialization being done through Python dictionaries, instead of just using the JSON format that is accessible from both the C++ and Python sides?

Serializing to JSON means that to serialize to another format one would have to do it again. The rational here is that any python io library like json, msgpack ... knows how to dump a python dictionary so serializing to dictionaries allow for easy extensions.

> It seems like a bit of a weird workaround on the C++ side to have to convert to a Python dictionary before saving

This is not a workaround, the python and c++ are talking to each other using JSON messages. On the C++ side the equivalent of dictionaries is the json type which is passed to python using JSON messages. It could be some other type of messages though.

> not to mention making C++ serialization available only when the Python bindings are compiled.

C++ serialization is the convertion to the json type so it is available without the bindings.

> Second, why does the external json library need to be updated? And is there a better way of encapsulating such upstream changes, e.g. by keeping the external repo as a submodule? (Maybe not the whole thing, since we only need one file...).

There was a bug fixed in later releases that was a problem hence the update. We could use submodules, we diceded not to at some point choosing to have the header directly in the repo. We should discuss this next meeting I guess.

> I'm confused what the get_init_params() function is for

it returns a dictionarry containing all the parameters used in the __init__() function. Thought the name was pretty litteral there. 

> why the _set_data() function appears to do nothing in many of the classes that implement it

This function is meant to recover data that was not part of the initialization, like the features of the sparse_point class. Some of our classes don't produce such data hence some empty _set_data(). I don't see where the problem is but I am happy to discuss other ways to organize the instanciation of python classes from files if you have suggestions.

> I think it would be best to convert it into an abstract base class, where every class that implements (inherits from) BaseIO must implement the minimum set of methods defined there.

We could do that to warn for missing `get_init_params()` or a few other methods. I haven't done it because I remember @ceriottm 's take on the topic

> the actual base class seems a lot heavier, implementing lots of default behaviour.

Isn't it the point of a base class ? to implement a shared infrastructure avoiding duplicated code ? Note that is it not an interface, I don't care, here, about the shared parent type ; it is there as you said to provide methods like the `to_dict()` function.

bindings/rascal/representations/coulomb_matrix.py
```diff
@@ -38,8 +39,7 @@ class SortedCoulombMatrix(object):
         Physical Review Letters, 108(5), 58301. https://doi.org/10.1103/PhysRevLett.108.058301
     """
 
-    def __init__(self, cutoff, sorting_algorithm='row_norm', size=10, central_decay=-1, interaction_cutoff=10, interaction_decay=-1,
-                 method='thread', n_workers=1, disable_pbar=False):
+    def __init__(self, cutoff, sorting_algorithm='row_norm', size=10, central_decay=-1, interaction_cutoff=10, interaction_decay=-1):
```
max-veit, line 42: Crazy long line here (the documentation should also be wrapped to 72 characters, see PEP8)

bindings/rascal/utils/cur.py
```diff
@@ -67,7 +69,9 @@ def __init__(self, representation, Nselect, act_on='sample per species', is_dete
         if act_on in ['sample', 'sample per species', 'feature']:
             self.act_on = act_on
         else:
-            raise 'Wrong input: {}'.format(act_on)
+            raise NotImplementedError(
```
max-veit, line 72: `__init__` should never raise a `NotImplementedError` --  this looks like more of a `ValueError` to me.

docs/source/io/io.rst
```diff
@@ -0,0 +1,26 @@
...
+===============================
+
+The library aims at being able to load/save three types of objects: atomic structures,
+their computed representation and trained machine learning models.
+Many tools already exist for reading atomic srtuctures in python and c++ so we rely on
+ASE_ on the python side and
+support the ASE `.json` atomic structure file format on the c++ side
+(:cpp:class:`AtomicStructure <rascal::AtomicStructure>`). An example is available
+in `json_structure.cc` and `spherical_invariants_example.cc`.
+At the moment we do not support saving representations in native Librascal format
```
max-veit, line 13: What about sparse points? It looked like you had an example for saving and loading those via JSON.

max-veit:
> Serializing to JSON means that to serialize to another format one would have to do it again. The rational here is that any python io library like json, msgpack ... knows how to dump a python dictionary so serializing to dictionaries allow for easy extensions.

I don't think there's any point in making it format agnostic, though. The model files will need to be read by the C++ code one way or another (specifically, when running a simulation that doesn't use Python, e.g. with lammps or i-PI), which means implementing anything other than JSON just means more work for us. I think the best way to approach the problem of saving and loading models is to pick a single, language-independent format (e.g. JSON with a schema that we define) and stick with it.

> This is not a workaround, the python and c++ are talking to each other using JSON messages. On the C++ side the equivalent of dictionaries is the json type which is passed to python using JSON messages. It could be some other type of messages though.

See above -- this just further indicates JSON as a good language-independent model storage format.

> > I'm confused what the get_init_params() function is for
> 
> it returns a dictionarry containing all the parameters used in the **init**() function. Thought the name was pretty litteral there.

Honestly no, I didn't think of the `__init__` function when I read that function name. Why is there a need for a function that returns the exact parameters used for initialization? Is there something wrong with just storing the hypers dict instead and giving the option to restore a class from that?

> > why the _set_data() function appears to do nothing in many of the classes that implement it
> 
> This function is meant to recover data that was not part of the initialization, like the sparse point features of the sparse_point class. Some of our classes don't produce such data hence some empty _set_data(). I don't see where the problem is but I am happy to discuss other ways to organize the instanciation of python classes from files if you have suggestions.

Exactly, so this is a specialized function that is only used by one or two classes implementing `BaseIO`, hence it doesn't belong as part of the interface that all classes need to implement.

The issue of external data files is actually kind of tricky, and it deserves some extra discussion. The way this was solved in QUIP was to write the sparse points into a separate file that had to live with the main parameters file; including them in the main XML often just overflows the stack (or some similar gruesome memory error) if you have a large number of sparse points. The solution implemented here (write out large arrays in NumPy format) works from Python, but will obviously be trickier when running without the Python interface -- can Eigen read NumPy arrays?

> Isn't it the point of a base class ? to implement a shared infrastructure avoiding duplicated code ? Note that is it not an interface, I don't care here about the shared parent type ; it is there as you said to provide methods like the `to_dict()` function avoiding as little duplication as possible.

Not really; inheritance is there to implement an "is-a" relationship. When designing an inheritance pattern, you should always ask yourself: "does it make sense to think of the child class as a more specialized version of the base class? In other words, if A inherits from B, does it makes sense to say A _is a_ type of B?" In this case it doesn't, because e.g. SparsePoints is _not_ a BaseIO. There are _much_ simpler, easier ways to reuse code than by inheritance - for instance, implement dictionary hypers serialization as a small factory object that takes an object of a compatible class and spits out properly formatted JSON (with large data fields written out to NumPy arrays).

And when I said `BaseIO` was too "heavy" for an interface class, I was concerned about it trying to be too general. The docstring says it's meant to serialize "arbitrary Python classes" to dictionaries -- but that's not what it should be for. We don't want to reinvent Python's already existing, well-tested serialization methods. What we want instead is a way to save models and hyperparameters (including sparse points, maybe also kernel matrices) to disk in a simple format and be able to restore them later, whether from Python or C++. The current implementation seems a bit ambitious for such a limited task.

max-veit:
(Another interesting take on the inheritance issue: https://blog.codinghorror.com/inherits-nothing/)

bindings/rascal/representations/coulomb_matrix.py
felixmusil: autopep8 does not work very well with [line breaks](https://github.com/hhatto/autopep8/issues/393) (one of the many example I found after a quick search). So the tool does not work at the moment. Should we ga back to format by hand ?

bindings/rascal/utils/cur.py
felixmusil: > __init__ should never raise a NotImplementedError

why not ?

docs/source/io/io.rst
felixmusil: sparse points are not a representation but basis points for the regression. And yes they can be saved to JSON format

felixmusil:
>  which means implementing anything other than JSON just means more work for us.

the json library we use supports JSON, a few binary JSON and message pack formats so without additional work we could use these format to comunicate with the c++ side. The main reson for departing from JSON would be speed, we might not need it in the end but it does not hurt much to prepare for it.

> See above -- this just further indicates JSON as a good language-independent model storage format.

it is as good as any of the other alternative I cited above. JSON is a choice by default at the moment that might not be the best in the ling run.

> Why is there a need for a function that returns the exact parameters used for initialization? Is there something wrong with just storing the hypers dict instead and giving the option to restore a class from that?

To instanciate a class we need to go through the __init__ function. There is just one of these so to allow for 2 instanciation mode we would need to have a switch on input value like an additional argument, e.g. a `file_data = None` field, to switch from. All other arguments would have default values. Then it is in the switch that you make sure how close you are to the original object. This could be done in a generic fashion, [with some caveats](https://stackoverflow.com/a/244116), by saving he __dict__ field of the object / injecting file_data into the __dict__ field of the new instance. To keep some control over the fields that would be saved/injected we probably would need to define a list of such fields.
I am not a big fan of this option as it forces us to make our objects to have all initialization arguments defaulted with some an additional `file_data` field. By not going through the __init__ step we make many modifications to this step breaking backward compatibility. 
Let's discuss that.

> Exactly, so this is a specialized function that is only used by one or two classes implementing BaseIO, hence it doesn't belong as part of the interface that all classes need to implement.

just to be clear, are you suggesting that an `if` on the availability of `_set_data()` or `try` instead of just calling it in the loading step would be better so that one does not have to write down empty functions ?

> The issue of external data files is actually kind of tricky, and it deserves some extra discussion. 

Agreed

> The way this was solved in QUIP was to write the sparse points into a separate file that had to live with the main parameters file

I don't think we have to do that hopefully since we don't use XML.

> The solution implemented here (write out large arrays in NumPy format) works from Python, but will obviously be trickier when running without the Python interface -- can Eigen read NumPy arrays?

This is not meant for being used in the python/c++ model serialization business as you noted so all numpy arrays should be converted to lists in this case. Eigen is not an io library so it does not but other library do but if it comes this I would move to a more suited format like hdf5.

> Not really; inheritance is there to implement an "is-a" relationship. When designing an inheritance pattern, you should always ask yourself: "does it make sense to think of the child class as a more specialized version of the base class? In other words, if A inherits from B, does it makes sense to say A is a type of B?"

I mostly agree with you in c++ (there are huge code base that use inheritance for functionality injection purposes) but in python that is not so clear since all classes inherit from `object` meaning that all our classes are somehow conceptually related, which is not true obviouly. 

> There are much simpler, easier ways to reuse code than by inheritance - for instance, implement dictionary hypers serialization as a small factory object that takes an object of a compatible class and spits out properly formatted JSON (with large data fields written out to NumPy arrays).

I am not sure to understand your simpler solution.

> it's meant to serialize "arbitrary Python classes" to dictionaries -- but that's not what it should be for. 

well arbitrary is a bit of an obverstatement when taken out of context since they have to follow some rules to be compatible. So it should be understood that classes that follow this set of rules can be serialized. My concern with being more specific about our use case is that it breaks or requires too many lines for what it is for.

bindings/rascal/representations/coulomb_matrix.py
max-veit: Hmm, perhaps, as long as autopep8 doesn't mess with it again after hand-formatting

bindings/rascal/utils/cur.py
max-veit: Because it would suggest that `__init__` isn't implemented, which is nonsense. In fact, [`NotImplementedError`](https://docs.python.org/3/library/exceptions.html#NotImplementedError) is only intended for the very specific purpose of indicating "abstract" methods in base classes. (The similarly-named [`NotImplemented`](https://docs.python.org/3/library/constants.html#NotImplemented) has a different, also very specific purpose.)

The correct exception for indicating a problem with the input parameters of a method is [`ValueError`](https://docs.python.org/3/library/exceptions.html#ValueError).

docs/source/io/io.rst
max-veit: Ok, sure, but there's a way to save a _model_ (sparse points with representation hypers), which is what most users care about. I don't really see a use case for saving the representation class itself, if the calculated representation can be copied into a `SparsePoints` and saved that way.

If there is such a special use case, I would like to see separate documentation for it (since it seems like such a special case). Otherwise, I'm assuming there's a reason you didn't have to implement Representation serialization to get model serialization working for most use cases.

bindings/rascal/utils/io.py
```diff
@@ -0,0 +1,376 @@
...
+
+    + _get_init_params is expected to return a dictionary containing all the
+    parameters used by the __init__() methods.
+
+    + _get_data is expected to return a dictionary containing all the data
+    that is not set by the initialization of the class.
+
+    + _set_data is expected to set the data that has been extracted by _get_data
+    """
+    @abstractmethod
```
max-veit, line 222: Nitpick here, but if the method implements default behaviour then it's not an abstract method (and this is not an abstract class). I recommend either removing the default behaviour (making everything `pass`), or better, removing the `@abstractmethod` from `_get_data()` and `_get_init_params()`, which would then let you remove the `_get_data()` overrides from the classes that don't need to implement it.

bindings/rascal/utils/io.py
```diff
@@ -0,0 +1,376 @@
...
+    return CURRENT_VERSION
+
+
+def get_supported_io_versions():
+    return list(dict2obj.keys())
+
+
+class BaseIO(ABC):
+    """
+    Definition of the interface of a python class that can be serialized by the
```
max-veit, line 211: [PEP 257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings) specifies that the "summary line" of a multi-line docstring must fit on one line. This summary could be shortened e.g. to "Interface of a Python class serializable by to_dict()".  If a further description is necessary, it should be added below the summary, separated by a blank line.

bindings/rascal/utils/io.py
```diff
@@ -0,0 +1,376 @@
...
+        state = dict(data=obj._get_data(),
+                     init_params=obj._get_init_params())
+    else:
+        raise ValueError(
+            'input object: "{}" does not inherit from "BaseIO"'.format(obj))
+    return state
+
+
+def to_dict(obj, version=CURRENT_VERSION):
+    """recursirvely convert the python object obj into a dictionary (serialized
```
max-veit, line 246: summary line too long; also spelling on "recursively". I would say "Recursively serialize to dict via the BaseIO interface".

bindings/rascal/utils/io.py
```diff
@@ -0,0 +1,376 @@
...
+                            ll.append(to_dict(val, version))
+                        else:
+                            ll.append(val)
+                    state[name][k] = ll
+    data = obj2dict[version](obj.__class__, state)
+    return data
+
+
+def from_dict(data):
+    """recursirvely convert a python dictionary describing an object inherited
```
max-veit, line 273: summary line too long.

Strunk & White Rule 17: Omit needless words.

bindings/rascal/utils/io.py
```diff
@@ -0,0 +1,376 @@
...
+        else:
+            # just transfer the data
+            data_obj[name] = entry
+
+    obj = dict2obj[version](data_obj)
+    return obj
+
+
+def to_file(fn, obj, version=CURRENT_VERSION):
+    """Saves the object to a file name 'fn' using the to_dict()
```
max-veit, line 307: summary line too long

bindings/rascal/utils/io.py
```diff
@@ -0,0 +1,376 @@
...
+            raise RuntimeError(
+                'The file: {}; does not contain a valid dictionary representation of an object.'.format(fn))
+
+    else:
+        raise NotImplementedError(
+            'Unknown file extention: {}'.format(file_extension))
+
+
+def _dump_npy(fn, data, class_name):
+    """Saves numpy array that are large into a file different from the
```
max-veit, line 343: summary line way too long (longer description separated by a blank line)

bindings/rascal/utils/io.py
```diff
@@ -0,0 +1,376 @@
...
+            v_bfn = os.path.basename(v_fn)
+            data[k] = v_bfn
+            np.save(v_fn, v)
+
+        elif is_npy(v) is True:
+            data[k] = ['npy', v.tolist()]
+
+
+def _load_npy(data, path):
+    """Loads a numpy array saved using _dump_npy(). The array is mmaped
```
max-veit, line 366: summary line too long

docs/source/io/io.rst
```diff
@@ -0,0 +1,26 @@
...
+but on the python side they can be converted to numpy arrays with the `get_features()`
+function and then saved.
```
max-veit, line 15: Use case?

max-veit:
> > (Another interesting take on the inheritance issue: https://blog.codinghorror.com/inherits-nothing/)
> 
> I am not really familiar with .NET stuff which might explain why I missed the point here.

It's not about .NET; this is good advice for any object-oriented programming language. The key quote is:

> **inheritance is only one way of many to achieve code reuse.**

with composition being the simpler, less error-prone alternative. In Python you can even just trust duck-typing to do its thing.

bindings/rascal/utils/io.py
```diff
@@ -0,0 +1,376 @@
...
+                elif isinstance(v, list):
+                    # make sure list of objects are properly serialized
+                    ll = []
+                    for val in v:
+                        if isinstance(val, BaseIO) is True:
+                            ll.append(to_dict(val, version))
+                        else:
+                            ll.append(val)
+                    state[name][k] = ll
+    data = obj2dict[version](obj.__class__, state)
```
Luthaf, line 268: Does that work for 'weird' python objects too? Like 

```python
# circular object
foo = {"bar: 3}
foo["blee"] = foo

# object containing C ressources (pointers, file descriptor, ...)
bar = {"file": open('tmp.xyz')}
```

docs/source/io/io.rst
felixmusil: > a way to save a model (sparse points with representation hypers)

the KRR model corresponds to sparse points, weights, kernel and representation. So each pieces can be saved on its own hence be saved in the bundle that is the KRR model.

> If there is such a special use case, I would like to see separate documentation for it (since it seems like such a special case). Otherwise, I'm assuming there's a reason you didn't have to implement Representation serialization to get model serialization working for most use cases.

I am really not following you here.

bindings/rascal/utils/io.py
felixmusil: the `@abstractmethod` wrapper is here to raise an error if the method is not implemented in the child class, so whatever happens in this methods is not used as a default and has to be called explicitly. So the fact that it is not `pass` is so that it is more convenient when one copy/paste these methods and writes the actual implementation an the actual class.

docs/source/io/io.rst
felixmusil: what are you talking about exactly ?

bindings/rascal/utils/io.py
felixmusil: > # circular object

I don't think we have a reasonable use case for suporting this added complexity so not at the moment. It is not hard to suport but again I would rather not implement it. Are you thinking of making a check so that en error is raised ?

> # object containing C ressources (pointers, file descriptor, ...)

file descriptor would not be possible. But the filename could be handled like it is done for large numpy arrays. 
As for objects that contain C++ resources, they have to call a `to_dict()` function implemented in the bindings that should use the c++ json library to achieve that. What use case do you have in mind exactly ?

Luthaf: I don't have specific use-cases right now, these are mainly tricky things to get right when doing Python serialization. So having a check for them would be good, or even better explicitly disallowing everything we don't know about. I would only allow string, numbers, list of the previous and dict of the previous (taking care of circular objects); and throw an error on anything else.

felixmusil: Do you have an idea on how to detect circularity ? setting a max depth of serialization seems simple but is there a better way ?

Luthaf: not on the top of my mind. I know the json module does it, so you could take inspiration from there (or just piggy-back on it `try: json.dump(x); except ValueError as e: if str(e) == "Circular reference detected" raise`)

felixmusil: ok I will set a max recursion parameter and raise for circular reference when it is reached

max-veit:
One more thing I noticed while trying this out locally: The pretty-printed JSON is too... well, long; lists that should be printed on one line are split into one entry per line, even if that entry is just an int.

It would be nice to have something more readable; there are packages [in Javascript](https://www.npmjs.com/package/json-stringify-pretty-compact) and [in Haskell](https://github.com/haroldcarr/json-pretty-print) to print human-readable, pretty JSON; do you know of something similar in Python? If all else fails, we can always just turn off pretty-printing on the Python side and filter the JSON file through one of these post-processors if we need the file to be human-readable.

docs/source/io/io.rst
max-veit: What's the use case for saving representations (rather than saving them as sparsepoints)? Why would anyone want to do this?

bindings/rascal/utils/io.py
max-veit: That doesn't address the issue. If the method has default behaviour, then it should not have `@abstractmethod`. Why would one copy/paste if the method already has default behaviour defined in the base class? That's such an inefficient and error-prone way to do code re-use when you already have inheritance working for you here, not to mention it violates [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).

felixmusil: you should tell to the people that implemented the abc module not me.

docs/source/io/io.rst
felixmusil: In ended up saving the feature matrices on many occasions but if I had to name one case it would be as part of a full covariance GPR/KRR model. It should not be saved in sparsepoints because you want to sum over the atoms in the structures of the training set too.

bindings/rascal/utils/io.py
max-veit: Right, on going back and reading the PEP on ABCs, it seems it's mostly about defining standardized tests for whether a class _conforms_ to an interface. Strictly speaking, all this ABC stuff only serves to translate the call `isinstance(obj, BaseIO)`, e.g. here: https://github.com/cosmo-epfl/librascal/blob/e25ca3edd491052d85066a5dd82fec74c6ebbee1/bindings/rascal/utils/io.py#L237 into checking whether the object implements all the methods marked `@abstractmethod` in the `BaseIO` class.

So on reconsidering, the `@abstractmethod` marker here should definitely be kept, and my previous comments about this not being an abstract class were wrong (based on Java-style thinking). But to improve robustness, we should use the mechanism that the people who implemented the abc module designed to handle just this case: The `super()` mechanism (see the note at the bottom of [this doc page](https://docs.python.org/3/library/abc.html#abc.abstractmethod)). Every class that wants to use the default behaviour for e.g. the `_get_data()` method should instead write that method as

```python
def get_data():
    return  super()._get_data()
```

instead. It's more verbose but more robust (and definitely not "shiny"): That way, if we later decide to change the default behaviour (e.g. return some default keys, or throw an exception), it doesn't have to be changed in every implementing subclass.

docs/source/io/io.rst
max-veit: Ok, so in that case the "use case" is if you want to save the features, keeping their association with individual structures (which the sparsepoints discards). It seems the easiest workaround would be to save both the feature matrix (using sparsepoints) and a separate array telling you how many atoms go in each structure. Wouldn't you need such an array anyways when extracting the feature matrix with `get_features()`?

bindings/rascal/models/IP_ase_interface.py
```diff
@@ -41,3 +42,14 @@ def calculate(self, atoms=None, properties=['energy'],
...
+    def _get_init_params(self):
+        init_params = dict(model=self.model, representation=self.representation)
+        init_params.update(**self.kwargs)
+        return init_params
+
+    def _set_data(self, data):
+        pass
+
+    def _get_data(self):
+        return dict()
```
max-veit, line 55:
```suggestion
        return super()._get_data()`
```

bindings/rascal/models/IP_ase_interface.py
```diff
@@ -41,3 +42,14 @@ def calculate(self, atoms=None, properties=['energy'],
...
         self.results['free_energy'] = energy
         self.results['forces'] = forces
+
+    def _get_init_params(self):
+        init_params = dict(model=self.model, representation=self.representation)
+        init_params.update(**self.kwargs)
+        return init_params
+
+    def _set_data(self, data):
+        pass
```
max-veit, line 52: I would even suggest it here; maybe in the future we will want to implement default behaviour for `_set_data()`.

```suggestion
        super()._set_data(data)
```

bindings/rascal/models/kernels.py
```diff
@@ -94,17 +94,30 @@ def __init__(self, representation, name='Cosine', kernel_type='Full', target_typ
...
+                           kernel_type=self.kernel_type,
+                           target_type=self.target_type)
+        init_params.update(**self._kwargs)
+        return init_params
+
+    def _set_data(self, data):
+        pass
+
+    def _get_data(self):
+        return dict()
```
max-veit, line 120:
```suggestion
        return super()._get_data()
```

bindings/rascal/models/kernels.py
```diff
@@ -94,17 +94,30 @@ def __init__(self, representation, name='Cosine', kernel_type='Full', target_typ
...
+    def _get_init_params(self):
+        init_params = dict(representation=self._rep,
+                           name=self.name,
+                           kernel_type=self.kernel_type,
+                           target_type=self.target_type)
+        init_params.update(**self._kwargs)
+        return init_params
+
+    def _set_data(self, data):
+        pass
```
max-veit, line 117:
```suggestion
        super()._set_data(data)
```

bindings/rascal/models/krr.py
```diff
@@ -82,8 +83,24 @@ def predict(self, managers, compute_gradients=False):
...
         return self.weights
 
+    def _get_init_params(self):
+        init_params = dict(weights=self.weights, kernel=self.kernel,
+                           X_train=self.X_train,
+                           self_contributions=self.self_contributions)
+        return init_params
+
+    def _set_data(self, data):
+        pass
```
max-veit, line 93:
```suggestion
        super()._set_data(data)
```

bindings/rascal/models/krr.py
```diff
@@ -82,8 +83,24 @@ def predict(self, managers, compute_gradients=False):
...
+
+    def _get_data(self):
+        return dict()
```
max-veit, line 96:
```suggestion
        return super()._get_data()
```

bindings/rascal/representations/coulomb_matrix.py
```diff
@@ -109,3 +107,20 @@ def get_size(self, managers):
...
+            sorting_algorithm=self.hypers['sorting_algorithm'],
+            size=self.hypers['size'],
+            central_decay=self.hypers['central_decay'],
+            interaction_cutoff=self.hypers['interaction_cutoff'],
+            interaction_decay=self.hypers['interaction_decay']
+        )
+        return init_params
+
+    def _set_data(self, data):
+        pass
```
max-veit, line 123:
```suggestion
        super()._set_data(data)
```

bindings/rascal/representations/coulomb_matrix.py
```diff
@@ -109,3 +107,20 @@ def get_size(self, managers):
...
+
+    def _get_data(self):
+        return dict()
```
max-veit, line 126:
```suggestion
        return super()._get_data()
```

bindings/rascal/representations/spherical_covariants.py
```diff
@@ -214,3 +213,31 @@ def get_keys(self, species):
...
+            gaussian_sigma_constant=gaussian_density['gaussian_sigma']['value'],
+            lam=self.hypers['lam'],
+            cutoff_function_type=cutoff_function['type'],
+            radial_basis=radial_contribution['type'],
+            cutoff_function_parameters=self.cutoff_function_parameters,
+        )
+        return init_params
+
+    def _set_data(self, data):
+        pass
```
max-veit, line 240:
```suggestion
        super()._set_data(data)
```

bindings/rascal/representations/spherical_covariants.py
```diff
@@ -214,3 +213,31 @@ def get_keys(self, species):
...
+
+    def _get_data(self):
+        return dict()
```
max-veit, line 243:
```suggestion
        return super()._get_data()
```

bindings/rascal/representations/spherical_expansion.py
```diff
@@ -235,3 +230,30 @@ def get_keys(self, species):
...
+            gaussian_sigma_constant=gaussian_density['gaussian_sigma']['value'],
+            cutoff_function_type=cutoff_function['type'],
+            radial_basis=radial_contribution['type'],
+            optimization_args=self.optimization_args,
+            cutoff_function_parameters=self.cutoff_function_parameters,
+        )
+        return init_params
+
+    def _set_data(self, data):
+        pass
```
max-veit, line 256:
```suggestion
        super()._set_data(data)
```

bindings/rascal/representations/spherical_expansion.py
```diff
@@ -235,3 +230,30 @@ def get_keys(self, species):
...
+
+    def _get_data(self):
+        return dict()
```
max-veit, line 259:
```suggestion
        return super()._get_data()
```

bindings/rascal/representations/spherical_invariants.py
```diff
@@ -285,3 +284,33 @@ def get_keys(self, species):
...
+            gaussian_sigma_constant=gaussian_density['gaussian_sigma']['value'],
+            cutoff_function_type=cutoff_function['type'],
+            radial_basis=radial_contribution['type'],
+            optimization_args=self.optimization_args,
+            cutoff_function_parameters=self.cutoff_function_parameters,
+        )
+        return init_params
+
+    def _set_data(self, data):
+        pass
```
max-veit, line 313:
```suggestion
        super()._set_data(data)
```

bindings/rascal/representations/spherical_invariants.py
```diff
@@ -285,3 +284,33 @@ def get_keys(self, species):
...
+
+    def _get_data(self):
+        return dict()
```
max-veit, line 316:
```suggestion
        return super()._get_data()
```

bindings/rascal/utils/io.py
```diff
@@ -0,0 +1,387 @@
...
+    def _set_data(self, data):
+        pass
+
+    @abstractmethod
+    def _get_init_params(self):
+        return dict()
+
+
+def _get_state(obj):
+    if isinstance(obj, BaseIO) is True:
```
max-veit, line 237:
```suggestion
    if isinstance(obj, BaseIO):
```
(`is True` is redundant)

bindings/rascal/utils/io.py
```diff
@@ -0,0 +1,387 @@
...
+        recursion_depth += 1
+        
+    state = _get_state(obj)
+
+    # loop over the 2 fields of state
+    for name, entry in state.items():
+        if isinstance(entry, dict):
+            # case of potentially nested objects
+            for k, v in entry.items():
+                if isinstance(v, BaseIO) is True:
```
max-veit, line 265:
```suggestion
                if isinstance(v, BaseIO):
```

bindings/rascal/utils/io.py
```diff
@@ -0,0 +1,387 @@
...
+                    state[name][k] = to_dict(v, version, recursion_depth)
+                elif isinstance(v, list):
+                    # make sure list of objects are properly serialized
+                    ll = []
+                    for val in v:
+                        if isinstance(val, BaseIO) is True:
```
max-veit, line 271:
```suggestion
                        if isinstance(val, BaseIO):
```

bindings/rascal/utils/io.py
```diff
@@ -0,0 +1,387 @@
...
+def from_dict(data):
+    """Recursirvely deserialize from dict via the BaseIO interface."""
+    # temporary dictionary to hold the object being recovered
+    data_obj = dict()
+    version = data['version']
+    for name, entry in data.items():
+        if isinstance(entry, dict):
+            data_obj[name] = dict()
+            for k, v in entry.items():
+                if is_valid_object_dict[version](v) is True:
```
max-veit, line 289:
```suggestion
                if is_valid_object_dict[version](v):
```

bindings/rascal/utils/io.py
```diff
@@ -0,0 +1,387 @@
...
+                    # in case of nested objects
+                    data_obj[name][k] = from_dict(v)
+                elif isinstance(v, list):
+                    # in case of list make sure to handle list of serialized
+                    # objects
+                    ll = []
+                    for val in v:
+                        if is_valid_object_dict[version](val) is True:
```
max-veit, line 297:
```suggestion
                        if is_valid_object_dict[version](val):
```

bindings/rascal/utils/io.py
```diff
@@ -0,0 +1,387 @@
...
+    filename, file_extension = os.path.splitext(fn)
+    if file_extension == '.json':
+        data = load_json(fn)
+        version = data['version']
+        if is_valid_object_dict[version](data):
+            _load_npy(data, path)
+            return from_dict(data)
+        else:
+            raise RuntimeError(
+                'The file: {}; does not contain a valid dictionary representation of an object.'.format(fn))
```
max-veit, line 343: long line

docs/source/io/io.rst
max-veit: Since the `get_features()` workaround was good enough for everything you tried -- along with perhaps the number of atoms per frame -- why do we need a "native librascal format" for representations? Just make the workaround official and that's it, until we find an application where it's not good enough.

max-veit:
Also, the Python json formatter you linked looks nice; it should definitely be mentioned in the docs (I wouldn't actually integrate it as a third-party tool yet; we can discuss that later).

docs/source/io/io.rst
felixmusil: > Wouldn't you need such an array anyways when extracting the feature matrix with get_features()?

I am not sure about what you mean. Anyhow that's basically the plan of @Luthaf for the new dataholder type that would be used to store representations without a structure manager. This new dataholder will be easy to serialize then.

docs/source/io/io.rst
felixmusil: > why do we need a "native librascal format" for representations?

we have one since we have some dedicated dataholder in c++. It is just not convenient to serialize for the moment so we don't but in the not so near future (hopefully) we will have a simpler dataholders (that can be functional without a structure manager).
I would not recommend using `get_features` atm because it basically does not allow to used rascal on the features (model, kernel...).

bindings/rascal/utils/io.py
felixmusil: but more explicit

max-veit: No. It's just confusing.

felixmusil: ok it is confusing to you and it is the opposite to me.

felixmusil: does it really matter ?

max-veit: Yes. It's 100% redundant and _will_ be confusing to anyone else who reads the code, because it implies that `isinstance` (or the function defined later for testing a valid dict) might return something that's not a true/false value.

felixmusil: also note that python implicitly evaluates many things to True in the if statements which can lead to unexpected behaviour.

felixmusil: > will be confusing to anyone else who reads the code, because it implies that isinstance (or the function defined later for testing a valid dict) might return something that's not a true/false value.

Aren't you over reading into things. Also I would not presume so much about what someone else might think while reading this statement.

docs/source/io/io.rst
max-veit: Now I'm more confused than ever. You're recommending against using `get_features()`, even though the next sentence in the docs acknowledges it as a workaround?

Ok, whatever, I'm willing to wait for the new dataholder type for an "official" way to save representations with all the necessary extra information.

bindings/rascal/utils/io.py
max-veit: PEP8 is explicitly against comparing Boolean values to True or False: https://www.python.org/dev/peps/pep-0008/#programming-recommendations. There is no more to say about this.

PR #270, opened 2020-04-20
Feat/reverse grad storage order

felixmusil:
Change the storage order for the gradients w.r.t. atomic positions. Before we were storing them like `\grad_i X_j` with `i` the slow index and `X_j` the representation vector for atom `j` while we now store them like `\grad_j X_i`. The main reason for this change is that it allows for further optimizations in the kernel gradient computation when the kernel depends on the species of the central atom. 
To make sure this change was properly implemented I added a numerical kernel gradient routine and a test for the compatibility of the numerical kernel gradient with the analytical ones.

This PR contains changes from #265


max-veit:
Could you please explain how this PR relates to #228 ? I'm still confused as to what `i` and `j` refer to above; what is the "slow index"? Which index refers to the atom that's moving, i.e. the one with respect to whose position the gradient is being taken?

felixmusil:
> Could you please explain how this PR relates to #254 ?

it's reverting the storage order to as it was before this PR.

> I'm still confused as to what i and j refer to above

as always atom indicies. so i would be the central atom and j a neighbor.

> what is the "slow index"? 

like the row index of a matrix stored in column major format is the slow index, one of i or j has to be the slow index while the other is the fast index. 

> Which index refers to the atom that's moving

the index that is associated with the gradient.

src/rascal/models/numerical_kernel_gradients.hh
```diff
@@ -0,0 +1,115 @@
...
+    // displace the atom
+    atomic_structure.displace_position(i_atom, disp);
+    // make sure all atoms are in the unit cell
+    atomic_structure.wrap();
+    // update the neighborlist
+    manager->update(atomic_structure);
+    // compute the representation of the modified structure
+    calculator.compute(manager);
+    // compute the kernel of the modified structure
+    std::vector<std::remove_const_t<Manager>> managers{};
```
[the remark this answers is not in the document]
felixmusil: I tried but const correctness here is a rabit hole that should be adressed in a standalone PR I think.

examples/Spherical_invariants_numerical_kernel_gradients.ipynb
```diff
@@ -0,0 +1,985 @@
+{
```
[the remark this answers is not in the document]
felixmusil: removed it.

tests/test_calculator.hh
```diff
@@ -865,25 +872,50 @@ namespace rascal {
...
       Key_t center_key{center.get_atom_type()};
       size_t n_entries_per_key{static_cast<size_t>(data_sparse.get_nb_comp())};
+      size_t n_entries_center{n_entries_per_key * keys_center.size()};
       size_t n_entries_neighbours{0};
       // Count all the keys in the sparse gradient structure where the gradient
       // is nonzero (i.e. where the key has an entry in the structure)
-      for (auto neigh : center.pairs_with_self_pair()) {
+      for (auto neigh : center.pairs()) {
+        if (this->structure_manager->is_ghost_atom(neigh)) {
+          // Don't compute gradient contributions onto ghost atoms
```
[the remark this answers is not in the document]
felixmusil: here I skip the ghost atoms because the finite difference gradients are only computed for the atoms in the unit cell.

tests/test_calculator.hh
```diff
@@ -952,6 +1006,57 @@ namespace rascal {
...
+     * Here we refer to atom_i as the center atom in the pair and atom_j
+     * as the neighboring atom in the pair that has been wrapped in the
+     * unit cell.
+     *
+     * @return std::vector of ClusterRefKeys or order 2 (pair keys) of all pairs
+     *         (j, i') where i' is either i or any of its periodic images within
+     *         the cutoff of j. The atom j, on the other hand, must be a real
+     *         atom (not a ghost or periodic image).
+     */
+    PairRefKey_t swap_pair_ref(const PairRef_t & pair_ref) {
```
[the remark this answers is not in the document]
felixmusil: yes it has.

src/rascal/models/numerical_kernel_gradients.hh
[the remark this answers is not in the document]
max-veit: Noted in #284

bindings/rascal/utils/filter_utils.py
```diff
@@ -62,7 +62,6 @@ def get_index_mappings_sample_per_species(managers, sps):
 
     return strides_by_sp, global_counter, map_by_manager, indices_by_sp
 
-
```
max-veit, line 65: Is this auto-formatting? PEP8 says two blank lines between top-level functions, so I'd just keep it that way.

src/rascal/models/numerical_kernel_gradients.hh
```diff
@@ -0,0 +1,111 @@
...
+    managers.emplace_back(manager);
+    math::Matrix_t KNM = kernel.compute(calculator, managers, sparse_points);
+    // reset neighborlist to the original structure
+    manager->update(structure_copy.template get<AtomicStructure<ThreeD>>());
+    return KNM;
+  }
+
+  template <class KernelImpl, class Calculator, class Manager,
+            class SparsePoints>
+  math::Matrix_t compute_numerical_kernel_gradient(
```
max-veit, line 65: Needs documentation (this function and the one below). Please explain that these utilities are intended for finite-difference numerical computation of the gradient of a kernel-based model, for the case where analytical gradients are not available. A mention of this functionality in the developer doc pages would also be most helpful.

bindings/rascal/utils/filter_utils.py
felixmusil: done

src/rascal/models/numerical_kernel_gradients.hh
felixmusil: how is it now ?

max-veit: Better. I'd say "gradient... for a _single_ atomic structure" to make it clearer how this function differs from the one below, and put in addition to just `@see the other function` a line for the `@param manager` that makes this function different from the multi-manager one.

Could you add a mention of this functionality to the developer docs, e.g. [here](https://cosmo-epfl.github.io/librascal/dev_guide/how_to_add_representation.html#implement-and-test-gradients) or [here](https://cosmo-epfl.github.io/librascal/dev_guide/how_to_test.html#testing-gradients)? We'll also need to work on updating the developer docs to reflect the existence of kernel-based models (the "Models" section is currently a stub).

docs/source/dev_guide/how_to_add_representation.rst
```diff
@@ -136,6 +136,12 @@ the GTO and DVR radial basis.  Until we come up with a general, standard way of
 implementing gradients for any representation, please see those implementations
 for guidance.
 
+Alternativelly, the utility :cpp:func:`compute_numerical_kernel_gradients` can
```
max-veit, line 139:
```suggestion
Alternatively, the utility :cpp:func:`compute_numerical_kernel_gradients` can
```

PR #278, opened 2020-05-22
Feat/change python formater

felixmusil:
It seems that autopep8 is not completely satisfactory so here a PR to discuss swapping for another python formater. After some searches I found the [black](https://black.readthedocs.io/en/stable/) tool which seems both simple and effective. The drawback is that there are no options so either we like it or we don't. An alternative could be [yapf](https://github.com/google/yapf) but it seems that it is not fully deterministic.

Please tell me your opinion on the newly formated python files and if you like the `black` formating.


ceriottm:
This is *very* relevant - discussing on the i-PI front where they suggest using black. Do you want to join the discussion? I was inclined to ask them to move over to autopep, that seems better established, but if you are experiencing problems I'd be happy to go the other direction.

felixmusil:
yes `autopep8` has been around for longer but the resulting formated code is not always very statisfactory (80 caracters per line not properly enforced, ...). So even if `black` is still in beta, it seems that it is more effective at enforcing one formating standard. The question then is whether this standard is to our liking.

ceriottm:
might be because 80 characters/line are too few for python. what happens if we relax that to 90 (which would be the black standard)?

felixmusil:
the line length is just one such example.

max-veit:
`autopep8` gives some pretty weird formatting in other cases (there was one case where it was inserting a line-break into an empty pair of parentheses). The advantage of using `black` is that it's opinionated and consistent enough to quash any useless debates we might otherwise get into about formatting minutiae.

Luthaf:
I am fine with either yapf or black, I've used both on other codes and both works nicely and produce good results. We may want to spend a bit of time to tune them so that they match our coding style/linter.

However, this PR will need to be rebased/re-created on top of master, since it contains 68 unrelated commits.

felixmusil:
we might want to use flake8 to check the python code in a similar way as cpplint checks the C++ code.

felixmusil:
adding flake8 and ensuring that the codebase follows its directives is too much work for this PR so it should be done at a later time through #283.

Luthaf:
Black should be added to `requirements.txt` to replace autopep8 there. Also, I think we should add a "black was run" CI check as part of this PR, using `black --check`.

cmake/black.cmake
```diff
@@ -17,7 +19,23 @@ if(BLACK)
     message(STATUS "BLACK auto formatter: ${BLACK}")
     # common target to concatenate all BLACK.py targets
     add_custom_target(${BLACK_TARGET})
+    add_custom_target(${BLACK_LINT_TARGET})
     set(BLACK_FOUND TRUE)
+
+    # run black on python files that are at the root directory
+    set(ROOT_DIR_FILES "${CMAKE_CURRENT_SOURCE_DIR}/setup.py")
+    add_custom_target(black.setup
+        COMMAND ${BLACK} "-l 79" "-v" ${ROOT_DIR_FILES}
```
Luthaf, line 28: 79 characters line are a bit short, what would you think of using 100 or 120 characters lines?

examples/center_select_example.py
```diff
@@ -11,40 +11,57 @@
...
 from rascal.neighbourlist.structure_manager import (
-        mask_center_atoms_by_species, mask_center_atoms_by_id)
+    mask_center_atoms_by_species,
+    mask_center_atoms_by_id,
+)
 
 
-molecules = ase.io.read('../reference_data/inputs/small_molecules-1000.xyz', ':100')
+molecules = ase.io.read(
+    "../reference_data/inputs/small_molecules-1000.xyz", ":100"
```
Luthaf, line 20: This is something that would look nicer with more than 79 characters per line

cmake/black.cmake
felixmusil: I would be more than happy to do that but we need to check with the others too. I would go for 100 characters lines.

max-veit: 100 is too long, but 88 (the default) works for me -- it's just about at the limit of what I can fit on split screens without wrapping.

CMakeLists.txt
```diff
@@ -204,8 +204,8 @@ endif()
 
 # target to run cpplint.py for all configured sources
 set(LINT_TARGET lint CACHE STRING "Name of C++ and python style checker target")
```
Luthaf, line 206: Why is this a configurable variable? What's the use case here?

cmake/black.cmake
Luthaf: I don't know where the 88 values come from: Python PEP8 is 79, most coding guides are 80, Rust, Java on Android and a few other default to 100.

I've found this article defending the 80 chars per line limit for the same reason: you can fit two panels on a standard 1080p monitor: https://nickjanetakis.com/blog/80-characters-per-line-is-a-standard-worth-sticking-to-even-today

I typically use tabs instead of split panels so I am not affected by this issue; but it seems to be a very valid reason (in addition to the PEP8 standard 79) to stay at 79 😃

max-veit: The rationale for the 88-character default used in "black" is here: https://github.com/psf/black/blob/master/docs/the_black_code_style.md#line-length (I'd also recommend watching the talk linked under "90-ish").

I usually use tabs as well for working on separate files, but side-by-side splits can be very useful in some cases (like diffing versions of a file).

CMakeLists.txt
felixmusil: That's how it was done for the other linter/formater so I copied it but it's true that it does not make much sense. I removed it from the configurable variable as well as the others.

cmake/black.cmake
felixmusil: So should the 88 default be applied ? I don't have any preferences on the question.

Luthaf: I'm happy with 88 too, we can also ask for more feedback on Slack if you think this is necessary

felixmusil: I changed the setting so that anyone can see the result. It feels much more reasonable to me.

mastricker: I would argue to take the same as for `c++`, which is currently at 80 characters (cpplint config). It makes the code look more homogeneous.

max-veit: Hmm, there's no reason _a priori_ to apply the same style rules to C++ and Python code. They're different languages with different syntax. In particular, C++ is a free-form language, which makes it much easier to keep lines to a certain length (you can put line breaks basically wherever you choose), whereas Python is much more restrictive in where line breaks can appear. If allowing slightly longer lines in Python gives us more readable code, then let's go for it.

mastricker: I completely agree. But the question was about an opinion, you got mine. And it is based on this:
My screen width, resolution, and editor space does not change with the language I am using. Therefore, practically speaking, the same spacial restrictions apply to different languages.

Where necessary, I am not opposing using longer lines - we had that option already previously also on the c++ side with the clang-format on/off and on the python side.

ceriottm:
I think that for C the 80 characters is a very long-lived standard
(although recently Linus himself has come out against it) so my worry would
be that linters and code formatters are probably
tuned to function well for 80 chars, and we might end up with
uglier  autoformatting if we change the length. Also the 88 limit to me
makes sense because it's the limit chosen by the guys whose
linter we use.

max-veit:
> I think that for C the 80 characters is a very long-lived standard (although recently Linus himself has come out against it) so my worry would be that linters and code formatters are probably tuned to function well for 80 chars, and we might end up with uglier autoformatting if we change the length.

This is all addressed in their [code style page](https://github.com/psf/black/blob/master/docs/the_black_code_style.md). The slightly longer line typically gives nicer autoformatting and more compact code (at the expense of making lines so long they're hard to read, at the extreme end of the spectrum).

> Also the 88 limit to me makes sense because it's the limit chosen by the guys whose linter we use.

That's the idea, yes. It's trivial to tune this limit with any good linter.

max-veit:
OK, this has been stalled without consensus for a long time, so if there's no objections by tomorrow I'll just go ahead and approve this. Sound good?

ceriottm:
:+1:

max-veit (approved):
let's do this

Issue #282, opened 2020-06-23
Bessel function should handle zero values


agoscinski:
When the ModifiedSphericalBessel is used with a zero values, one only gets in debug build an error message
```
src/src/rascal/math/bessel.cc:161: void rascal::math::ModifiedSphericalBessel::calc(double, double): Assertion `this->bessel_values.isFinite().all()' failed.
```
This should be handled somehow that the user is also informed in release build.
This can currently happen with the interpolator, when begin range is set to zero, the workaround for now would be to not set the begin range 0 but a value near 0 like 0.1

Issue #281, opened 2020-06-15
RadialScaling wrong documentation


agoscinski:
For addding the cutoff_function_parameters hyperparameters

This does not work
https://github.com/cosmo-epfl/librascal/blob/98d2933914cfd1333be75e2b8253ea7c807a1687/bindings/rascal/representations/spherical_invariants.py#L142-L144
but this does
```
dict(rate=0,
       scale=5,
       exponent=2)
```

Issue #141, opened 2019-10-02
Documented nonspecialized template function


agoscinski:
Add the commented out functions into the code using static_assert(false)
https://github.com/cosmo-epfl/librascal/blob/4f32d2a6cb8c3d053316e798c85d3942b248377a/src/representations/calculator_spherical_expansion.hh#L202-L206

Luthaf:
A bit more context here: It should be possible to document these functions by only declaring (and not **defining**) them:

```cpp
template<class T>
struct Base {
    /// Some docs
    Vector_Ref compute_center_contribution();
    /// Some more docs
    Matrix_Ref compute_neighbour_contribution();
}
```

The main issue doing so is that the compiler will no longer emit an error if the base class is passed when instantiating the template. Instead, the error will happen at link time, with some variation of `undefined symbol error: Base::compute_center_contribution`.

---

It might also be possible to explicitly forbid instantiation with `static_assert`:

```cpp
template<class T>
struct Base {
    /// Some docs
    Vector_Ref compute_center_contribution() {
        static_assert(false, "this class should not be used directly");
    }
    /// Some more docs
    Matrix_Ref compute_neighbour_contribution() {
        static_assert(false, "this class should not be used directly");
    }
}
```

Depending on the usage of the class, and when exactly static_assert is checked (I can't remember right now) this might or not be possible.

--- 

Finally, if static_assert assert does not work and we still want compile-time errors, this might be possible by using constexpr, although I am not sure it is powerful enough in C++14.

Luthaf:
The rational for doing this that it greatly improves the documentation of these functions, and what exactly is expected to be defined when using template "inheritance".

agoscinski:
Implementation seems not so simple as thought in this case. The function is apparently is used in the examples if it is implemented. If I change to
```
      //! define the contribution from the central atom to the expansion
      Vector_Ref compute_center_contribution() {
        static_assert(false, "This method should not be used directly");
        return Vector_Ref(Vector_t::Zero());
      }

```
I get the following error
```
In file included from /home/alexgo/lib/librascal/examples/json_structure.cc:30:0:
/home/alexgo/lib/librascal/src/rascal/representations/calculator_spherical_expansion.hh: In member function ‘rascal::internal::RadialContributionBase::Vector_Ref rascal::internal::RadialContributionBase::compute_center_contribution()’:
[...]

PR #289, opened 2020-09-11
Fixes for issues #141 #280 #281 #282 #285

### LINKED ISSUE #285: RadialScaling scale=0 case does not work as described in the doc
If one sets m=0 (scale = 0), the radial scaling does not change to u(r)=1. Currently, one has to set in addition c=0 (rate = 0), but the doc says
```
      u(r) = \\begin{cases}
            \\frac{1}{(r/r_0)^m} &\\text{if c=0,}\\\\
            1 &\\text{if m=0,} \\\\
            \\frac{c}{c+(r/r_0)^m} &\\text{else},
            \\end{cases}
```
Adapt code to doc.
Reminder for me 
> its quite simple, at line 232 and 245 in cutoff_functions.hh you need to check that exponent is not 0 too

agoscinski:
### This pull request solves the following issues:

There is a commit for every issue , except for #282 there are three

#141 Documented nonspecialized template function
- enforce implementation with error during runtime
- moved long documentation to function in base class and left short documentation for derived member functions, please check if this makes sense

#280 Add documentation for DVR radial basis on python side
- documentation has been added and reordered to match input parameter order

#281 Add documentation for DVR radial basis on python side
- adapted documentation as mentioned in the issue

#282 Bessel function should handle zero values
- added constant `constexpr double SPHERICAL_BESSEL_FUNCTION_FTOL = 1e-5;` which is lowest value the bessel function can be used with. A value of 1e-7 would also work, but a value of 1e-9 already returns test error.
- too low distances (below `SPHERICAL_BESSEL_FUNCTION_FTOL`) result in an error
- interpolator now uses the `SPHERICAL_BESSEL_FUNCTION_FTOL` value as range begin and the cutoff as range end, therefore the user input parameter `range_begin` and `range_end` have been removed
- several tests in test_calculator had a mismatch in interaction cutoff between manager and representation, this caused an error when used with the interpolator, now this mismatch has been fixed

#285 RadialScaling scale=0 case does not work as described in the doc:
- program logic is now closer to the logic in the documentation/paper
- added test `radial_scaling_exponent_zero_test` for  exponent is zero (m=0) test into `test_cutoff_functions.cc`

Additional stuff:
- updated clang-format requirement in README

### Tasks before review:

          elsewhere in the code.
          auto-formatting is sensible
          own branch


src/rascal/representations/cutoff_functions.hh
```diff
@@ -229,28 +230,28 @@ namespace rascal {
...
-        if (this->rate > math::DBL_FTOL) {
-          factor = this->rate / (this->rate + math::pow(distance / this->scale,
-                                                        this->exponent));
+        if (std::abs(this->rate) <= math::DBL_FTOL) {
+          factor = 1. / math::pow(distance / this->scale, this->exponent);
         } else if (this->exponent == 0) {
           factor = 1.;
         } else {
-          factor = 1. / math::pow(distance / this->scale, this->exponent);
+          factor = this->rate / (this->rate + math::pow(distance / this->scale,
```
felixmusil, line 238: There might be a small gain to have this case, the most likely, first in the if branch. 
this->rate and  this->exponent don't change after initialization so you could check for these 3 cases there and set an intermediate variable to fit any of the 3 cases with this one being first ?

src/rascal/representations/cutoff_functions.hh
```diff
@@ -229,28 +230,28 @@ namespace rascal {
...
       }
 
       double grad(double distance) {
         double factor{0.};
-        if (this->rate > math::DBL_FTOL) {
-          double ff{math::pow(distance / this->scale, this->exponent)};
-          factor = -this->rate * this->exponent * ff / distance /
-                   math::pow(this->rate + ff, 2_size_t);
+        if (std::abs(this->rate) <= math::DBL_FTOL) {
+          factor = -this->exponent / distance /
```
felixmusil, line 247: same as above

tests/test_calculator.cc
```diff
@@ -149,21 +171,25 @@ namespace rascal {
...
-            collection.get_features(representations.back());
-
-        BOOST_CHECK_EQUAL(feat_prop.rows(), feat_col.rows());
-        BOOST_CHECK_EQUAL(feat_prop.cols(), feat_col.cols());
-
-        auto diff_rep{math::relative_error(feat_prop, feat_col)};
-        BOOST_CHECK_LE(diff_rep.maxCoeff(), 6e-12);
+        double representation_cutoff{
+            extract_interaction_cutoff_from_representation_hyper(hyper)};
+        if (manager->get_cutoff() == representation_cutoff) {
```
felixmusil, line 176: perfect that's just how it should be !

src/rascal/representations/calculator_spherical_expansion.hh
```diff
@@ -1082,8 +1099,10 @@ namespace rascal {
             radial_contribution_hypers.at("optimization").template get<json>();
 
         double accuracy{this->get_interpolator_accuracy(optimization_hypers)};
-        double range_begin{this->get_range_begin(optimization_hypers)};
-        double range_end{this->get_range_end(optimization_hypers)};
+        // minimal distance such that it is still stable with the interpolated
+        // function
+        double range_begin{math::SPHERICAL_BESSEL_FUNCTION_FTOL};
+        double range_end{this->interaction_cutoff};
```
felixmusil, line 1105: The reason to keep this as a user input was to allow for a NL with `r_c > this->interaction_cutoff` so one could set `range_end` to `r_c`. At the moment it is not really used so I guess it is fine to remove it from the user inputs

felixmusil (changes requested):
Good catch on the version of the package in the requirements. It makes all the tests fail on the other branches...

I agree with most of the changes, I just feel that the bessel part could be handled without an error so could you just look into it a bit ?

src/rascal/representations/cutoff_functions.hh
agoscinski: do you really think we should try to outsmart branch prediction, without any benchmarks, with a small loss in readability?

src/rascal/representations/calculator_spherical_expansion.hh
agoscinski: could you introduce a case where this might happen in the future? I have a hard time to imagine when this would be the case

src/rascal/math/bessel.hh
```diff
@@ -38,6 +38,8 @@
 
 namespace rascal {
   namespace math {
+    // Minimal value which can be evaluated with the spherical bessel function
+    constexpr double SPHERICAL_BESSEL_FUNCTION_FTOL = 1e-7;
```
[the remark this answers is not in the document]
agoscinski: right, forgot to change it

src/rascal/representations/cutoff_functions.hh
felixmusil: when using the radial scaling this is the main use case. Even if the branching is properly predicted 2 comparison are evaluated in any cases with this order so it would not be as good as the other way around.

src/rascal/representations/calculator_spherical_expansion.hh
felixmusil: it is fine as is, I am just mentioning the reason why it was like that in the first place.

src/rascal/math/bessel.cc
```diff
@@ -129,6 +130,13 @@ void ModifiedSphericalBessel::calc(double distance, double fac_a) {
   this->bessel_arg = (2. * fac_a * distance) * this->x_v;
   this->bessel_arg_i = this->bessel_arg.inverse();
 
+  if (distance < SPHERICAL_BESSEL_FUNCTION_FTOL) {
+    this->bessel_values.setZero();
+    // 0th order approximation
+    this->bessel_values.col(0) = Eigen::exp(-this->x_v.square() * fac_a);
+    this->set_small_bessel_values_to_zero();
+  }
```
felixmusil, line 138: I am not sure about the set_small_bessel_values_to_zero, why do you need it here ? Also this approximation implies that the other terms (order > 0) are 0 so it should stop here. The gradients would be 0 too.
```suggestion
    this->set_small_bessel_values_to_zero();
    if (this->compute_gradients) {
      this->bessel_gradients.setZero();
      // this->bessel_gradients.col(0) = - 2.*fac_a*distance * this->bessel_values.col(0);
    }
    return ;
  }
```

felixmusil (approved):

Issue #288, opened 2020-09-08
Using type `X` for atoms yields nonsense SOAP vectors


rosecers:
PDF of jupyter notebook:
[issue_with_xtype.pdf](https://github.com/cosmo-epfl/librascal/files/5188756/issue_with_xtype.pdf)

Code for reference:

```python
import ase
from ase import Atoms
print("ase version:", ase.__version__)

import rascal
from rascal.representations import SphericalInvariants as SOAP
print("rascal version:", rascal.__version__)

import matplotlib
from matplotlib import pyplot as plt
print("Matplotlib version:", matplotlib.__version__)

import sys
print("Python version:", sys.version)

hypers = dict(max_radial=12, max_angular=12,
              soap_type="PowerSpectrum",
              gaussian_sigma_type="Constant",
              gaussian_sigma_constant=0.5,
              cutoff_smooth_width=0.5,
              interaction_cutoff=2.0
             )

# simple, non-primitive BCC structure
frameX = Atoms('X2',
                 positions=[[0, 0, 0], [0.5, 0.5, 0.5]],
                 cell=[1, 1, 1],
                 pbc=[1, 1, 1])

soap = SOAP(**hypers)
c = soap.transform(frameX)
rep = c.get_features_by_species(soap)

print('Species keys: ', rep.keys())

for key in rep.keys():
    plt.plot(rep[key][0], label=key)
plt.legend()
plt.show()

# simple, non-primitive BCC structure
frame = Atoms('C2',
                 positions=[[0, 0, 0], [0.5, 0.5, 0.5]],
                 cell=[1, 1, 1],
                 pbc=[1, 1, 1])

soap = SOAP(**hypers)
c = soap.transform(frame)
rep = c.get_features_by_species(soap)

print('Species keys: ', rep.keys())

for key in rep.keys():
    plt.plot(rep[key][0], label=key)
plt.legend()
plt.show()
```

rosecers:
Using type `X`
![image](https://user-images.githubusercontent.com/47536110/92486445-678b9c00-f1ec-11ea-8625-45abc24a5512.png)

Using type `C`
![image](https://user-images.githubusercontent.com/47536110/92486456-6ce8e680-f1ec-11ea-9e8f-ca0dcfb6d12f.png)

max-veit:
[...]

PR #292, opened 2020-09-16
fix the atomic number 0 issue

felixmusil:
Fix the interoperability with ASE that interpret atom symbol 'X' as atomic number 0 which was not properly handled in spherical invariants (see #288).

At the moment rascal::MaxChemElements is 130 so in principle 12 non physical atomic types are supported. I don't think it is important enough to try to make it more systematic so let's keep this as a future feature.


Luthaf (changes requested):
The fix looks good to me, I think we should also check somewhere that all species/atomic number values we get are actually below `MaxChemElements`, to prevent silent failure if someone passes a higher value.

Luthaf:
On a more fundamental point of view, I don't understand why we are creating the keys for all possible pairs below `MaxChemElements` (https://github.com/cosmo-epfl/librascal/pull/292/files#diff-9d8839ce162b641d7825d239692b7daaR460-R461) instead of taking the pairs that actually exist in the input? The current code create 8k entries in the map, while we could get away with only ~10 if the input contains 5 species, removing a large number of allocations.

src/rascal/representations/calculator_spherical_invariants.hh
```diff
@@ -426,8 +426,8 @@ namespace rascal {
 
         Key_t sparsified_type{0, 0};
         // there are 118 atomic elements atm
```
Luthaf, line 428: Is this comment outdated? It reads like it is a leftover from when the loop was `for (int sp1{1}; sp1 < 118; sp1++)`

felixmusil:
> On a more fundamental point of view, I don't understand why we are creating the keys for all possible pairs below MaxChemElements

The main reason for this is to keep the logic 'sparcified' vs 'non sparcified' in the initialization of the class: key_map maps to different values depending on this. Also the calculators are thought as unchanging entities after initialization which would break if key_map is modified afterwards. I am not saying your remark should not be followed, just that it warrants some more discussion/thoughts.

> I think we should also check somewhere that all species/atomic number values we get are actually below MaxChemElements, to prevent silent failure if someone passes a higher value

done and tested.

Luthaf (approved):

Luthaf:
Looks good to me as a fix for the original issue =).

> The main reason for this is to keep the logic 'sparcified' vs 'non sparcified' in the initialization of the class: key_map maps to different values depending on this. Also the calculators are thought as unchanging entities after initialization which would break if key_map is modified afterwards. I am not saying your remark should not be followed, just that it warrants some more discussion/thoughts.

I think this is an unfortunate side effect of having calculator store their data inside the systems instead of managing it separately. Agreed that this should be part of a larger scale discussion

rosecers (changes requested):
Looks good to me! Provided it passes CI I'll approve

felixmusil:
You can approve now then. The CI tests are blocking any merge with master.

rosecers (approved):

PR #293, opened 2020-09-16
Feat/gap pred

felixmusil:
Implement a dedicated routine to compute the force predictions in the case of the GAP model. 
In the GAP model the force predictions are given by:
```
F(X_i) = - \sum_n \alpha_n  \sum_{j\inX_i}   [z* (X_j \dot T_n)^{z-1}] T_n \dot dX_j/dr_i
```

The predictions were computed by computing the kernel gradients elements (like in the training phase) and then multiplying with the weights of the model. This routine switches the order of operation like follow:
```
F(X_i) = -  ( \sum_{j\inX_i}  ((\sum_n (\alpha_n** [z* (X_j \dot T_n)^{z-1}])_1 T_n)_2 \dot dX_j/dr_i)_3)_4
```
which improves the computational scaling of the overall operation.


bindings/rascal/models/krr.py
```diff
@@ -81,7 +81,7 @@ def predict(self, managers, compute_gradients=False):
             return Y0 + np.dot(KNM, self.weights).reshape((-1))
         else:
             rep = self.kernel._representation
-            gradients = -compute_forces(rep, self.kernel._kernel, managers.managers, self.X_train._sparse_points, self.weights.reshape((1, -1)))
+            gradients = -compute_sparse_kernel_forces(rep, self.kernel._kernel, managers.managers, self.X_train._sparse_points, self.weights.reshape((1, -1)))
```
Luthaf, line 84: why is there a minus here? Shouldn't the C++ code produce results directly with the right sign?

felixmusil: `compute_sparse_kernel_forces` gives forces while we want gradients at the model level in python to keep the logic of the model.

Luthaf: Should their be a `compute_sparse_kernel_gradient` instead then?

felixmusil: I don't think it is needed

max-veit: I think the point is that it seems a bit odd to flip the sign of these quantities _twice_ -- once (implicitly) when computing forces, since those are the negative gradients of the atomic energies, and once here to get _back_ to gradients again (and a third time when whatever MD program we're using requests forces, not gradients!). So I think we should be more consistent with either sticking with forces or with gradients, so that we don't get confused by too many sign flips. My vote is to stick with forces.

ceriottm:
I would actually be inclined to stick with gradients. Energy/forces are
obviously super-important, but we are building a model of a target
quantity, and it makes sense to use the most general and abstract form.

bindings/rascal/models/krr.py
```diff
@@ -74,11 +75,16 @@ def predict(self, managers, compute_gradients=False):
         np.array
             predictions
         """
-        KNM, Y0 = self._preprocess_input(managers, compute_gradients)
+
         if compute_gradients is False:
+            KNM, Y0 = self._preprocess_input(managers, compute_gradients)
```
max-veit, line 80: Now that the kernel gradient predictions are moved into a separate function, wouldn't it be a good idea to do the same with forces? I'm a bit confused as to why this function returns _either_ energies _or_ forces, and not both (when you want the forces, you usually also want the energies, right?). It seems a better idea to split this up into two functions, one that computes energies and one that computes forces (gradients) rather than switching between two essentially separate functions in here.

src/rascal/models/sparse_kernel_predict.hh
```diff
@@ -0,0 +1,256 @@
...
+namespace rascal {
+
+  /**
+   * Compute the partial gradients of a structure w.r.t atomic positions
+   * using the GAP model (see
+   * @ref SparseKernelImpl<internal::SparseKernelType::GAP>::compute_derivative)
+   *
+   * The point of this function is to provide a faster prediction routine
+   * compared to computing the kernel element and then multiplying them with
+   * the weights of the model.
```
max-veit, line 47: Perhaps mention here why the latter method is slower? How does it scale compared to this function?

src/rascal/models/sparse_kernel_predict.hh
```diff
@@ -0,0 +1,256 @@
...
+    std::set<int> unique_species{};
+    for (auto center : manager) {
+      unique_species.insert(center.get_atom_type());
+    }
+
+    // find shared central atom species
+    std::set<int> species_intersect{
+        internal::set_intersection(unique_species, sparse_points.species())};
+
+    // find offsets alongs the sparse points direction
```
max-veit, line 102: Isn't this (finding offsets) something that should be handled by PropertyBlockSparse?

src/rascal/models/sparse_kernel_predict.hh
```diff
@@ -0,0 +1,256 @@
...
+          djfi[neigh] =
+              sparse_point_scaled.dot_derivative(a_sp, prop_grad[neigh]);
+        }
+      }
+    }  // center
+  }
+
+  /**
+   * Compute the gradients of a structure w.r.t atomic positions
+   * using a sparse GPR model. Only GAP model is implemented at the moment (see
```
max-veit, line 181: I'm confused. What's the difference between a "generic" sparse GPR model and GAP? I mean, GAP is literally just sparse GPR applied to energies and forces.

src/rascal/models/sparse_kernel_predict.hh
```diff
@@ -0,0 +1,256 @@
...
+    std::map<int, int> offsets{sparse_points.get_offsets()};
+    Keys_t rep_keys{prop_grad.get_keys()};
+    std::map<int, Keys_t> keys_intersect{};
+    for (const int & sp : species_intersect) {
+      keys_intersect[sp] =
+          internal::set_intersection(rep_keys, sparse_points.keys_sp.at(sp));
+    }
+
+    // attach partial gradients array to manager
+    auto && djfi{
```
max-veit, line 112: Is this variable referring to the derivative of a force...? It looks like "d/d x_j F_i"...

src/rascal/models/sparse_kernel_predict.hh
```diff
@@ -0,0 +1,256 @@
...
+
+    if (species_intersect.empty()) {
+      return;
+    }
+
+    math::Vector_t weights_scaled(weights.size());
+    size_t i_row{0};
+    for (auto center : manager) {
+      const int a_sp{center.get_atom_type()};
+      // \alpha_n^{scaled} = \alpha_n * [z* (X_j \dot T_n)^{z-1}]
```
max-veit, line 132: In words: This is what comes out of the derivative of the kernel when the chain rule is applied.

bindings/rascal/models/krr.py
```diff
@@ -74,11 +75,16 @@ def predict(self, managers, compute_gradients=False):
...
             return Y0 + np.dot(KNM, self.weights).reshape((-1))
         else:
-            return np.dot(KNM, self.weights).reshape((-1, 3))
+            rep = self.kernel._representation
+            gradients = compute_sparse_kernel_gradients(rep, self.kernel._kernel, managers.managers, self.X_train._sparse_points, self.weights.reshape((1, -1)))
```
max-veit, line 84: Long long line (how did this get past the linter? I had to scroll in the web interface to see the whole thing)

max-veit (commented):
Ok, overall easy enough to understand, but still with some unnecessary complexity / overgeneralization creeping in at points.

Also, is the old gradients prediction routine still available from the Python bindings? If so, shouldn't it be clearly marked as deprecated because it's slower than this implementation?

src/rascal/models/sparse_kernel_predict.hh
Luthaf: My understanding is that `PropertyBlockSparse` deals with sparse features; while this is dealing with sparse environments. I might be wrong though =)

src/rascal/models/sparse_kernel_predict.hh
felixmusil: a GAP model is a very specific flavor of sparse GPR: Subset of regressor with a cosine kernel function that does not mix different central atomic types.

src/rascal/models/sparse_kernel_predict.hh
felixmusil: the comment tells what it is

src/rascal/models/sparse_kernel_predict.hh
felixmusil: there is the models weights so not quite I would say.

bindings/rascal/models/krr.py
felixmusil: there is not python linter at the moment. The black PR is still in limbo.

src/rascal/models/sparse_kernel_predict.hh
felixmusil: removed because unused

bindings/rascal/models/krr.py
felixmusil: I don't see the difference between this way and having separate routines. I would rather not make such change now as it would require many modifications of the examples and tests that are quite unrelated to the PR.

max-veit: The difference is: Two separate, well-documented functions, each doing one thing, returning a single array of predictable shape, versus this, which is just two functions rolled into one and returns a different thing depending on a boolean switch in the input. I know there are functions especially in the NumPy library that do this sort of thing (return different things depending on input switches), but (1) they usually only do this for returning _extra_ information besides the main return array, and (2) it's unpredictable and confusing for new users.

As for the separate examples and tests: They're not "unrelated" to this PR if they have to do with property (and gradient) prediction. And you can always make this into a "compatibility" function that just delegates to two functions that do the work, then mark this as deprecated.

src/rascal/models/sparse_kernel_predict.hh
max-veit: No it's not. GAP doesn't specify the cosine kernel; if you look in the QUIP source you'll see several other kernel types implemented (dot-product, aka "cosine", along with squared-exponential, bond real-space, and a few other things). All define a GAP model, because the GAP just refers to (sparse) Gaussian process regression applied to potential energy surfaces. Nothing more, nothing less.

src/rascal/models/sparse_kernel_predict.hh
max-veit: Trust me, it doesn't. And if you need a comment to tell you what a variable is for, then it's a bad variable name.

src/rascal/models/sparse_kernel_predict.hh
max-veit: ...so model weights times chain-rule detritus? It's much easier to understand conceptually what's being done here when you write it in words (it's easier than you think) rather than equations, which anyway kinda just repeat the line below.

bindings/rascal/models/krr.py
max-veit: Hmm. Well, if there's no objections, I'll just go ahead and approve that then?

src/rascal/models/sparse_kernel_predict.hh
felixmusil: then what would you recommend ?

src/rascal/models/sparse_kernel_predict.hh
felixmusil: I personally prefer the equation as it tells me which operation is being done precisely while a more conceptual comment is not self sufficient to understand the following line. The relation between the code and the equation is far from direct to me.

src/rascal/models/sparse_kernel_predict.hh
felixmusil: you are right but in the case of GAP with SOAP I don't think any of the other possibilities that you mention have been used in an article, am I wrong ? So I have been using 'GAP model' as a short hand for 'sparse GPR additive energy model with SOAP and cosine kernel that does not mix different central atomic types'. Do you have a better and yet short alternative ?

bindings/rascal/models/krr.py
felixmusil: this issue will be dealt with in #300.

src/rascal/models/sparse_kernel_predict.hh
felixmusil: also `djfi` refers to `d/dx_j f_i` which is the derivative of function `f` centered on atom `i` with respect to the position of atom `j`

src/rascal/models/sparse_kernel_predict.hh
max-veit: SOAP-GAP is what I've been calling it.

src/rascal/models/sparse_kernel_predict.hh
max-veit: How about something with "partial" or "pairwise force" in the name? Like `pair_force_atom_i_xj` or something slightly less verbose? Just `f` for "function" is too easy to confuse with forces, especially in this context.

src/rascal/models/sparse_kernel_predict.hh
max-veit: And that's precisely why you need both a plain-text explanation (for when we come back to this in six months and don't remember what the symbols mean) and the equation, which doesn't necessarily have to be in an inline comment. We could, for instance, put the different steps in a LaTeX document in the developer documentation, like we've done with the spherical expansion derivation (I think this needs updating btw...) -- it's just that inline comments are horrible to read for those of us without a LaTeX compiler in our heads.

ceriottm:
How about you just say that only a polynomial kernel is implemented ATM
without getting too much bogged down with branding?

max-veit:
> How about you just say that only a polynomial kernel is implemented ATM without getting too much bogged down with branding?

Branding...? This is more about reducing confusion for those who are used to working with GAP models, especially those with e.g. 2b an 3b baselines that use a squared-exponential rather than a dot-product kernel. In any case, we'll really only need to make the distinction once we get a sq-exp kernel implemented on the same footing as the current "SOAP-GAP" one.

max-veit (approved):
In any case, looks good to me (with the rest to be addressed in related issues).

PR #296, opened 2020-09-28
Cleaning up of the SphericalCovariant parameters in preparation for review

ceriottm:
We are using SphericalCovariants in a couple of reviews. In the process some quirks became clear that would be nice to be fixed before the review is finished. 

* rename "lam" to "covariant_lambda" to have homogeneous level of verbosity (cf. "max_angular" etc)
* change (for the moment on the python side) the order of the axes of the covariant features so that
it is more logical to compute covariant kernels
* miscellaneous cleaning ups that emerged in the process of recreating the references for tests


src/rascal/representations/calculator_spherical_covariants.hh
```diff
@@ -105,12 +105,15 @@ namespace rascal {
...
-                int m3s{static_cast<int>(m3 - l3)};
+
+          // NB: the order of this loop MATTERS because it mirrors the
+          // order of evaluation of the covariants further down
+          for (size_t m3{0}; m3 < 2 * l3 + 1; m3++) {
+            int m3s{static_cast<int>(m3 - l3)};
+            for (size_t m1{0}; m1 < 2 * l1 + 1; m1++) {
+              int m1s{static_cast<int>(m1 - l1)};
+              for (size_t m2{0}; m2 < 2 * l2 + 1; m2++) {
+                int m2s{static_cast<int>(m2 - l2)};
```
felixmusil, line 116: do you mean to say that m2 is the fast index here while m3 is slower ?

felixmusil (changes requested):
seems good to merge after resolving conflicts

ceriottm:
It must be implemented this way - or at least, it must be implemented in
the same order as the loop that computes the actual lambda-SOAP further
down.
The way it traverses the list of 3J coefficients must be the same. It's a
mystery how it was giving covariant results, before.

ceriottm:
I'm closing this PR because it got stale, and merging will be a nightmare. I will open a new PR in which I cherry-picked the relevant commits.

PR #297, opened 2020-10-05
radial integral gto storage mismatch

felixmusil:
For some reason the storage in the GTO radial integral was col major while the rest was working with row major. It had a significant overhead when working with the spline interpolator.


Luthaf:
Very nice find! Do you have some performance number?

It is really baffling that changing `Matrix_t` to `Eigen::Map<Matrix_t>` in a function return type can have such a big performance impact =/ I understand what is going on here (implicit construction of a `Matrix_t` with colum-major storage by default from a row major stored value), but this really feel like an easy performance trap from Eigen. 

Correct me if I'm wrong, but there is both a copy of the full matrix; and a change of the storage order going on, both with performance implications. I don't have a better solution here short of rolling our own matrix code =|

(side note, I've found [arrayfire](https://arrayfire.com/) when looking around for alternatives, which has the benefit of automagic GPU support. Might be nice to have a deeper look)

felixmusil:
> It is really baffling that changing Matrix_t to Eigen::Map<Matrix_t> in a function return type can have such a big performance impact =/ I understand what is going on here (implicit construction of a Matrix_t with colum-major storage by default from a row major stored value), but this really feel like an easy performance trap from Eigen.

The main problem was coming from the order mismatch. Turns out that the copy has to happen since without it the data is not held anywhere...

> Do you have some performance number?

~ a factor of 2 in a particular case and for splined GTO.

performance/profiles/profile_radial_integral.cc
```diff
@@ -0,0 +1,313 @@
...
+      for (auto neigh : center.pairs()) {
+        const double & dist{manager->get_distance(neigh)};
+        coefs_n = radial_integral->template compute_neighbour_contribution(
+            dist, neigh);
+      }
+    }
+  }
+};
+
+int main(int argc, char * argv[]) {
```
Luthaf, line 252: I don't see the point including this file (and corresponding JSON input) in the repository, who would use it? 

It could be nice to have this transformed into a benchmark though, so that we can make sure there are no performance regression. This would require some extra work though, so I'm fine to leave it out.

Luthaf (commented):
The fix looks good with me, but I don't see the point to include the "profile" file

performance/profiles/profile_radial_integral.cc
felixmusil: That's how I found out about this bug.

Luthaf: That's a quite low probability of being useful again for 350+ new lines of code that we would have to make sure keep compiling as we change the rest of the code

Luthaf:
Thanks! Could you also squash your commits here? Else I can do it when merging this PR =)

Luthaf (approved):

Luthaf:
> Could you also squash your commits here? Else I can do it when merging this PR =)

There is a "squash and merge" option on the github UI to prevent polluting git history with random commits, it would have been nice to use it, or squash manually.

agoscinski:
yes it would have been nice, "ff" as commit message is ridiculous uninformative

max-veit:
Especially as it's now the most recent commit that everyone sees when they go on the project page. Nice.

felixmusil:
There are a few PR to merge so it won't stay there for long.

felixmusil:
Also note that squash merge has been disabled for the librascal repo.

Luthaf:
In that case we have to work with manual squashing & force-push

max-veit:
Might be a good idea to get in the habit of rebasing/squashing a series of commits before pushing them up to the repo in the first place? I don't know what VS Code's approach to this is -- if it just auto-pushes commits as soon as you add them, then that's very bad behaviour and should be turned off.

PR #302, opened 2020-10-16
Improve documentation for global_species

Luthaf:
I tried to use it as `global_species = ["C", "H", "N"]` and got a cryptic error message, so let's specify that this needs to be the atomic number.


felixmusil (approved):
thanks for the clarification

PR #304, opened 2020-11-18
Feat/fix property order 0

felixmusil:
Fix the construction of Property<Order==0> and add a dedicated access routine. Some other functionalities are suppressed with SFINAE because they don't make sense in this context.


agoscinski (approved):
Looks good now. Just for a description of this PR here the added doc what order zero property is.
```
   * Special cases exist for Order=0 properties which are related to the
   * structure. Unlike higher orders, it does not support multiple
   * layers, therefore its layer is set to 0. The underlying data can be
   * directly accessed with operator() for individual elements, operator[0]
   * to get a Map<Matrix> and .view() to get a Map<const Matrix>.
   */

PR #307, opened 2020-12-11
i-PI integration

max-veit:
Add support for i-PI using a dedicated calculator, which is also usable as a "generic" MD interface for any other MD code. Works with the branch https://github.com/cosmo-epfl/i-pi/tree/feat/librascal

The new calculator is implemented in `bindings/rascal/models/genericmd.py`.

Still to be worked out:
- ~Confirm sign convention of virial~ Confirmed experimentally
- ~Make units handling more robust (fewer assumptions)~ moved to i-PI
- ~Does i-PI run non-periodic simulations? If so, how is this signalled to the force driver?~ No

Tasks before review:

          elsewhere in the code.
          auto-formatting is sensible
          own branch

Co-written with Lorenzo (@lgigli)


bindings/rascal/models/krr.py
```diff
@@ -1,9 +1,14 @@
-from ..utils import BaseIO
+from ..utils import BaseIO, is_notebook
 from ..lib import compute_sparse_kernel_gradients, compute_sparse_kernel_neg_stress
 
 import numpy as np
 import ase
 
+if is_notebook():
```
max-veit, line 7: Just a note, this includes `tqdm` as a strict dependency (as in, the file cannot be imported without it). Is this something we want to do? I'd rather make it optional, to be honest.

bindings/rascal/models/krr.py
```diff
@@ -199,11 +204,43 @@ def get_representation_calculator(self):
...
+    feat = representation.transform([frame])
+    en_row = kernel(feat, X_sparse)
+    grad_rows = kernel(feat, X_sparse, grad=(True, False))
+    return en_row, grad_rows
+
+
+def compute_KNM(frames, X_sparse, kernel, soap):
+    Nstructures, Ngrads, Ngrad_stride = get_strides(frames)
+    KNM = np.zeros((Nstructures + Ngrads, X_sparse.size()))
+    pbar = tqdm(frames, desc="compute KNM", leave=False)
```
max-veit, line 229: In fact, I'm not sure how I feel about including `tqdm` at such a deep level -- it seems wrong for the code to need to know about something that's in principle unrelated, but I'm not sure I see a better solution.

One idea would instead be to wrap `frames` in a `tqdm` _before_ passing it to this function in whatever notebook uses these - will look into this later.

bindings/rascal/models/krr.py
```diff
@@ -199,11 +207,43 @@ def get_representation_calculator(self):
...
+
+
+def compute(i_frame, frame, representation, X_sparse, kernel):
+    feat = representation.transform([frame])
+    en_row = kernel(feat, X_sparse)
+    grad_rows = kernel(feat, X_sparse, grad=(True, False))
+    return en_row, grad_rows
+
+
+def compute_KNM(frames, X_sparse, kernel, soap):
```
max-veit, line 229: needs doc

bindings/rascal/models/krr.py
```diff
@@ -199,11 +207,43 @@ def get_representation_calculator(self):
...
+    Ngrads = 0
+    for frame in frames:
+        n_at = len(frame)
+        Ngrad_stride.append(n_at * 3)
+        Ngrads += n_at * 3
+    Ngrad_stride = np.cumsum(Ngrad_stride) + Nstructures
+    return Nstructures, Ngrads, Ngrad_stride
+
+
+def compute(i_frame, frame, representation, X_sparse, kernel):
```
max-veit, line 222: compute what?? (needs doc and tests)

bindings/rascal/models/krr.py
max-veit: and tests...

Luthaf:
A few high level comments before I dive into this for a review

- could you fix conflicts with master?
- do we really need an additional 11Mb file `reference_data/tests_only/simple_gap_fit_params.json` just for the tests? Could this be made much smaller by using less sparse points when training? We don't need an accurate potential just to check that the machinery around it works. Related to this, could you then remove the file completely from history by rebasing this branch or squashing the corresponding commits?
- where is the format for `reference_data/tests_only/simple_gap_fit_params.json` defined and documented? My current understanding is that this file contains the raw JSON serialization of all C++ classes involved in the model, plus the weights and sparse points, am I right? I'm afraid to start distributing / encourage users to generate and distribute models with an under-defined and un-documented file format. Do we want to guarantee support for this format? Or should people re-train their models whenever we change librascal internals? Overall I feel that the format we want to use to store trained models warrant a larger discussion than on the side of semi-related PR.

max-veit:
>     * could you fix conflicts with master?

I'll do that when I rebase this branch. The (tiny) conflicts here were caused by a rebase/squash of yesterday's bugfix branch (that was extracted from this one).

>     * do we really need an additional 11Mb file `reference_data/tests_only/simple_gap_fit_params.json` just for the tests? Could this be made much smaller by using less sparse points when training? We don't need an accurate potential just to check that the machinery around it works. Related to this, could you then remove the file completely from history by rebasing this branch or squashing the corresponding commits?

Didn't realize it was that large... I was already using what felt like a very small number of sparse points, but I'll try to cut it down further.

>     * where is the format for `reference_data/tests_only/simple_gap_fit_params.json` defined and documented? My current understanding is that this file contains the raw JSON serialization of all C++ classes involved in the model, plus the weights and sparse points, am I right? I'm afraid to start distributing / encourage users to generate and distribute models with an under-defined and un-documented file format. Do we want to guarantee support for this format? Or should people re-train their models whenever we change librascal internals? Overall I feel that the format we want to use to store trained models warrant a larger discussion than on the side of semi-related PR.

That's really a discussion for #305 (which is going to be my main "project" after this PR; it was also the branch used to generate this model file). To summarize for now, the model JSON is the serialization of a Python `KRR` object, which can be restored (via the librascal deserialization mechanism) and used to evaluate the model via the public `KRR` interface.
It does contain the serialized representation and kernel parameters, as well as the weights and sparse points -- but as to the concern about model files changing every time we change something in the internals of librascal, that seems unlikely as the serialization happens on a relatively high level and only uses the Python interface.

And anyway, the file lives in `reference_data/*tests_only*`, not `examples/`, so I hope that would be enough of an indication for users not to take this as a definitive model specification. The only point to having this model file right now is to be able to test the i-PI interface.

bindings/rascal/models/IP_generic_md.py
```diff
@@ -0,0 +1,103 @@
+import ase.io
```
Luthaf, line 1: This file name is a bit weird. What does IP mean here? Interatomic Potential? How about moving this to `bindings/rascal/models/md.py`? This already implies an interatomic potential IMO.

bindings/rascal/models/IP_generic_md.py
```diff
@@ -0,0 +1,103 @@
...
+import ase.units
```
Luthaf, line 2: I'm not fond of the coupling with ASE here, but I understand we are already using it for structures loading. In order to reduce such coupling, could we use a different solution for unit management?

bindings/rascal/models/IP_generic_md.py
```diff
@@ -0,0 +1,103 @@
...
+from ..neighbourlist.structure_manager import AtomsList, unpack_ase
+
+
+class GenericMDCalculator(BaseIO):
+
+    """Generic MD driver for a librascal model
+
+    Initialize with model JSON and a structure template, and calculate
+    energies and forces based on position/cell updates _assuming the
+    order and identity of atoms does not change_.
```
Luthaf, line 15: Is this assumption checked? Checking that we have a constant number of atoms should be cheap enough, and even the atomic identities should be fast to check.

bindings/rascal/models/IP_generic_md.py
```diff
@@ -0,0 +1,103 @@
...
+        """Initialize a model and structure template
+
+        Parameters
+        ----------
+        model_json  Filename for a JSON file defining the potential
+        structure_template
+                    Filename for an ASE-compatible Atoms object, used
+                    only to initialize atom types and numbers
+        assume_pbc  Force the PBC to True (to avoid subtle issues that
+                    sometimes arise when no cell is defined), default True
```
Luthaf, line 28: This feels a lot like a hack. What kind of issue arise when no cell is defined? And could we work around these issues automatically?

bindings/rascal/models/IP_generic_md.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def __init__(self, model_json, structure_template, assume_pbc=True):
+        """Initialize a model and structure template
+
+        Parameters
+        ----------
+        model_json  Filename for a JSON file defining the potential
+        structure_template
+                    Filename for an ASE-compatible Atoms object, used
+                    only to initialize atom types and numbers
```
Luthaf, line 26: How about taking a `species` array/list of integer here instead? This would remove the need to write a file just to communicate atomic species.

bindings/rascal/models/IP_generic_md.py
```diff
@@ -0,0 +1,103 @@
...
+        stress_voigt = self.model.predict_stress(self.manager)
+        stress_matrix = np.zeros((3, 3))
+        stress_matrix[tuple(zip(*self.matrix_indices_in_voigt_notation))] = stress_voigt
+        # Symmetrize the stress matrix (replicate upper-diagonal entries)
+        stress_matrix += np.triu(stress_matrix).T
+        return energy, forces, stress_matrix
+
+    def _get_init_params(self):
+        init_params = dict(
+            model_json=self.model_filename, structure_template=self.template_filename
```
Luthaf, line 94: I'm not fond of introducing a filesystem dependency in the serialization of `GenericMDCalculator`, but I also understand not wanting to duplicate all data from the model in this class serialization.

Why does this class needs to be serializable in the first place?

bindings/rascal/models/krr.py
```diff
@@ -199,11 +232,78 @@ def get_representation_calculator(self):
         return self.kernel._rep
 
 
+# TODO(max, felix) I think this belongs in utils; it's not KRR-specific
```
Luthaf, line 235: Agreed! Could you make sure it is part of the sphinx documentation as well?

bindings/rascal/models/krr.py
```diff
@@ -199,11 +232,78 @@ def get_representation_calculator(self):
...
+    en_row = kernel(feat, X_sparse)
+    grad_rows = kernel(feat, X_sparse, grad=(True, False))
+    return en_row, grad_rows
+
+
+def compute_KNM(frames, X_sparse, kernel, soap):
+    """Compute kernel of the (new) structure against the sparse points
+
+    Parameters:
+        frame       New structure to compute kernel for
```
Luthaf, line 280: the parameter is `frames` with an `s`. It looks like this function supports passing a list of frames, but the documentation disagree =)

bindings/rascal/models/krr.py
Luthaf: > will look into this later.

Did you try this? I would find it much more elegant, and it gives the user more control over the tqdm display name/positions/style etc.

bindings/rascal/models/krr.py
```diff
@@ -199,11 +232,78 @@ def get_representation_calculator(self):
...
+    return KNM
+
+
 def train_gap_model(
     kernel,
-    managers,
+    frames,
     KNM_,
-    X_pseudo,
+    X_sparse,
```
Luthaf, line 306: Mostly unrelated to this PR, but I don't really like the sparse terminology: this is not really sparse, but rather low-rank. Even if the historical literature on GAP is using sparse doesn't mean we have to use it as well =D. 

What do you feel about changing the name to **pseudo points** or **support points** all over librascal?

examples/iPi/zundel/run.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# remove the driver file if it already exists
+rm /tmp/ipi_zundel
+# make sure rascal can be imported if not installed
+export PYTHONPATH="../../../build_b/:$PYTHONPATH"
```
Luthaf, line 6: This will not run as-is on most people computers, since it is refering to files outside of the librascal repository. Could you replace/improve this file with a README explaining all that is required to run this example?

Luthaf: Or even just replace this line and the  RASCAL_DRIVER  below with `PYTHONPATH="<path/to/rascal/install>:$PYTHONPATH"` and in the README tell people they need to edit this file before running anything.

examples/iPi/zundel/zundel_IP.ipynb
```diff
@@ -0,0 +1,777 @@
...
+    "$$\n",
+    "\n",
+    "where $E_0$ represents an energy baseline (given usually by the sum of atomic self-contributions), $\\bf{d_i}$ ($i = 1, \\dots, N_{atoms}$) are the set of (normalized SOAP) descriptors describing an environment centred around atom $i$, $\\bf{d_s}$ the set of descriptors corresponding to the environments of the sparse set (of size $M$) and K a kernel matrix that describes the similarity of two different atomic environments. In our application, the kernel is just the dot product, raised to some integer power $\\zeta$: \n",
+    "\n",
+    "$$\n",
+    "        K(\\bf{d_i} , \\bf{d_s} ) \\propto \\left| \\bf{d_i} \\cdot \\bf{d_s} \\right|^{\\zeta}           \n",
+    "$$\n",
+    "\n",
+    "Finally $\\alpha_s$ represents the weights of each sparse environment, to be determined using Kernel-Ridge Regression (KRR).\n",
+    "For extensive details on the SOAP GAP-model fitting procedure and interesting physical applications, we invite the reader to refer to...(link to the chemrev Review, once available). Details on the implementation in Librascal instead are given in https://cosmo-epfl.github.io/librascal/SOAP.html."
```
Luthaf, line 30: should the librascal link be set to the paper now?

examples/iPi/zundel/zundel_IP.ipynb
```diff
@@ -0,0 +1,777 @@
...
+   "metadata": {},
+   "source": [
+    "## Getting started "
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In order to be able to fit a potential with Librascal (the model evaluator) and interface it with i-Pi (the MD engine) we first need to have both softwares available and correctly installed. To enable communication between them, two specific branches have been designed: \n",
```
Luthaf, line 44: The following lines only refer to one branch (in i-Pi), not two.

examples/iPi/zundel/zundel_IP.ipynb
```diff
@@ -0,0 +1,777 @@
...
+    "```\n",
+    "pip install git+https://github.com/cosmo-epfl/i-pi.git@feat/librascal\n",
+    "```\n",
+    "Make sure that you git pull these branches after cloning the two github repositories.\n",
```
Luthaf, line 48: This sentence is confusing here, the pip install above should do the trick

examples/iPi/zundel/zundel_IP.ipynb
```diff
@@ -0,0 +1,777 @@
...
+    "\n",
+    "Next step is installing Librascal using the instructions of the top-level README which could be:\n",
+    "```\n",
+    "pip install .\n",
+    "```\n"
```
Luthaf, line 53:
```suggestion
    "Next step is installing librascal:\n",
    "```\n",
    "pip install git+https://github.com/cosmo-epfl/librascal\n",
    "```\n"
```

examples/iPi/zundel/zundel_IP.ipynb
```diff
@@ -0,0 +1,777 @@
...
+    "### Importing all the necessary Librascal modules"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Let us start from the import of all the necessary modules. This includes the basic modules of Librascal to compute: \n",
+    "\n",
+    "1) the SOAP descriptors of a structure, by means the $\\bf{SphericalInvariants}$ class;\n",
```
Luthaf, line 69: The styling is really weird here, why not use fixed-width font instead?
```suggestion
    "1) the SOAP descriptors of a structure, by means the `SphericalInvariants` class;\n",
```

examples/iPi/zundel/zundel_IP.ipynb
```diff
@@ -0,0 +1,777 @@
...
+    "\n",
+    "2) the kernels between a set of environments and the sparse set (the $\\bf{Kernel}$ class);\n",
```
Luthaf, line 71:
```suggestion
    "2) the kernels between a set of environments and the sparse set (the `Kernel` class);\n",
```

examples/iPi/zundel/zundel_IP.ipynb
```diff
@@ -0,0 +1,777 @@
...
+    "\n",
+    "3) the GAP model itsself, which is saved as an object of the $\\bf{KRR}$ class. The predict method of the same class will allow us to give predictions to new unseen structures. \n",
```
Luthaf, line 73:
```suggestion
    "3) the GAP model itself, which is saved as an object of the `KRR` class. The predict method of the same class will allow us to give predictions to new unseen structures. \n",
```

examples/iPi/zundel/zundel_IP.ipynb
```diff
@@ -0,0 +1,777 @@
...
+    "from rascal.representations import SphericalInvariants\n",
+    "from rascal.neighbourlist import AtomsList\n",
+    "from rascal.utils import from_dict, to_dict, CURFilter, dump_obj, load_obj"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In the snippets below we extract the relevant properties for each ASE frame. We load the total potential energies and we use the $\\bf{ASE.Atoms.arrays}$ methods to get the atomic forces. For more information on how to use ASE-related methods, check the documentation at https://wiki.fysik.dtu.dk/ase/ase/atoms.html.  "
```
Luthaf, line 119:
```suggestion
    "In the snippets below we extract the relevant properties for each ASE frame. We load the total potential energies and we use the `ASE.Atoms.arrays` dictionary to get the atomic forces. For more information on how to use ASE-related methods, check the documentation at https://wiki.fysik.dtu.dk/ase/ase/atoms.html.  "
```

examples/iPi/zundel/zundel_IP.ipynb
```diff
@@ -0,0 +1,777 @@
...
+   },
+   "outputs": [],
+   "source": [
+    "# Number of structures to train the model with\n",
+    "n = 800\n",
+    "\n",
+    "global_species = []\n",
+    "for frame in frames:\n",
+    "    global_species.extend(frame.get_atomic_numbers())\n",
+    "global_species = np.unique(global_species)\n",
```
Luthaf, line 197: nitpick: this can get quite inefficient, why not use a set?
```suggestion
    "global_species = set()\n",
    "for frame in frames:\n",
    "    global_species.add(frame.get_atomic_numbers())\n",
    "global_species = np.array(global_species)\n",
```

examples/iPi/zundel/zundel_IP.ipynb
```diff
@@ -0,0 +1,777 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Now we proceed with the actual calculation of the SOAP vectors of our training set. We need to specify an hyperparameters dictionary, which Librascal uses to compute the structural features. The meaning of each single command and how to correctly set them is reported in https://cosmo-epfl.github.io/librascal/reference/python.html?highlight=soap#rascal.representations.SphericalInvariants. These hyperparameters can be used as default values, but a careful optimization of the interaction cutoff might be required in the case the material under investigation might present some mid- or long-range order. \n",
+    "\n",
+    "For the actual calculation of the SOAP features, we first create an object of the SphericalInvariants class, defined by its hyperparameters.The methods that we then need to use are $\\textbf{transform()}$, which yields a second object called the $\\textbf{manager}$ containing the representation, while $\\textbf{get_features()}$ converts it into an $NxM$ matrix, $N$ being the number of atomic environments in the training set and M the number of features per each environment. \n",
```
Luthaf, line 237:
```suggestion
    "For the actual calculation of the SOAP features, we first create an object of the SphericalInvariants class, defined by its hyperparameters.The methods that we then need to use are `transform()`, which yields a second object called the `manager` containing the representation, while `get_features()` converts it into an $NxM$ matrix, $N$ being the number of atomic environments in the training set and M the number of features per each environment. \n",
```

examples/iPi/zundel/zundel_IP.ipynb
```diff
@@ -0,0 +1,777 @@
...
+    "\n",
+    "It is however important to stress that at this point the '$\\textbf{compute_gradients}$' hyper should be set to False. We will use in fact the SOAP representation of the training set structures to compute only the sparse set using the CUR decomposition. Setting it to 'True' would require Librascal to compute all the gradients of the SOAP representation w.r.t. atomic coordinates, thus making the structural managers unnecessarily memory expensive. We will instead compute later on the gradients of the sparse kernels using the $\\textbf{compute_KNM}$ method to fit the GAP model.\n",
```
Luthaf, line 239:
```suggestion
    "It is however important to stress that at this point the '$\\textbf{compute_gradients}$' hyper should be set to False. We will use in fact the SOAP representation of the training set structures to compute only the sparse set using the CUR decomposition. Setting it to 'True' would require Librascal to compute all the gradients of the SOAP representation w.r.t. atomic coordinates, thus making the structural managers unnecessarily memory expensive. We will instead compute later on the gradients of the sparse kernels using the `compute_KNM` method to fit the GAP model.\n",
```

examples/iPi/zundel/zundel_IP.ipynb
```diff
@@ -0,0 +1,777 @@
...
+    "\n",
+    "NOTE: the Librascal structure manager only works if the atoms have been preliminary wrapped within the cell provided in the input file. It might however give issues whenever atoms are too close to the cell boundary, so in the case you want to model a system with periodic boundary conditions, it is common practice to displace the atoms by some non-zero vector and then wrap them back using the $\\textbf{ase.wrap()}$ function."
```
Luthaf, line 241:
```suggestion
    "NOTE: the Librascal structure manager only works if the atoms have been preliminary wrapped within the cell provided in the input file. It might however give issues whenever atoms are too close to the cell boundary, so in the case you want to model a system with periodic boundary conditions, it is common practice to displace the atoms by some non-zero vector and then wrap them back using the `ase.wrap()` function."
```

Luthaf: There are a bunch more of these in the document, and they look really weird especially when function names have a `_` inside.

Luthaf:
The updated example model file is much better, and I'm fine with leaving the discussion of the serialization format for models for a later PR!

README.rst
```diff
@@ -92,11 +92,11 @@ The following packages are required for building some optional features:
 +==================+=============+====================+
 | Documentation    | pandoc      | (latest)           |
 +------------------+-------------+--------------------+
-|                  | sphinx      | 2.1.2              |
+|                  | sphinx      | 2.1.2 or later     |
```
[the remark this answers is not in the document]
max-veit: We could do, want to open an issue about it?

bindings/rascal/models/IP_generic_md.py
max-veit: It was initially for consistency with `IP_ase_interface.py`, but I wouldn't mind a shorter module name.

bindings/rascal/models/IP_generic_md.py
max-veit: Units are already handled by i-PI, so this import is now unused. I'll take care of it with the rest of the changes.

bindings/rascal/models/IP_generic_md.py
max-veit: No, but almost any MD code operates under this same assumption (apart from some edge cases in LAMMPS supporting particle flow / GCMC simulations, which i-PI doesn't do AFAIK). Checking the number of atoms will be fast but note that we don't get identities from a code like i-PI (see the signature of the `calculate()` interface).

bindings/rascal/models/IP_generic_md.py
max-veit: This is the automatic workaround. The problem is that ASE sets the cell to be non-periodic if you read a file with no (or unreadable) cell information, _even if you explicitly set a cell later_, so you have to remember to set PBC explicitly as well. I've lost weeks of work because of this misunderstanding.

Although probably the issue will be partially addressed in #323 ; let's discuss it more there.

bindings/rascal/models/IP_generic_md.py
max-veit: That would be more of a pain for the user, actually. Many MD codes (including i-PI, and also LAMMPS) take a structure file for initialization of system parameters like the number and type of atoms, the topology, and sometimes other things. It's easiest to just pass the same structure file you used for initialization and let the code extract the information from there.

bindings/rascal/models/IP_generic_md.py
max-veit: > Why does this class needs to be serializable in the first place?

This may again be a relic of basing the class off of `IP_ase_interface.py`. I agree that it doesn't really need to be serializable since there's only one well-defined initialization pathway; I'll see about taking it out.

bindings/rascal/models/krr.py
max-veit: eh, typo. I had to add this documentation myself because the initial method wasn't documented :/

bindings/rascal/models/krr.py
max-veit: Not yet; I ended up just using the fix in cf0e660.

bindings/rascal/models/krr.py
max-veit: First off, I hate the name "pseudo-points". "Support" or "basis" points is much nicer (and more accurate), but "sparse points" isn't *wrong* (sparse GPs are a thing, see Rasmussen & Williams). If you really want to launch a campaign to change this all across librascal, then by all means, open an issue and let's debate it there.

README.rst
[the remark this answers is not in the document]
Luthaf: #327

bindings/rascal/models/IP_generic_md.py
Luthaf: I think the same argument applies for `IP_ase_interface.py` to be renamed to `ase_interface.py`/`ase_driver.py`

bindings/rascal/models/IP_generic_md.py
Luthaf: (removing the assumption of fixed number of atoms is a pet peeve of mine, since I worked on GCMC simulations for my PhD)

I'm fine with having the assumption clearly documented for now!

bindings/rascal/models/IP_generic_md.py
Luthaf: It is quite easy to write a efficient version taking a file as input if we have a version taking an array, but the reverse is not true. Since this is Python, could we do a `isintance()` check? If we have a string, use it as a path; else use it as an array of species.

bindings/rascal/models/krr.py
Luthaf: #328

bindings/rascal/models/IP_generic_md.py
max-veit: Ah, sorry, didn't mean to call GCMC an "edge case"! I'll add a check for # atoms consistency, since that's easy to do here.

bindings/rascal/models/IP_generic_md.py
max-veit: I would actually prefer a more explicit version: Take either a structure template, or a species array, and if neither is provided then complain. Sound good?

Luthaf: As keyword arguments? Yes, that sounds good!

max-veit:
I'm pretty much done on the bindings side; perhaps @lgigli can have a look at the notebook comments since he's the one who wrote it?

examples/iPi/zundel/zundel_IP.ipynb
max-veit: Also this is now outdated by the fix #322 so we can probably just take it out.

bindings/rascal/models/krr.py
max-veit: Ok, I found a way that works -- it's not pretty, because the function iterates through the frames twice and I can't find a way to reset a tqdm bar once it's finished (you'd think that would be a pretty basic feature, but whatever).

max-veit:
not sure what's happening with the doc build... Sphinx is complaining about some utf-8 characters in a file it's trying to interpret in ASCII mode (why??? why won't ASCII die already??) -- but I'm pretty sure I didn't add any funny characters or change file encodings or the build configuration since the last successful doc build. And it runs just fine on my machine.

bindings/rascal/models/genericmd.py
```diff
@@ -0,0 +1,130 @@
...
+    def calculate(self, positions, cell_matrix):
+        """Calculate energies and forces from position/cell update
+
+        positions   Atomic positions (Nx3 matrix)
+        cell_matrix Unit cell (in ASE format, cell vectors as rows)
+                    (set to zero for non-periodic simulations)
+
+        The units of positions and cell are determined by the model JSON
+        file; for now, only Å is supported.  Energies, forces, and
+        stresses are returned in the same units (eV and Å supported).
```
Luthaf, line 88: This Å is the non ascii character I think. The error is a Python 2 one, which can be fixed by switching to Python 3 or adding `# -*- coding: UTF-8 -*-` as the first line of this file

max-veit: ...ok, but that was present in the last working version (3281e07) too. And why would it be using Python 2 to compile the docs? I thought we were pretty clear that this project is Python 3 only.

max-veit: The most maddening thing is that Sphinx won't even tell me which file has the offending characters, so I'm wondering if `nbsphinx` has something to do with this.

max-veit:
> not sure what's happening with the doc build... Sphinx is complaining about some utf-8 characters in a file it's trying to interpret in ASCII mode (why??? why won't ASCII die already??) -- but I'm pretty sure I didn't add any funny characters or change file encodings or the build configuration since the last successful doc build. And it runs just fine on my machine.

This is full (!) traceback of the error, looks like it happens when trying to load the `ExtBabel` extension for some reason. Why this is failing _now_ and not in earlier builds, I have no idea.

```
# Sphinx version: 3.5.3
# Python version: 3.6.9 (CPython)
# Docutils version: 0.17 release
# Jinja2 version: 2.11.3
# Last messages:

# Loaded extensions:
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/sphinx/cmd/build.py", line 279, in build_main
    args.tags, args.verbosity, args.jobs, args.keep_going)
  File "/usr/local/lib/python3.6/dist-packages/sphinx/application.py", line 241, in __init__
    self.setup_extension(extension)
  File "/usr/local/lib/python3.6/dist-packages/sphinx/application.py", line 402, in setup_extension
    self.registry.load_extension(self, extname)
  File "/usr/local/lib/python3.6/dist-packages/sphinx/registry.py", line 417, in load_extension
    mod = import_module(extname)
  File "/usr/lib/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 994, in _gcd_import
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/usr/local/lib/python3.6/dist-packages/sphinx/builders/latex/__init__.py", line 25, in <module>
    from sphinx.builders.latex.util import ExtBabel
  File "/usr/local/lib/python3.6/dist-packages/sphinx/builders/latex/util.py", line 13, in <module>
    from docutils.writers.latex2e import Babel
  File "/usr/local/lib/python3.6/dist-packages/docutils/writers/latex2e/__init__.py", line 575, in <module>
    for line in fp:
  File "/usr/lib/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 69: ordinal not in range(128)
```

max-veit:
omg no way, the build is failing due to a bug in the latest version (0.17) of `docutils`. Specifically, the '©' character in the copyright notice of `docutils.sty`... already reported [here](https://sourceforge.net/p/docutils/bugs/414/) and fixed in the latest Sourceforge version, but for us I think we can just force the previous version in `requirements.txt` until a hotfix is pushed to pypi.

examples/i-PI/zundel/zundel_IP.ipynb
```diff
@@ -0,0 +1,783 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Let us now compute the overall performance of the model. This is done by calculation of the Root-Mean-Square Error (RMSE), i.e. the standard deviation of the residuals according to the standard formula:\n",
+    "\n",
+    "$$\n",
+    "       \\text{RMSE} = \\sqrt{\\frac{1}{n_{\\text{test}}}\\sum_i (y_{\\text{pred}}^{i} - y_{\\text{test}}^{i})^2}\n",
+    "$$\n",
+    "   \n",
+    "which we can compare to the standard deviation of the test set itsself to quantify how much the model captures the energy variations in the test set. The $\\% \\text{RMSE}$ of our model is about $5 \\%$ of the training set STD, which is sufficiently accurate to run MD safely. \n",
```
Luthaf, line 536:
```suggestion
    "which we can compare to the standard deviation of the test set itself to quantify how much the model captures the energy variations in the test set. The $\\% \\text{RMSE}$ of our model is about $5 \\%$ of the training set STD, which is sufficiently accurate to run MD safely. \n",
```

Luthaf (approved):
The code looks good, feel free to merge after squashing related commits together!

PR #309, opened 2020-12-11
Make tests compile in gcc 10

### LINKED ISSUE #308: Cannot compile tests with gcc 10
It seems our tests aren't compiling with gcc 10 because `size_t` is no longer implicitly defined:

```
[ 51%] Built target test_adaptor_neighbour_list
[ 51%] Building CXX object tests/CMakeFiles/test_key_standardisation.dir/test_key_standardisation.cc.o
In file included from /home/max/Work/librascal/tests/test_key_standardisation.cc:28:
/home/max/Work/librascal/src/rascal/utils/key_standardisation.hh:38:25: error: ‘size_t’ has not been declared
   38 |   template <typename T, size_t MaxOrder>
      |                         ^~~~~~
/home/max/Work/librascal/src/rascal/utils/key_standardisation.hh:47:15: error: ‘size_t’ has not been declared
   47 |     template <size_t Order>
      |               ^~~~~~
/home/max/Work/librascal/src/rascal/utils/key_standardisation.hh:48:53: error: ‘Order’ was not declared in this scope
   48 |     explicit KeyStandardisation(const std::array<T, Order> & tup)
      |                                                     ^~~~~
/home/max/Work/librascal/src/rascal/utils/key_standardisation.hh:48:58: error: template argument 2 is invalid
   48 |     explicit KeyStandardisation(const std::array<T, Order> & tup)
```

Fortunately, a simple `using std::size_t;` seems to fix it. Maybe it's just another one of those old carryovers from C that C++ compilers accepted until now, and gcc 10 is just getting more strict about it.

For reference:
```
% gcc --version
gcc (GCC) 10.2.0
Copyright (C) 2020 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

max-veit:
Fix #308 

Also in this PR, replace gcc-9 with gcc-10 as the latest GCC version (it seems to be the default in Ubuntu now as well: https://packages.ubuntu.com/groovy/gcc).

Tasks before review:

          elsewhere in the code.
       - gcc-10 tests on circleci (in progress)

### DIFF
--- a/.circleci/Dockerfile
+++ b/.circleci/Dockerfile
@@ -25,6 +25,7 @@ RUN apt-get update && \
     apt-get install -y \
         gcc-5 g++-5 \
         gcc-9 g++-9 \
+        gcc-10 g++-10 \
         clang-4.0 \
         clang-9 \
         clang-format-9
--- a/.circleci/Dockerfile
+++ b/.circleci/Dockerfile
@@ -24,7 +24,6 @@ RUN wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -  && \
 RUN apt-get update && \
     apt-get install -y \
         gcc-5 g++-5 \
-        gcc-9 g++-9 \
         gcc-10 g++-10 \
         clang-4.0 \
         clang-9 \
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ version: 2
 # indicated by the CC/CXX environment variables.
 default_job: &default_job
   docker:
-    - image: cosmoepfl/rascal-ci:4
+    - image: cosmoepfl/rascal-ci:5
   steps:
     - checkout
     - run:
@@ -69,8 +69,8 @@ jobs:
     <<: *default_job
   valgrind:
     environment:
-      CC: gcc-9
-      CXX: g++-9
+      CC: gcc-10
+      CXX: g++-10
       CMAKE_EXTRA: -DRASCAL_TESTS_USE_VALGRIND=ON -DTYPE_ARCHITECTURE=core2
     <<: *default_job
 
@@ -188,8 +188,8 @@ workflows:
       - valgrind
       - coverage
       - gcc-5
-      - gcc-10-debug
       - gcc-10
+      - gcc-10-debug
       - clang-4
       - clang-9
       - clang-9-debug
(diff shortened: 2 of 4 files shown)


Luthaf:
>  @Luthaf, do you have the login? If so, would you mind updating the image? I already updated and tested the Dockerfile, so it should be pretty straightforward.

I do, and so does @ceriottm. I can update it myself, or add you to the owners of the image, as you prefer!

max-veit:
In that case, it might be easiest just to give me the login, since I already have the image pre-built on my machine. And you can always revert it if I make some mistake and everything goes pear-shaped, right?

Luthaf:
Once you have an account on docker hub, you will need to build & push a new version of the image (tag it as version 4) and then update the circleCI yaml with `image: cosmoepfl/rascal-ci:4`

max-veit:
Ok, this codecov block is actually really annoying for small PRs like this, especially since it's a minor bugfix that doesn't even change total code coverage. Is there any way to get around it?

Luthaf:
> Ok, this codecov block is actually really annoying for small PRs like this, especially since it's a minor bugfix that doesn't even change total code coverage. Is there any way to get around it?

The issue here is that the block just started working, so we merged a few PR decreasing coverage since codecov was added. 

The coverage was around 80% when it was installed, and dropped to 76 when merging #262 (see the timeline in https://app.codecov.io/gh/cosmo-epfl/librascal/branch/master). I think the best course of action here is to reset the target to 75% coverage, and bring coverage up later on. You can change the target coverage in the `codecov.yml` file.

max-veit:
...what? Then how did any subsequent PRs pass the codecov check? When was this check implemented, exactly? Because I find it a bit weird that a PR that decreases coverage by 3.5% gets merged with no problems, but a tiny one like this is held up on a technicality.

Luthaf:
The coverage collection was implemented in april, as part of #261. But it was not adding a checkmark/github status to the PR until a week ago, when @ceriottm added the codecov app to the cosmo-epfl repositories. This was an oversight on my part at the time, due to changes in how codecov works since the first time I used it.

The coverage decrease in the mean time is a bit unfortunate, but we can change the target coverage for now and work to improve coverage later, to not block this PR.

Luthaf (approved):

Luthaf:
You should consider squashing commits together before merging this!

PR #312, opened 2021-01-20
Implementation of radial dimension reduction

agoscinski:
Implements a possibility of reducing the radial channel for representations based on spherical expansion by giving a projection matrix on the python side (e.g. computed by PCA). This PR contains two very verbose commit messages

The first commit contains only structural changes, minor bug fixes as preperation for the radial dimension reduction:

- The Spline hyper changes 
   ```py
   {"type": "Spline", "accuracy": 1e-8} -> {"Spline": {"accuracy": 1e-8}}
  ```
  have been adapted in the `spherical_expansion_reference.ubjson`
  directly by using the ubjson file, and *not* by touching the reference calculation. Some minor bugs in the generation script have been fixed but the generation script has not been executed to generate the file. Instead this has been done:
  ```
  python -mubjson tojson spherical_expansion_reference.ubjson spherical_expansion_reference.json
  ```
  ```py
  import json
  import ubjson
  with open('spherical_expansion_reference.json') as f: 
      data = json.load(f) 
  for i in range(len(data['rep_info'])):
      for j in range(len(data['rep_info'][0])):
          if data['rep_info'][i][j]['hypers']['radial_contribution']['optimization']['type']== 'None':
              data['rep_info'][i][j]['hypers']['radial_contribution']['optimization'] = dict()
          else:
              data['rep_info'][i][j]['hypers']['radial_contribution']['optimization'] = dict(Spline=dict(accuracy=1e-8))

  with open('spherical_expansion_reference.ubjson', "wb") as f: 
      ubjson.dump(data, f) 
  ```

The second commit implements radial dimension reduction as new optimization argument, which allows a reduction of the radial channels of spherical representations. This can be combined with Spline to get a dimension reduction basically computationally for free:

  - Implementations for the new `OptimizationType`s `RadialDimReduction` and `RadialDimReductionSpline` have been added to the `RadialContributionHandler`. On the python side, the user still gives independent optimization arguments (Spline and RadialDimReduction) which are "mixed" to one enum on the C++ side. If more optimization args will be implemented in the future, one has to think about splitting `RadialContributionHandler` into multiple classes.

  - `finalize_coefficients_der` has been replaced by `finalize_neighbour_derivative` which does the inner loop part of `finalize_coefficients_der` to separate the `pairs_with_self_pair()` into `self_pair` (accessed by `center.get_atom_ii()`) part and `pairs()`. I thought this would be necessary to make `RadialDimReduction` work with gradient, but I am not sure now (see open problems)

  - If `RadialDimReduction` is used `max_radial` parameter has to agree with the reduced `max_radial` (`n_components`). This was the more convenient way (instead of choosing the expanded `max_radial`), when computing `RadialContribution`, it operates with the expanded `max_radial` value which is inferred from the `projection_matrices` size. 
 - ubjson do not work with jsons containing arrays (`projection_matrices` is an array), therefore the generator scripts cannot be used for tests. Tests for the derivative have been added `spherical_expansion_radial_derivative` to check if it runs through for the identity matrix on the C++ side while a test for spherical expansion with unitary `projection_matrices` has been added on the python side. With unitary `projection_matrices` the dot product kernel should be still the same. The python test adds 0.4s testing time on the big cosmo machines.

  - had to remove one `(max_radial, max_angular)` pair parameter in the gradient `spherical_expansion_radial_derivative` test, because the `projection_matrices` in `RadialDimReduction` must have a dimension of shape `(max_angular+1, max_radial, expanded_max_radial)`, to prevent a check in the test fixture I removed one `(max_radial, max_angular)` pair, so we only have one pair.

  - I am not 100% sure but I think some spline tests were not executed before (at least not with the parameters in the tests), because the `{"accuracy": 1e-8}` block was interpreted as array and then ignored in the setup switching to the default accuracy which is 1e-8. In general I spent a lot of time in this PR debugging "jsons are interpreted as arrays" errors because the rules for it seem a bit arbitrary (sometimes you should use {{ ... }} sometimes just { ... }, certain definition work only in a `json_obj["key"] =` definition but not in a compile time definition).

Here a code example with timings which gives reasonable results

```py
import timeit

import numpy as np

import ase.io

import rascal
from rascal.representations import (
    SphericalExpansion,
    SphericalInvariants,
)

nb_executions = 3

max_angular = 4
max_radial = 5
expanded_max_radial = 15


frames = ase.io.read("/home/alexgo/lib/librascal/reference_data/inputs/small_molecules-1000.xyz", ":")

species_bus = [1,6,7,8]
projection_matrices = {species: np.random.rand(max_angular+1, max_radial, expanded_max_radial).tolist() for species in species_bus}

for representation in [SphericalExpansion, SphericalInvariants]:
    print("\n\n\nUsing representation", representation.__name__)

    def no_optimization():
        hypers = {
            "interaction_cutoff": 6.0,
            "cutoff_smooth_width": 1.0,
            "max_radial": expanded_max_radial,
            "max_angular": max_angular,
            "gaussian_sigma_type": "Constant",
            "gaussian_sigma_constant": 0.5,
            "expansion_by_species_method": "user defined",
            "global_species": species_bus,
        }
        rep = representation(**hypers)
        return rep.transform(frames).get_features(rep)

    print("  no_optimization:")
    print("  features.shape", no_optimization().shape)
    print(" ", timeit.timeit(no_optimization, number=nb_executions))

    def radial_dim_reduction():
        hypers = {
            "interaction_cutoff": 6.0,
            "cutoff_smooth_width": 1.0,
            "max_radial": max_radial,
            "max_angular": max_angular,
            "gaussian_sigma_type": "Constant",
            "gaussian_sigma_constant": 0.5,
            "expansion_by_species_method": "user defined",
            "global_species": species_bus,
            "optimization": {"RadialDimReduction": {"projection_matrices": projection_matrices}}
        }
        rep = representation(**hypers)
        return rep.transform(frames).get_features(rep)
    print()
    print("  radial_dim_reduction")
    print("  features.shape", radial_dim_reduction().shape)
    print(" ", timeit.timeit(radial_dim_reduction, number=nb_executions))


    def radial_dim_reduction_spline():
        hypers = {
            "interaction_cutoff": 6.0,
            "cutoff_smooth_width": 1.0,
            "max_radial": max_radial,
            "max_angular": max_angular,
            "gaussian_sigma_type": "Constant",
            "gaussian_sigma_constant": 0.5,
            "expansion_by_species_method": "user defined",
            "global_species": species_bus,
            "optimization": {
                "Spline": {"accuracy": 1e-8},
                "RadialDimReduction": {"projection_matrices": projection_matrices}
                }
        }
        rep = representation(**hypers)
        return rep.transform(frames).get_features(rep)
    print()
    print("  radial_dim_reduction_spline")
    print("  features.shape", radial_dim_reduction_spline().shape)
    print(" ", timeit.timeit(radial_dim_reduction_spline, number=nb_executions))
```
with output
```
Using representation SphericalExpansion
  no_optimization:
  features.shape (17743, 1500)
  6.833654425005079

  radial_dim_reduction
  features.shape (17743, 500)
  6.9693316519988

  radial_dim_reduction_spline
  features.shape (17743, 500)
  3.2171284960058983



Using representation SphericalInvariants
  no_optimization:
  features.shape (17743, 11250)
  29.891001459007384

  radial_dim_reduction
  features.shape (17743, 1250)
  10.382615477006766

  radial_dim_reduction_spline
  features.shape (17743, 1250)
  6.498090021006647
```

Open problems:
- [x] Derivative for GTO without Spline does not work at the moment, the spherical expansion has to be adapted. I think the center contribution is not only orthonormalized in the current implementation, but I don't see quite through the derivative computation.
- [x] There is no species check on the C++ side that the users gives projection matrices for all required species. Not sure how to do add a check in structural nice way.

TODOs:
- [x] add documentation on the python side
- [x] make workable example how to use different optimization args

Tasks before review:

          elsewhere in the code.
          auto-formatting is sensible
          own branch


felixmusil:
Nice work @agoscinski, I will have time to review it properly only in a couple of weeks though... 

I am not sure to understand the open problem on the GTO derivative. Was it already a problem? What does it break? What should be done to solve it?

agoscinski:
So it is the same problem we had with the orthonormalization of the spherical expansion coefficients when using a spline. It happens at the end applied on all coefficients with the `finalize_coefficients_der` but when using `RadialDimReduction` without Spline, the orthonormalization has to happen in the `compute_neighbour_derivative` such that the `finalize_coefficients_der` function can be skipped. For the spline we did not have to do this, because the derivative of the spline is used.

ceriottm:
@agoscinski I think the syntax for "optimization" arguments has changed in master - can you update this so it is consistent?

src/rascal/representations/calculator_spherical_expansion.hh
```diff
@@ -2181,6 +2191,67 @@ namespace rascal {
...
 
+    // species in projection matrices have to be checked here, because
+    // calculators do not have access to manager during construction
+    if (hypers.count("radial_contribution") &&
+        hypers.at("radial_contribution").get<json>().count("optimization") &&
+        hypers.at("radial_contribution")
+            .get<json>()
+            .at("optimization")
+            .get<json>()
+            .count("RadialDimReduction")) {
```
felixmusil, line 2202: in set_hyperparameters() you already know if there is a RadialDimReduction so I would set a flag there to simplify this block.

src/rascal/representations/calculator_spherical_expansion.hh
```diff
@@ -2181,6 +2191,67 @@ namespace rascal {
...
+        err_str
+            << "Projection matrices should be of type std::map<std::string, "
+               "std::vector<std::vector<std::vector<double>>>>.";
+        throw std::logic_error(err_str.str());
+      }
+      std::vector<int> projection_matrices_species{};
+      for (auto it = json_projection_matrices.begin();
+           it != json_projection_matrices.end(); ++it) {
+        projection_matrices_species.push_back(std::stoi(it->first));
+      }
```
felixmusil, line 2235: in `set_hyperparameters()` you already access this data. Why not set `projection_matrices_species` there if `RadialDimReduction` and use the fact that if `projection_matrices_species.count()>0` then `RadialDimReduction` was set or something like this with a simple boolea ?

felixmusil (changes requested):
first round of comments

src/rascal/representations/calculator_spherical_expansion.hh
```diff
@@ -1962,8 +2441,17 @@ namespace rascal {
...
+        auto & coefficients_center_pair_gradient{
+            expansions_coefficients_gradient[center_pair]};
+        radial_integral->template finalize_neighbour_derivative<ThreeD>(
+            coefficients_center_pair_gradient);
+        for (auto neigh : center.pairs()) {
+          auto & coefficients_neigh_gradient{
+              expansions_coefficients_gradient[neigh]};
+          radial_integral->template finalize_neighbour_derivative<ThreeD>(
+              coefficients_neigh_gradient);
+        }  // for (neigh : center)
```
felixmusil, line 2454: so using just `center.pairs_with_self_pair()` and removing the `get_atom_ii` specific code is not equivalent to what is happening here?

felixmusil (changes requested):
Overall it's in a good shape but there are a few outstanding points:

+ What do you think about making the change from 'optimization_args' to 'optimization' backward compatible for some time at least? In the spirit of #353.

+
>finalize_coefficients_der has been replaced by finalize_neighbour_derivative which does the inner loop part of finalize_coefficients_der to separate the pairs_with_self_pair() into self_pair (accessed by center.get_atom_ii()) part and pairs(). I thought this would be necessary to make RadialDimReduction work with gradient, but I am not sure now (see open problems)

I am not sure why it would be required.

+
>ubjson do not work with jsons containing arrays (projection_matrices is an array), therefore the generator scripts cannot be used for tests.  

projection_matrices can be converted to a list of list of list which should work with ubjson. That's basically the type of container for the representations stored for the tests. So one should be able to add a small test on the cpp side too.


>In general I spent a lot of time in this PR debugging "jsons are interpreted as arrays" errors because the rules for it seem a bit arbitrary

Yeah I had the same problems many times.


+ Could you have a look at the offending notebook?

+ you forgot to put `optimization=radial_contribution["optimization"],` in `init_params` of `_get_init_params()` of spherical_covariants.py like in spherical_expansion.py.

agoscinski (commented):
> What do you think about making the change from 'optimization_args' to 'optimization' backward compatible for some time at least? In the spirit of #353.

Have done this

>finalize_coefficients_der has been replaced by finalize_neighbour_derivative which does the inner loop part of finalize_coefficients_der to separate the pairs_with_self_pair() into self_pair (accessed by center.get_atom_ii()) part and pairs(). I thought this would be necessary to make RadialDimReduction work with gradient, but I am not sure now (see open problems)


Have restored finalize_coefficients_der, the point was important if we would offer to use RadialDimReduction without Spline, but we decided not to do it.

> projection_matrices can be converted to a list of list of list which should work with ubjson. That's basically the type of container for the representations stored for the tests. So one should be able to add a small test on the cpp side too.

I don't know why I did not do this in the first place. I have added a test case to the ubjson. The test should be automatically added now, right?

> Could you have a look at the offending notebook?

have fixed something, not sure if this fixes everything, as soon as the test run through here it is fixed.

> you forgot to put optimization=radial_contribution["optimization"], in init_params of _get_init_params() of spherical_covariants.py like in spherical_expansion.py.

added

src/rascal/representations/calculator_spherical_expansion.hh
felixmusil: It is since `compute_neighbour_contribution` is called `n_neigh` times instead of `n_atom` times. In any case I still feel that it's cleared to have this check in `compute_impl` with a loop over atoms and a flag to check if `RadialDimReduction` is on but I don't want to argue about it either. It works well as is, you could change `.at()` with a direct access since at that point you know that you can access this element.

src/rascal/representations/cutoff_functions.hh
```diff
@@ -224,7 +224,7 @@ namespace rascal {
         if (this->rate < 0) {
           throw std::runtime_error("RadialScaling's rate should be positive");
         }
-        this->exponent = hypers.at("exponent").at("value").get<int>();
+        this->exponent = hypers.at("exponent").at("value").get<double>();
```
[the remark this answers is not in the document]
felixmusil: Please make it clear that it can be any positive double on the python side so that there is no confusion for the user.

src/rascal/representations/calculator_spherical_expansion.hh
```diff
@@ -1086,31 +1124,389 @@ namespace rascal {
...
+        this->reduced_radial_neighbour_derivative.resize(this->n_components,
+                                                         this->max_angular + 1);
+        this->reduced_radial_neighbour_derivative.setZero();
+      }
+
+      // Returns the precomputed center contribution
+      template <size_t Order, size_t Layer>
+      Vector_Ref
+      compute_center_contribution(ClusterRefKey<Order, Layer> & /*center*/,
+                                  int center_type) {
```
felixmusil, line 1354: I just realized `compute_neighbour_contribution` and `compute_center_contribution` with `RadialDimReductionSpline` now requires the center type as input. This change of the signature is unnecessary since you have `center` at your disposable to get this information in the function. Moreover it these functions are expected to have a particular signature so to make this can of changes one would have to change the signature of every other `compute_center_contribution` and `compute_neighbour_contribution`.
Please remove the atom type input and get it from `center` and `neighbor`.

agoscinski: I did this because in the level of `compute_center_contribution` it is StructureManager where i can use `get_atom_type`, but within ` compute_[center,neighbour]_contribution` it is upcasted to ClusterRefKey and has no `get_atom_type`

src/rascal/representations/calculator_spherical_expansion.hh
agoscinski: Yes but for a correct check one needs to check all neighbours types, since one could mask central species atoms, which then only appear as neighbours and not anymore as central atoms I had this problem with the chemical shielding dataset.
So I need to check all central atom types and all neighbour types

src/rascal/representations/cutoff_functions.hh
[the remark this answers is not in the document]
agoscinski: Would **positive float** be clear and correct?
```py
        .. code:: python

            dict(rate=...,
                 scale=...,
                 exponent=...)

        where :code:`...` should be replaced by the desired positive float.
```

src/rascal/representations/calculator_spherical_expansion.hh
felixmusil: you are right, forgot about this difference. Probably `ClusterRefKey` should also have the atom type but that's another story.

src/rascal/representations/calculator_spherical_expansion.hh
felixmusil: I did not think of that right.

src/rascal/representations/cutoff_functions.hh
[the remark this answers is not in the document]
felixmusil: yes

felixmusil (approved):
Looks good to me now! You just need to fix the docs CI before merging.

PR #317, opened 2021-03-03
Feat/export grad

felixmusil:
Improve flexibility by providing the users with a simple way to get the feature gradients and connection information as numpy arrays.

The functions `get_features_gradient` and `get_ij` have been added to `AtomsList` and a python test has been added.


bindings/bind_py_structure_manager.cc
```diff
@@ -616,6 +618,96 @@ namespace rascal {
...
+          }
+
+          return features;
+        },
+        R"(Get the dense feature matrix associated with the calculator and
+        the collection of structures (managers) using the list of keys
+        provided. Only applicable when Calculator uses BlockSparseProperty.)");
+
+    manager_collection.def(
+        "get_ij",
```
Luthaf, line 678: I'd call this function `get_neighbors_for_gradient` or something similar. `get_ij` is not really telling me what this function returns.

bindings/bind_py_structure_manager.cc
```diff
@@ -798,8 +890,8 @@ namespace rascal {
   void bind_cluster_refs(py::module & m_internal) {
     add_cluster_refs<1, 0, 6>::static_for(m_internal);
     add_cluster_refs<2, 0, 6>::static_for(m_internal);
-    add_cluster_refs<3, 0, 6>::static_for(m_internal);
-    add_cluster_refs<4, 0, 6>::static_for(m_internal);
+    // add_cluster_refs<3, 0, 6>::static_for(m_internal);
```
Luthaf, line 893: Why comment the code out? If it no longer works, it should be removed

bindings/rascal/neighbourlist/structure_manager.py
```diff
@@ -114,6 +114,31 @@ def get_features(self, calculator, species=None):
...
+        calculator : Calculator (an object owning a _representation object)
+
+        species :  list of atomic number to use for building the dense feature
+        matrix computed with calculators of name Spherical*
+
+        Returns
+        -------
+        dX_dr : ndarray of size (3*(n_neighbor+n_atom), n_features)
+            returns the gradient of representation with respect to the atomic positions that have been computed with the calculator as a dense matrix.
+        """
```
Luthaf, line 130: This documentation should refer to `get_ij` as the way to get atomic indexes corresponding to the entries in the matrix returned by this function.

src/rascal/structure_managers/property_block_sparse.hh
```diff
@@ -968,13 +974,32 @@ namespace rascal {
             const auto & neigh_key_val = neigh_val[key];
             ConstMapSoapGradFlat_t neigh_key_val_flat(neigh_key_val.data(),
                                                       ThreeD, inner_size);
+            // for (int i_der{0}; i_der < ThreeD; i_der++) {
```
Luthaf, line 977: why is this commented out?

felixmusil:
I am not quite sure about the error that is happening. It seems that the CI machine can't reach a python package from internet. Have you ever had such issues ?

bindings/rascal/neighbourlist/structure_manager.py
```diff
@@ -128,6 +156,48 @@ def get_features_by_species(self, calculator):
         """
         return self.managers.get_features_by_species(calculator._representation)
 
+    def get_neighbors_for_gradient(self):
```
Luthaf, line 159: This needs unit tests

bindings/rascal/neighbourlist/structure_manager.py
```diff
@@ -128,6 +156,48 @@ def get_features_by_species(self, calculator):
...
+        -------
+        ij : np.array of size (n_neighbor+n_atom, 5)
+            Get informations necessary to the computation of gradients returned
+            by `get_features_gradient`. It has as many rows as as the number
+            gradients and each columns correspond to the index of the atomic
+            structure, central atom, the neighbor atom and their atomic species.
+        """
+        return self.managers.get_neighbors_for_gradient()
+
+    def get_atoms_for_predictions(self):
```
Luthaf, line 171: This needs unit tests

bindings/rascal/neighbourlist/structure_manager.py
```diff
@@ -128,6 +156,48 @@ def get_features_by_species(self, calculator):
...
+        -------
+        ij : np.array of size (n_atoms, 3)
+            Get informations necessary to the computation of predictions using
+            the representation from `get_features`. It has as many rows as the
+            number representations and they correspond to the index of the
+            structure, the central atom and its atomic species.
+        """
+        return self.managers.get_atoms_for_predictions()
+
+    def get_direction_vectors(self):
```
Luthaf, line 183: This needs unit tests

bindings/rascal/neighbourlist/structure_manager.py
```diff
@@ -128,6 +156,48 @@ def get_features_by_species(self, calculator):
...
+        """
+        Returns
+        -------
+        direction_vector : np.array of size (n_neighbor+n_atom, 3)
+            Get the direction vectors from the atoms to their neighbors.
+        """
+        return self.managers.get_direction_vectors()
+
+    def get_distances(self):
```
Luthaf, line 192: This needs unit tests

Luthaf (commented):
The code look OK, some function are not unit tested though. I'd like to see unit tests, checking this function only on a very simple example (H2O molecule for example); and verifying that we get the right output.

bindings/bind_py_structure_manager.cc
```diff
@@ -616,6 +618,196 @@ namespace rascal {
...
+    manager_collection.def(
+        "get_features_gradient",
+        [](ManagerCollection_t & managers, const Calculator & calculator,
+           py::list & all_keys_l) {
+          using Manager_t = typename ManagerCollection_t::Manager_t;
+          using Prop_t =
+              typename Calculator::template PropertyGradient_t<Manager_t>;
+          using Keys_t = typename Prop_t::Keys_t;
+          if (managers.size() == 0) {
+            throw std::runtime_error(
```
Luthaf, line 630: Why is this an error instead of returning a 0x0 or 0xn_features matrix?

bindings/bind_py_structure_manager.cc
```diff
@@ -616,6 +618,196 @@ namespace rascal {
...
+                R"(There are no structure to get features from)");
+          }
+          Keys_t all_keys;
+          if (all_keys_l.size() > 0) {
+            // convert the list of keys from python in to the proper type
+            for (py::handle key_l : all_keys_l) {
+              auto key = py::cast<std::vector<int>>(key_l);
+              all_keys.insert(key);
```
Luthaf, line 638: what happen if an invalid key is passed by the user? I did not find a check that all keys are valid.

bindings/bind_py_structure_manager.cc
```diff
@@ -616,6 +618,196 @@ namespace rascal {
...
+            for (auto key_l : managers.get_keys(calculator)) {
+              all_keys.insert(key_l);
+            }
+          }
+
+          auto property_name{managers.get_calculator_name(calculator, true)};
+
+          const auto & property_ =
+              *managers[0]->template get_property<Prop_t>(property_name);
+          // assume inner_size is consistent for all managers
```
Luthaf, line 651: is there a way to check this assumption? Will the copy in the full matrix below will fail with a runtime error in this case?

bindings/bind_py_structure_manager.cc
```diff
@@ -616,6 +618,196 @@ namespace rascal {
...
+            i_row += n_rows_manager;
+          }
+
+          return features;
+        },
+        R"(Get the dense feature matrix associated with the calculator and
+        the collection of structures (managers) using the list of keys
+        provided. Only applicable when Calculator uses BlockSparseProperty.)");
+    manager_collection.def(
+        "get_atoms_for_predictions",
```
Luthaf, line 677: It's nice to have this function as well! I'm not sure with the name though, since this is not necessarily related to predictions. Something like `get_atoms_metadata` might be a bit better but feels too generic. Thoughts?

bindings/bind_py_structure_manager.cc
```diff
@@ -616,6 +618,196 @@ namespace rascal {
...
+        [](ManagerCollection_t & managers) {
+          if (managers.size() == 0) {
+            throw std::runtime_error(
+                R"(There are no structure to get features from)");
```
Luthaf, line 681: why not return an empty matrix here?

bindings/bind_py_structure_manager.cc
```diff
@@ -616,6 +618,196 @@ namespace rascal {
...
+          }
+          return ij_mat;
+        },
+        R"(Get informations necessary to the computation of predictions. It has
+        as many rows as the number representations and they correspond to the
+        index of the structure, the central atom and its atomic species.
+        np.array with shape (n_atoms, 3))");
+
+    manager_collection.def(
+        "get_neighbors_for_gradient",
```
Luthaf, line 709: This function name should be similar to the name of `get_atoms_for_predictions`. So maybe `get_gradients_metadata`, or something else if you have a better idea!

bindings/bind_py_structure_manager.cc
```diff
@@ -616,6 +618,196 @@ namespace rascal {
...
+        [](ManagerCollection_t & managers) {
+          if (managers.size() == 0) {
+            throw std::runtime_error(
+                R"(There are no structure to get features from)");
```
Luthaf, line 713: same with returning an empty matrix in this case

bindings/bind_py_structure_manager.cc
```diff
@@ -616,6 +618,196 @@ namespace rascal {
...
+          }
+          auto n_neighbors{0};
+          for (auto & manager : managers) {
+            n_neighbors += manager->get_nb_clusters(2);
```
Luthaf, line 717: how does this deal with half vs full neighbor list? The number of clusters of size 2 already deals with this distinction?

bindings/bind_py_structure_manager.cc
```diff
@@ -616,6 +618,196 @@ namespace rascal {
...
+        of the structure, the central atom, the neighbor atom and their atomic
+        species.
+        np.array with shape (n_neighbor+n_atoms, 5))");
+
+    manager_collection.def(
+        "get_direction_vectors",
+        [](ManagerCollection_t & managers) {
+          if (managers.size() == 0) {
+            throw std::runtime_error(
+                R"(There are no structure to get features from)");
```
Luthaf, line 752: same with the empty matrix

bindings/bind_py_structure_manager.cc
```diff
@@ -616,6 +618,196 @@ namespace rascal {
...
+        The zeros vectors correspond to the central atom to itself so that this
+        matrix matches the shape of the array returned by
+        `get_neighbors_for_gradient`.
+        np.array with shape (n_neighbor+n_atoms, 3))");
+    manager_collection.def(
+        "get_distances",
+        [](ManagerCollection_t & managers) {
+          if (managers.size() == 0) {
+            throw std::runtime_error(
+                R"(There are no structure to get features from)");
```
Luthaf, line 784: same with empty matrix

bindings/bind_py_structure_manager.cc
```diff
@@ -794,12 +986,11 @@ namespace rascal {
     bind_update_unpacked<Manager_t>(manager);
   }
 
-  //! Bind the ClusterRef up to order 4 and from Layer 0 to 6
+  //! Bind the ClusterRef up to order 3 and from Layer 0 to 6
```
Luthaf, line 989: out of curiosity, why remove the order 4?

Luthaf (commented):
I have a few final comments on what we do on some corner cases and invalid input

bindings/bind_py_structure_manager.cc
felixmusil: I would say that if one calls this function and did not give a structure then it's not intentional so it's better not to silently carry on.

bindings/bind_py_structure_manager.cc
felixmusil: the assumption is valid since it's the same `property_name` which by default encodes a uniform `inner_size`. Nevertheless I guess that some twisted user could in principle have non uniform `inner_size` in a set of structures but he would have had to try hard to get to this point...

bindings/bind_py_structure_manager.cc
felixmusil: by construction get_nb_clusters(2) will get the effective number of clusters. Then it's up to the user to know what is going on and adapt the code accordingly.

bindings/bind_py_structure_manager.cc
felixmusil: it's not used at the moment and probably won't be used in the near future so it just increases the compilation time for nothing.

bindings/bind_py_structure_manager.cc
felixmusil: what do you mean by invalid? there might be keys that are not in the representation at hand (and that's exactly the point of this possibility actually) so zeros will be inserted there.
If by invalid you mean that they can't be cast to std::vector<int> then an error will be raised by the relevant code.

bindings/bind_py_structure_manager.cc
felixmusil: renamed

bindings/bind_py_structure_manager.cc
felixmusil: renamed

bindings/bind_py_structure_manager.cc
Luthaf: Ok, that's good!

bindings/bind_py_structure_manager.cc
Luthaf: I don't understand: is the effective number of cluster is the number of clusters as returned by the neighbor list, meaning `FullNL::get_nb_clusters(2) == 2 * HalfNL::get_nb_clusters(2)`? Or the actual number of different clusters, i.e. `FullNL::get_nb_clusters(2) == HalfNL::get_nb_clusters(2)`? 

As a Python user, I don't have control over the full or half neighbor list, so I can not really adapt the code here.

felixmusil: `get_nb_clusters(2)` returns the number of clusters of order 2 in the current manager so for sure `FullNL::get_nb_clusters(2) != HalfNL::get_nb_clusters(2)`.
As a python user you know which type of manager you are working with, e.g. full or half NL, so when you work with the `dX_dr` with the `ij` mapping you know that things will probably need to be different depending on the type of NL you have been using. For instance in the tests I know that I have a full NL so I compute the kernel matrix elements with that in mind so the code would have to be modified a bit to properly handle a half NL.

I am not quite sure what is the problem here actually.

Luthaf: The "standard" way of using rascal is the following, right?

```
frames = ase.io.read("file.xyz", ":")
soap = SphericalInvariants(...)

soap.transform(frames)
```

When I'm doing this, I have no idea is the code is using a full or half neighbors list, and I don't know how to control this parameter. I'd guess one kind of neighbor list is selected for me, but which one exactly is not documented nor guaranteed to stay the same.

The problem I see is that since I don't know which NL the code is using, I might write some code with one assumption, only to get wrong results later if my assumption of half/full NL turned out to be wrong.

felixmusil: I see. You can't choose at the moment because half NL is not exposed to python. So this is something that we should keep in mind when this feature is made available but not at the moment. Also this kind of choice would be made explicit in the hypers of `SphericalInvariants`.

felixmusil: These features are independent of the representation so to check which kind of NL is being used should look at the `AtomsList`. More specifically `AtomsList.nl_options` argument or the name of `AtomsList.managers` which are very explicit about this topic. So in your example doing:
```
frames = ase.io.read("file.xyz", ":")
soap = SphericalInvariants(...)
features = soap.transform(frames)
print(features.nl_options, features.managers) 
```
will give you well enough information. I am not sure how it could be easier to access this information?

Luthaf: I did not knew this was possible, but it looks like it is not documented in https://cosmo-epfl.github.io/librascal/reference/python.html. 

Then I think this should work as is, provided we add the corresponding documentation to the website.

docs/source/reference/python.rst
```diff
@@ -9,6 +9,8 @@ This list is incomplete. You can help by *expanding it*!
 Atoms and cells
 ===============
 
+.. autofunction:: rascal.neighbourlist.structure_manager.AtomsList
```
Luthaf, line 12: I think this should be `.. autoclass:: rascal.neighbourlist.structure_manager.AtomsList`

Luthaf (approved):
Let's merge this! I would squash the commits while/before merging, since the history in this branch is not very informative.

PR #318, opened 2021-03-10
add the possibility of pickling object inheriting from `BaseIO`

felixmusil:
`BaseIO` implements `__setstate__` and `__getstate__` so that the objects inheriting from it can be pickled.

### DIFF
--- a/bindings/rascal/representations/spherical_expansion.py
+++ b/bindings/rascal/representations/spherical_expansion.py
@@ -161,7 +161,7 @@ class SphericalExpansion(BaseIO):
         Required arguments are all the hyperparameters named in the
         class documentation
         """
-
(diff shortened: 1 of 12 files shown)


bindings/rascal/representations/spherical_expansion.py
```diff
@@ -161,7 +161,6 @@ def __init__(
         Required arguments are all the hyperparameters named in the
         class documentation
         """
-
```
agoscinski, line 164: To make a clean merge, I would remove this change, since the file was not touched otherwise

agoscinski (approved):
Remove only this one line in `bindings/rascal/representations/spherical_expansion.py`, since you do not touch really it.

Othwerwise looks fine, just keep in mind to squash merge to one meaningful commit message.

PR #321, opened 2021-03-17
Bugfix/covariant redux

ceriottm:
Fixes a nasty bug in the covariant calculator, and renames the options to follow the more verbose style we use throughout

A "code cleanup" effort had messed up the order of loops, which mysteriously leads to covariant features, that are however NOT the right thing. This basically reverts these changes and gets correct lambda-SOAP features.

Remaining tasks, out of scope of this pull request: (optional)

Make sphericalcovariants *really* usable, e.g. with also kernel calculation and general usage


felixmusil (approved):

max-veit:
Just a note, this was merged without any tests added to confirm that the bug has been fixed and will not come up again. That's against our developer guidelines; let's not get lax on them again.

max-veit:
For example, let's have a test that checks that the spherical covariants features really are covariant for a simple system. That would have caught this bug a long time ago!

ceriottm:
> For example, let's have a test that checks that the spherical covariants features really are covariant for a simple system. That would have caught this bug a long time ago!

Actually, and mysteriously, not. Whatever this was computing, it was covariant. Only it was not what we have always written as a lambda-SOAP, two loops were swapped. There is already a regtest, that for some reasons failed to identify this problem - I suspect because it was reset after the bug was introduced, or because the test was created after the bug was introduced. 

I agree that once we implement more functionalities, a test against TENSOAP is much needed.

PR #322, opened 2021-03-17
fix #310

### LINKED ISSUE #310: Error with structure wrapping with ASE: atoms outside the unit cell - neighbourlist of structures failed
[Structure_wrap_bug.xyz.txt](https://github.com/cosmo-epfl/librascal/files/5801119/Structure_wrap_bug.xyz.txt)

In order to compute the SOAP vectors of a given structure (previously saved as a simple .xyz file) in librascal, it is necessary to first load it using ase.io.read and then wrap the structure inside the unit cell. The following code snippet has been used in a practical case:

```py
HYPERS = {
    'soap_type': 'PowerSpectrum',
    'interaction_cutoff': 6,
    'max_radial': 6,
    'max_angular': 4,
    'gaussian_sigma_constant': 0.5,
    'gaussian_sigma_type': 'Constant',
    'cutoff_smooth_width': 0.5,
    'radial_basis': 'GTO',
    'normalize' : False,
    'inversion_symmetry': True,
}
soap = SOAP(**HYPERS)

f = read("Structure_wrap_bug.xyz", format="xyz")
f.wrap(eps=1e-18)
cell = f.get_cell()

feat = soap.transform(f).get_features(soap)
x.append(feat.mean(axis=0))
x = np.asarray(x)
```

For the structure that I report in the attachment, despite the ase wrapping had been successful (because all the atom positions lie inside the unit cell), librascal sent this error message: 

**RuntimeError: Neighbourlist of structures failed because: Error: At least one atom is outside the unit cell. This adaptor expects atoms to be folded into the original unit cell. ccoord = (7, 2, 6), nboxes = (8, 8, 8)**

This error persists even by changing the librascal branch, or even by just using the up-to-date master branch.

felixmusil:
This PR aims at fixing #310. The problem is the result of some unexpected interactions between the unit cell, the cutoff, and the procedure to generate periodic images in the linked cell algorithm and I solve it by introducing a tolerance `bound_tol==1e-4` that potentially increases a bit the number of unit cell duplication and decreases a bit the area in which atoms are binned. The core of the problem is that atoms positions could be accepted for binning because they would be barely inside. But for the binning, an addition would lead to rounding that would place the atom outside of the binning boundary, e.g. `1.999999999998<2.` but `1.999999999998-1. -> 1. >= 1.`.
In addition to the fix, the elements of the problematic structure of #310 are now tested. 

Could try out this branch and see if the recurring problems you had with the NL still persists? @max-veit @ceriottm @lgigli @sshepherd637


ceriottm:
hallelujah! works, which is an excellent starting point. two questions though:
1. if it's a "rounding error" problem, why such a large tolerance? what are the consequences of having this large tolerance?
2. similarly, why increasing by so much the cell "skin"?

ceriottm:
Empirically, I reduced tol down to 1e-12 and my dataset-of-death runs smoothly.

Luthaf (commented):
I don't know how this part of the code works, so I can not reasonably review this. 

Since you understand what is going on here, one thing that would help both my review and anyone having to work on this in the future would be to write some high level documentation of what the algorithm is doing, and how. I think such documentation could live either at the class level for `AdaptorNeighbourList`; or inside the manual.

felixmusil:
> Since you understand what is going on here, one thing that would help both my review and anyone having to work on this in the future would be to write some high level documentation of what the algorithm is doing, and how. I think such documentation could live either at the class level for AdaptorNeighbourList; or inside the manual.

There are ample comments on the general idea and details of the implementation of the method, aren't they instructive enough ? if not what seems to be missing ?

felixmusil:
> if it's a "rounding error" problem, why such a large tolerance? what are the consequences of having this large tolerance?

The tolerance is large with respect to the typical precision around 1 but if the cells becomes larger it might not be so large anymore. The check could be made on scaled coordinates but that would require large changes to the code that don't seem warranted here. 
In any case, if we assume that we work with the typical atomic units (or any unit system that would result in positions and cell between 0.1 and 10000) a tolerance of 1e-4 or 1e-6 is good enough. Moreover, the risk of having a large tolerance here would be to reject a ghost atom from the NL that is actually within the cutoff of an atom of the unit cell. Considering that we have very small contributions from atoms that are at the edge of the interaction cutoff, I think that 1e-4 or 1e-6 are not huge risks.

Luthaf:
> There are ample comments on the general idea and details of the implementation of the method, aren't they instructive enough ? if not what seems to be missing ? 

There a some comment on functions in https://github.com/cosmo-epfl/librascal/blob/f49b0ccdbc9eb89ae95bb794a8df96ea02ffd57d/src/rascal/structure_managers/adaptor_neighbour_list.hh, but this is missing a high level overview. Ideally, one should be able to understand the algorithm used without having to read all the code.

One concrete thing that would help (and I thing a neighbor list is well adapted to this) are schematic representations of the steps: assigning atoms to bins, dealing with ghosts, ...

ceriottm:
> > if it's a "rounding error" problem, why such a large tolerance? what are the consequences of having this large tolerance?
> 
> The tolerance is large with respect to the typical precision around 1 but if the cells becomes larger it might not be so large anymore. The check could be made on scaled coordinates but that would require large changes to the code that don't seem warranted here.
> In any case, if we assume that we work with the typical atomic units (or any unit system that would result in positions and cell between 0.1 and 10000) a tolerance of 1e-4 or 1e-6 is good enough. Moreover, the risk of having a large tolerance here would be to reject a ghost atom from the NL that is actually within the cutoff of an atom of the unit cell. Considering that we have very small contributions from atoms that are at the edge of the interaction cutoff, I think that 1e-4 or 1e-6 are not huge risks.

One has to be very careful with this. Atoms close to the cutoff can have unexpected consequences, e.g. if the cutoff is not smooth you can get jumps in finite-difference derivatives, a problem that took me and Venkat two months to debug. If a ghost disappears 1e-4 before the cutoff, then jumps might appear even when the cutoff is smooth.

ceriottm:
In all event, this is a urgent bugfix, and so I'd only continue this discussion if it can be actioned on easily - otherwise a tiny jump is way better than a non-working NL

felixmusil:
> but this is missing a high level overview

The docstring of `make_full_neighbour_list` seems quite clear on the general idea to me and I feel that this is the place where it should be detailed, don't you think?  What is missing there in your opinion so that I can try to clarify it? @Luthaf

Luthaf:
At the risk of repeating myself:

> One concrete thing that would help (and I thing a neighbor list is well adapted to this) are schematic representations of the steps: assigning atoms to bins, dealing with ghosts, ...

----

Reading the doxygen comment in `make_full_neighbour_list`, the text is not using full sentences, and some terms are not defined. Some things unclear to me while reading:
- which kind of `mesh` are we using? 
- how is `"Triclinicity accounted for"`?
- why does "All atoms are expected to be inside the unit cell.", instead of wrapping them ourself?
- what are `i-atoms`?

Hope this make it clearer what is unclear in this part of the documentation

felixmusil:
I tried to improve the description of `make_full_neighbour_list`, tell me if it's clearer now. The schematics and general documentation efforts don't belong to this bugfix PR though.

I also changed the definition of the tolerance so that is proportional to the largest length of the bounding box. Hopefully, this will be good enough to cover future uses.

Luthaf:
It is much better already. One thing missing is a description of how the algorithm deals with triclinic cells. Given this is non trivial, I'd rather have it well documented. From the doc, I'd guess that triclinic cells are wrapped inside a bigger orthorhombic cell that contains the full cell and a lot of empty space, am I right? If so, this should be added to the description.

Some related questions to this: what happen with highly skewed cells (cells with angles <30°)? Or if the cell is not aligned with the usual axis systems (**a** along **x** and **b** in the **xy** plane)?

> the schematics and general documentation efforts don't belong to this bugfix PR though.

I disagree: since writing documentation is quite hard, it is better to do it in a piecewise fashion, as we touch related code. Otherwise, writing/improving documentation is left to *some time later* and never happens in practice.

felixmusil:
> One thing missing is a description of how the algorithm deals with triclinic cells. 

Actually triclinicity is not dealt with as a special case. Since we create a cubic box that contains the whole unit cell plus some cutoff any shape of the lattice will by definition be properly handled. This is why I don't think it needs more details except for a clear mention that this is a general routine that can be applied to all types of lattices.  

> Some related questions to this: what happen with highly skewed cells (cells with angles <30°)? Or if the cell is not aligned with the usual axis systems (a along x and b in the xy plane)?

These corner cases are tested already and as mentioned before they are properly handled by definition. That being said for performance purposes I would recommend unskewing these types of unit cells beforehand since they would be quite inefficient to work with.

ceriottm (approved):
We could dig a bit deeper into the most robust and efficient way to do a neighbor list, but this is such a pressing bug that I'd rather see this merged as it is, and then we may one day revisit.

PR #325, opened 2021-03-22
Quick fix for circular imports

### LINKED ISSUE #294: Circular python import of SparsePoints
On the python side if one imports something from the `rascal.models` before importing `rascal.representations` one gets the following error message.
```
ImportError                               Traceback (most recent call last)
<ipython-input-3-939eeea548d7> in <module>
----> 1 from rascal.models import Kernel
~/lib/librascal/build/rascal/models/__init__.py in <module>
----> 1 from .sparse_points import SparsePoints
      2 from .krr import train_gap_model, KRR
      3 from .kernels import Kernel
~/lib/librascal/build/rascal/models/sparse_points.py in <module>
----> 1 from ..utils import BaseIO
      2 from ..lib._rascal.models import kernels
      3 from ..neighbourlist import AtomsList
      4 
      5 # names of existing pseudo points implementation on the pybinding side.
~/lib/librascal/build/rascal/utils/__init__.py in <module>
      5 from .io import (BaseIO, to_dict, from_dict, get_current_io_version,
      6                  get_supported_io_versions, dump_obj, load_obj)
----> 7 from .cur import CURFilter
      8 from .scorer import get_score, print_score
      9 
~/lib/librascal/build/rascal/utils/cur.py in <module>
      5                                   convert_selected_global_index2perstructure_index)
      6 
----> 7 from ..models.sparse_points import SparsePoints
      8 import numpy as np
      9 from scipy.sparse.linalg import svds
ImportError: cannot import name 'SparsePoints'
```
There is probably a circular import of SparsePoint happening somewhere in `rascal.models` which gets resolved by `rascal.representations`

max-veit:
Fix #294 

I can't guarantee that this fixes _all_ our circular import problems, but I think this addresses the one triggered by trying to run

```python
from rascal.models import <something>
```

I'm not sure how exactly to test this -- it pops up pretty randomly when changing something in the Python code, and depends on import ordering (which is a nightmare), so somehow it has not been caught by the Python binding tests until now. Suggestions welcome.

### DIFF
--- a/bindings/rascal/utils/__init__.py
+++ b/bindings/rascal/utils/__init__.py
@@ -1,7 +1,3 @@
-from .fps import fps, FPSFilter
-from ..lib._rascal.utils import ostream_redirect
-from ..lib import utils
-from copy import deepcopy
 from .io import (
     BaseIO,
     to_dict,
@@ -11,6 +7,10 @@ from .io import (
     dump_obj,
     load_obj,
 )
+from .fps import fps, FPSFilter
+from ..lib._rascal.utils import ostream_redirect
+from ..lib import utils
+from copy import deepcopy
 from .cur import CURFilter
 from .scorer import get_score, print_score
 
--- a/bindings/rascal/utils/__init__.py
+++ b/bindings/rascal/utils/__init__.py
@@ -7,27 +7,14 @@ from .io import (
     dump_obj,
     load_obj,
 )
+from .misc import is_notebook
+
+# Warning potential dependency loop: FPS imports models, which imports KRR,
+# which imports this file again
 from .fps import fps, FPSFilter
+
+# function to redirect c++'s standard output to python's one
 from ..lib._rascal.utils import ostream_redirect
-from ..lib import utils
 from copy import deepcopy
 from .cur import CURFilter
 from .scorer import get_score, print_score
-
-# function to redirect c++'s standard output to python's one
-ostream_redirect = utils.__dict__["ostream_redirect"]
-
-
-def is_notebook():
-    from IPython import get_ipython
-
-    try:
-        shell = get_ipython().__class__.__name__
-        if shell == "ZMQInteractiveShell":
-            return True  # Jupyter notebook or qtconsole
-        elif shell == "TerminalInteractiveShell":
-            return False  # Terminal running IPython
-        else:
-            return False  # Other type (?)
-    except NameError:
-        return False  # Probably standard Python interpreter
--- a/bindings/rascal/utils/misc.py
+++ b/bindings/rascal/utils/misc.py
@@ -1,0 +1,17 @@
+"""Miscellaneous useful utilities"""
+
+
+def is_notebook():
+    """Is this being run inside an IPython Notebook?"""
+    from IPython import get_ipython
+
+    try:
+        shell = get_ipython().__class__.__name__
+        if shell == "ZMQInteractiveShell":
+            return True  # Jupyter notebook or qtconsole
+        elif shell == "TerminalInteractiveShell":
+            return False  # Terminal running IPython
+        else:
+            return False  # Other type (?)
+    except NameError:
+        return False  # Probably standard Python interpreter
--- a/bindings/rascal/models/__init__.py
+++ b/bindings/rascal/models/__init__.py
@@ -1,3 +1,2 @@
-from .sparse_points import SparsePoints
 from .krr import train_gap_model, KRR
 from .kernels import Kernel
--- a/bindings/rascal/utils/fps.py
+++ b/bindings/rascal/utils/fps.py
@@ -1,5 +1,6 @@
 from .io import BaseIO
 from ..lib import sparsification
+from ..models.sparse_points import SparsePoints
 
 from .filter_utils import (
     get_index_mappings_sample_per_species,
@@ -236,7 +237,6 @@ class FPSFilter(BaseIO):
             raise NotImplementedError("method: {}".format(self.act_on))
 
     def filter(self, managers, n_select=None):
-        from ..models import SparsePoints
 
         if n_select is None:
             n_select = self.Nselect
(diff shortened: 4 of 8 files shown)


max-veit:
And a note for the future: This whole mess was mostly caused by creating "convenience imports" that just rename existing functions in librascal. Let's try to avoid those and try to keep our internal dependency graph as simple as possible to avoid running into such issues again.

max-veit:
Please note that I had to take the `SparsePoints` alias-import out of `bindings/rascal/models/__init__.py` in order to resolve the circular import, so you can no longer just do `from rascal.models import SparsePoints`. The correct way to import these is now `from rascal.models.sparse_points import SparsePoints`. The examples and tests have been updated accordingly.

Note that this style of "aliasing" (importing) objects into a higher package level's `__init__.py` file is generally discouraged for exactly this reason, so eventually we'll want to take the rest of those alias-imports out, especially if they cause problems in the future.

bindings/rascal/utils/__init__.py
```diff
@@ -11,23 +7,19 @@
...
     load_obj,
 )
+from .misc import is_notebook
+
+# Warning potential dependency loop: FPS imports models, which imports KRR,
+# which imports this file again (including is_notebook() above)
+from .fps import fps, FPSFilter
+
+# TODO(max-veit) not needed...?
+# from ..lib import utils
```
Luthaf, line 17: Is this TODO resolved? If yes, you should remove the comment

bindings/rascal/utils/__init__.py
```diff
@@ -11,23 +7,19 @@
...
-        shell = get_ipython().__class__.__name__
-        if shell == "ZMQInteractiveShell":
-            return True  # Jupyter notebook or qtconsole
-        elif shell == "TerminalInteractiveShell":
-            return False  # Terminal running IPython
-        else:
-            return False  # Other type (?)
-    except NameError:
-        return False  # Probably standard Python interpreter
+# TODO(max-veit) we already have this imported...?
```
Luthaf, line 24: same about TODO

bindings/rascal/utils/misc.py
```diff
@@ -0,0 +1,17 @@
+"""Miscellaneous useful utilities"""
+
+
+def is_notebook():
```
Luthaf, line 4: I don't really see why we have this function in librascal ... Is this used anywhere else in the code or is this just a convenience function for librascal users?

Luthaf (commented):
The fix looks good as far as fixing circular imports go. 

> Note that this style of "aliasing" (importing) objects into a higher package level's __init__.py file is generally discouraged for exactly this reason, so eventually we'll want to take the rest of those alias-imports out, especially if they cause problems in the future.

What makes you think this is discouraged? 

I generally prefer to use it as much as possible because it is nicer from the user perspective, which in turns force me to structure my classes and modules as a cycle-free graph (usually a tree). If necessary, I then use late imports (directly inside class methods) to break up module cycles. At the same time, I've mostly worked with relatively small python packages, so I'd like to read more on this subject if you have links!

bindings/rascal/utils/__init__.py
max-veit: Ah, I had this because the following statement seemed to be a re-phrasing of `from rascal._rascal.utils import ostream_redirect` and is therefore not needed. Would you agree?

Luthaf: that does look like it is doing the same thing yes. It is a bit weird that both are around. Maybe @felixmusil knows more about why this was here?

bindings/rascal/utils/misc.py
max-veit: It was used to decide whether to import `tqdm` in the command-line version or the notebook version, in one of the tutorials (yet to be merged, I think). I know we wanted to make the dependency on `tqdm` optional, but if it is present, we will probably want to import the appropriate version.

bindings/rascal/utils/__init__.py
max-veit: Seems it's used almost exclusively in reference-data scripts, so it's not critical either way. I'll just take it out then.

bindings/rascal/utils/misc.py
max-veit: It was one of the i-PI tutorials actually, see e.g.: https://github.com/cosmo-epfl/librascal/pull/307/commits/af2c7bd0b1e3917f8c7aaab10b850418f500d46a

Luthaf: Ok, ideally I would move this function to the tutorial code itself, and not inside librascal. But this can happen as part of #307

Luthaf (approved):
Looks good! Feel free to merge once CI is happy!

Also, please squash the commits before or during merge 😃

max-veit:
> What makes you think this is discouraged?

I might be misremembering a blog post or StackOverflow answer I read once, which of course I can't find again. Although I think [this one](https://stackoverflow.com/questions/7336802/how-to-avoid-circular-imports-in-python/37126790#37126790) does a pretty good job of explaining the issue and various fixes.

> I generally prefer to use it as much as possible because it is nicer from the user perspective, which in turns force me to structure my classes and modules as a cycle-free graph (usually a tree).

Agree, I was hoping this was one step towards making our dependency graph easier to understand.

>If necessary, I then use late imports (directly inside class methods) to break up module cycles. At the same time, I've mostly worked with relatively small python packages, so I'd like to read more on this subject if you have links!

I don't like late imports because then you can no longer see the module's dependencies just by looking at the top of the file. Note that there were already several instances of `SparsePoints` being imported within a function, which might actually make things worse if the import error is only triggered on running a function... Anyway, I took those out, and since SparsePoints is a fairly low-level object anyway I don't think we really need to expose it in `models`.

bindings/rascal/utils/misc.py
max-veit: That PR is scheduled to be merged soon anyway (I split this fix off from that PR because I thought it was more urgent)

PR #334, opened 2021-04-07
special adaptor for the k-space implementation

felixmusil:
Implements an adaptor for the k-space version of the spherical expansion. The adaptor works on the order 1 (list of atoms) and build the order 2. This calculator needs to iterate over neighbors that are all of the other atoms in the structure without cutoff which is what is done.


max-veit:
Do we really need a whole new adaptor (and all the additional code and complexity that comes with it) just to be able to iterate over all atom pairs? I thought the solution currently implemented in the k-space branch worked, was very simple, and didn't involve any ugly hacks.

In any case, it seems that iterating over all pairs is much more useful than just for k-space methods, no? So why is this just called a "k-space adaptor"? Wouldn't a name like "all-pairs" be simpler and more indicative of what it actually does?

felixmusil:
> Do we really need a whole new adaptor (and all the additional code and complexity that comes with it) just to be able to iterate over all atom pairs?

Unfortunately, we do in order to handle derivatives properly, e.g. we need 'pair' to address a gradient.

> In any case, it seems that iterating over all pairs is much more useful than just for k-space methods, no? 

It is not as general as that actually because it corresponds to a very special case. For the k-space method, we want a double iteration over all of the registered atoms for structures that are fully periodic. So no periodic image is considered here even though in the case you are mentioning they actually would have to be considered.
This is why I decided to make this very specific Adaptor. Do you agree with this rationale?

max-veit:
> Unfortunately, we do in order to handle derivatives properly, e.g. we need 'pair' to address a gradient.

Ok, sure, but let's give some more thought to see if there's a simpler way to do this. To me, it seems like it would be _much_ simpler to just have a plain neighbourlist (AdaptorLinkedCell or whatever it's called, although in this case an $N^2$ neighbourlist would be just fine if we're storing all pair entries anyway), set the cutoff to the largest cell dimension, and exclude ghosts from the iteration. This should be possible, no? (and if it's not, let's make it possible. I would much prefer that to the hundreds of lines of duplicated functionality and added complexity required to create a new adaptor.)

> It is not as general as that actually because it corresponds to a very special case. For the k-space method, we want a double iteration over all of the registered atoms for structures that are fully periodic. So no periodic image is considered here even though in the case you are mentioning they actually would have to be considered.
> This is why I decided to make this very specific Adaptor. Do you agree with this rationale?

No, because what you're describing is exactly the "all-pairs" iteration pattern. I didn't say anything about considering periodic images. Even if we can't come up with a use case beyond k-space methods (including Ewald and particle-mesh) right now, it would be less confusing to call it an "all-pairs" iterator in case we do find future applications beyond k-space methods that require such an iteration pattern.

agoscinski (changes requested):
Seems mostly fine for me. The kspace adaptor needs slight improvements in the documentation. The main confusion comes from the fast that ghost neighbors are considered in the documentation while the adaptor's purpose is to remove them.

I am still not sure why we did not add additional arguments to our full neighborlist adaptor to determine the desired type of neighborlist, since this has already done and we already have plans to restructure the code in a new project, I am fine with it.

There are a lot of files which don't really have changed but are listed under changes. I am not sure why. I would be interested in why this happens and in fixing this, but it is not that important that I would make it a requirement for merging. You can see it for example in the files:
- examples/needs_updating/Spherical_invariants_and_database_exploration.ipynb
- bindings/rascal/models/asemd.py
- bindings/rascal/models/krr.py

Issue #340, opened 2021-04-20
`convert_to_structure_list` fails with ASE 3.21 when given a single frame


Luthaf:
https://github.com/cosmo-epfl/librascal/blob/060126beaa2d1d342c325751f66b6393b0542c47/bindings/rascal/neighbourlist/structure_manager.py#L230-L246

When passing a single frame (not a list) (as done for example in the Zundel_IP.ipynb notebook), the following error is raised:
```
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-17-b4d24b25ef96> in <module>
     11     f.set_positions(positions+[1,1,1])
     12     f.wrap(eps=1e-18)
---> 13     m = soap.transform(f)
     14     y_pred.append(model.predict(m))
     15     f_pred.append(model.predict_forces(m))

/opt/miniconda3/lib/python3.7/site-packages/rascal/representations/spherical_invariants.py in transform(self, frames)
    330         """
    331         if not isinstance(frames, AtomsList):
--> 332             frames = AtomsList(frames, self.nl_options)
    333 
    334         self._representation.compute(frames.managers)

/opt/miniconda3/lib/python3.7/site-packages/rascal/neighbourlist/structure_manager.py in __init__(self, frames, nl_options, start, length, managers)
     74         else:
     75             # if python structure
---> 76             structures = convert_to_structure_list(frames)
     77             managers = StructureCollectionFactory(nl_options)
     78             try:

/opt/miniconda3/lib/python3.7/site-packages/rascal/neighbourlist/structure_manager.py in convert_to_structure_list(frames)
    244             else:
    245                 raise RuntimeError(
--> 246                     "Cannot convert structure of type {}".format(type(frame))
    247                 )
    248         structure = sanitize_non_periodic_structure(structure)

RuntimeError: Cannot convert structure of type <class 'ase.atom.Atom'>
```

The issue is that `ase.Atoms()` is now an instance of `Iterable`, and thus is not wrapped in a list before iterating over it. This changed recently on the ASE side. In ASE 3.19, `isinstance(ase.Atoms(), Iterable)` returns False; but on the latest version (3.21) this returns True.

PR #341, opened 2021-04-20
fix issue #340

felixmusil:
Change how the inputs are checked in `convert_to_structure_list` to keep the compatibility with various versions of ASE.

### DIFF
--- a/bindings/rascal/neighbourlist/structure_manager.py
+++ b/bindings/rascal/neighbourlist/structure_manager.py
@@ -1,3 +1,4 @@
+from collections.abc import Iterable
 from functools import reduce
 from operator import and_
 
@@ -241,9 +242,8 @@ def convert_to_structure_list(frames):
     """
     if isinstance(frames, Atoms):
         frames = [frames]
-
-    assert isinstance(frames, list)
-    assert isinstance(frames[0], Atoms) or isinstance(frames[0], dict)
+    elif not isinstance(frames, Iterable):
+        raise ValueError("Must pass either an ase.Atoms object or an iterable")
 
     structure_list = neighbour_list.AtomicStructureList()
     for frame in frames:
--- a/bindings/rascal/neighbourlist/structure_manager.py
+++ b/bindings/rascal/neighbourlist/structure_manager.py
@@ -1,10 +1,10 @@
-from collections.abc import Iterable
 from functools import reduce
 from operator import and_
 
 import numpy as np
 
 from ase.geometry import wrap_positions
+from ase import Atoms
 
 from ..lib import neighbour_list
 from .base import (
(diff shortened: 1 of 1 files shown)


max-veit:
This seems OK to me, but why use assertions to check for the type of atoms list? I would rather have it check first for ASE Atoms, then for Iterable, and if neither of the two applies, then fail with an informative error message (saying e.g. that you need to pass either Atoms or a list of Atoms).

felixmusil:
It's true that it is typically done as you outline in the codebase but I feel this makes the docstring quite redundant in this case and thus heavier to extend/maintain later on. The `assert` seems clear enough to me and if not then the docstring is there to help. What do you think?

max-veit:
Assertion errors are not clear as error messages; they're typically used instead _by programmers_ to debug conditions _they_ think should hold true, no matter what the user input. A clear error message, on the other hand, does not make the docstring redundant, instead telling the user what exactly they did wrong. This will make maintenance _much easier_ later on because if one of us passes the wrong thing by mistake, we know exactly what's going on and don't have to waste 15 minutes figuring out where in the code their error got spit out from.

bindings/rascal/neighbourlist/structure_manager.py
```diff
@@ -228,8 +228,23 @@ def get_neighbourlist(structure, options):
...
+    Returns
+    -------
+    structure_list : AtomicStructureList
+
+    """
+    if isinstance(frames, Atoms):
         frames = [frames]
+
+    assert isinstance(frames, list)
+    assert isinstance(frames[0], Atoms) or isinstance(frames[0], dict)
```
max-veit, line 246:
```suggestion
    elif not isinstance(frames, Iterable):
        raise ValueError("Must pass either an ase.Atoms object or an iterable")
```
The check of whether the objects being iterated over are Atoms (or dict, or similar) is handled later on (see line 257 in this version).

bindings/rascal/neighbourlist/structure_manager.py
```diff
@@ -1,10 +1,10 @@
-from collections.abc import Iterable
 from functools import reduce
```
max-veit, line 1:
```suggestion
from collections.abc import Iterable
from functools import reduce
```

max-veit (changes requested):
Here, it's very simple but makes for a much better user (and developer) experience.

max-veit:
> Assertions are a systematic way to check that the **internal state** of a program is as the programmer expected, with the goal of catching bugs. In particular, they're good for **catching false assumptions** that were made while writing the code, or abuse 

https://wiki.python.org/moin/UsingAssertionsEffectively (emphasis mine)

max-veit (approved):
Ok, all good from my end, provided the tests pass **and** that you squash these into one commit with an informative message.

PR #357, opened 2021-05-26
Quick fixes for automated build issues

### LINKED ISSUE #355: CircleCI doc test fails "has no attribute 'tags'"
Hello, I have this error in my branch, which seems not limited to my branch. It also happens when I undo all changes to the notebook
```
Notebook error:
UndefinedError in examples/zundel_i-PI.ipynb:
'nbformat.notebooknode.NotebookNode object' has no attribute 'tags'
docs/CMakeFiles/doc.dir/build.make:57: recipe for target 'docs/CMakeFiles/doc' failed
make[3]: *** [docs/CMakeFiles/doc] Error 2
CMakeFiles/Makefile2:1256: recipe for target 'docs/CMakeFiles/doc.dir/all' failed
make[2]: *** [docs/CMakeFiles/doc.dir/all] Error 2
CMakeFiles/Makefile2:1268: recipe for target 'docs/CMakeFiles/doc.dir/rule' failed
make[1]: *** [docs/CMakeFiles/doc.dir/rule] Error 2
Makefile:591: recipe for target 'doc' failed
make: *** [doc] Error 2
```
https://app.circleci.com/pipelines/github/cosmo-epfl/librascal/2515/workflows/8a4a5e50-bd34-4754-a6ac-9a3756083475/jobs/21487

Haven't investigated it yet, but maybe someone already has a solution for this

### LINKED ISSUE #356: cpplint warns about C-style casts but clang-format doesn't fix it
Apparently the newest version of [cpplint](https://github.com/cpplint/cpplint) is warning about C-style casts, viz. this line: https://github.com/cosmo-epfl/librascal/blob/43a410c6d7cea3d17852e9c8c6fe332ec46f13df/tests/test_utils_sparsify.cc#L72
which is something `clang-format` doesn't seem to touch -- at least, it doesn't convert them to the Proper C++ Cast format.

To fix the issue, we can change the cast, or we could just fix the `cpplint` version in `requirements.txt` to mitigate future random upstream update lossage.

max-veit:
Fix two mostly unrelated issues, with the only commonality being that they were caused by upstream updates to dependencies.

Fix #356 - fix the offending line and pin down the cpplint version so we're not affected by future updates

Fix #355 - likewise, pin down some package versions to known working ones

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,14 +6,15 @@ jupyter
 matplotlib
 #qml
 # tqdm
-# for documentation
-sphinx==4.0.1
-breathe==4.30.0
+# Developers tools for compiling documentation
+sphinx>=2.1.2
+breathe>=4.14.1
+# Jinja2 v3 breaks nbsphinx
 jinja2==2.11.3
 sphinx_rtd_theme
-nbsphinx==0.8.5
-pygments==2.9.0
-docutils==0.17
+nbsphinx>=0.8.1
+pygments>=2.4.1
+# Developers tools for linting and formatting
 cpplint==1.5.5
 black==20.8b1
 # Developers tools - data generation scripts
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,7 +15,7 @@ pygments>=2.4.1
 # Weird bug in latest version (0.17) of docutils; can remove once fixed
 docutils==0.16
 # Developers tools - code formatting
-cpplint
+cpplint==1.5.5
 black==20.8b1 
 # Developers tools - data generation scripts
 ase
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,16 +7,15 @@ matplotlib
 #qml
 # tqdm
 # for documentation
-sphinx>=2.1.2
-breathe>=4.14.1
+sphinx==4.0.1
+breathe==4.30.0
+jinja2==2.11.3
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -7,16 +7,15 @@ matplotlib
...
-# Weird bug in latest version (0.17) of docutils; can remove once fixed
-docutils==0.16
-# Developers tools - code formatting
-cpplint
-black==20.8b1 
+nbsphinx==0.8.5
+pygments==2.9.0
+docutils==0.17
+cpplint==1.5.5
+black==20.8b1
```
Luthaf, line 18: I'm not fond of pinning all dependencies, but at least these are only development dependencies. 

We will have to be careful to update these regularly to not fall too much behind upstream changes.

Luthaf (approved):
Looks good to me, as long as Circle CI is happy =)

requirements.txt
Luthaf: > I'm just not very fond of all this random upstream breakage in the CI that seems to be happening recently.

That's an issue of pip/the Python ecosystem not having lockfiles and a good dependency update story. I mean, there is no real "per-project dependency" mechanism except duplicating a full python environment and changing your PATH (i.e. virtualenv). So pinning in "requirements.txt" is fine for me, but we need to actively update.

PR #369, opened 2021-06-29
Make SPH consistent with standard phase convention

### LINKED ISSUE #346: Real spherical harmonics include Condon-Shortley phase
As documented [here](https://cosmo-epfl.github.io/librascal/reference/cpp.html#_CPPv4N6rascal4math18SphericalHarmonicsE), our definition of the "real" spherical harmonics effectively still includes the Condon-Shortley phase of (-1)^m, making these closer to "complex flattened" spherical harmonics.

It seems a [common convention](https://en.wikipedia.org/wiki/Spherical_harmonics#Real_form), though, is to cancel out this phase when converting to real, and especially Cartesian, spherical harmonics (as otherwise the l=1 spherical harmonics would be equal to [-y, z, -x] for m=[-1, 0, 1] -- modulo some normalization factors). If this is the convention most commonly used by quantum chemistry codes, then we should switch to that convention.

Additionally, we should make the documentation on spherical harmonics a little clearer on the subject of exactly where the Condon-Shortley phase is included.

ceriottm:
The current implementation of SPH differs from the standard convention. This PR fixes it and adds a few demos/examples.  

Fix #346 

Tasks before review:

          elsewhere in the code.
          auto-formatting is sensible
          own branch


max-veit:
It seems that the calculator tests are failing now because of the sign change. These tests basically only check for backwards consistency, and since we're breaking that here, we need to regenerate the test data for the spherical expansion and covariants.

max-veit:
This PR seems to be feature-creeping quite a bit... I'm really not keen, for instance, on pulling in sympy for the list of dependencies (it's rather heavy); at the very least, we should make it optional (and deactivate the "new, improved" Python lambda-soap if it's not found).

ceriottm:
> This PR seems to be feature-creeping quite a bit... I'm really not keen, for instance, on pulling in sympy for the list of dependencies (it's rather heavy); at the very least, we should make it optional (and deactivate the "new, improved" Python lambda-soap if it's not found).

It's the most "mature" option around for computing CG and Wigner D. Made it import on call, so overhead and dependencies only manifest themselves if you use it. In all events, I wouldn't include pythonic lambda-SOAP in this PR, I'd leave it as an example in the demo notebook.

bindings/rascal/utils/cg_utils.py
```diff
@@ -0,0 +1,340 @@
...
+    return rcg
+
+
+# Convert from real to complex spherical harmonics (or corresponding
+# coefficients) and back. Uses the convention from Wikipedia
+isqrt2 = 1.0 / np.sqrt(2)
+sqrt2 = np.sqrt(2)
+
+
+def _r2c(sp):
```
Luthaf, line 53: Can I get some documentation on what this is doing, and the expected layout of the input?

bindings/rascal/utils/cg_utils.py
```diff
@@ -0,0 +1,340 @@
...
+    l = (len(sp) - 1) // 2
+    rc = np.zeros(len(sp), dtype=np.complex128)
+    rc[l] = sp[l]
+    for m in range(1, l + 1):
+        rc[l + m] = (sp[l + m] + 1j * sp[l - m]) * isqrt2 * (-1) ** m
+        rc[l - m] = (sp[l + m] - 1j * sp[l - m]) * isqrt2
+    return rc
+
+
+def _c2r(cp):
```
Luthaf, line 63: Same as above, can I get some documentation on what this is doing, and the expected layout of the input?

bindings/rascal/utils/cg_utils.py
```diff
@@ -0,0 +1,340 @@
...
+    rs = np.zeros(len(cp), dtype=np.float64)
+    rs[l] = np.real(cp[l])
+    for m in range(1, l + 1):
+        rs[l - m] = (-1) ** m * sqrt2 * np.imag(cp[l + m])
+        rs[l + m] = (-1) ** m * sqrt2 * np.real(cp[l + m])
+
+    return rs
+
+
+def spx_roll(spx, hypers):
```
Luthaf, line 74: This name is not very clear, and it is a public function. How about `reshape_spherical_expansion`?

Also, I would take `n_max` & `l_max` explicitly as parameters instead of relying on hypers.

bindings/rascal/utils/cg_utils.py
```diff
@@ -0,0 +1,340 @@
...
+    nmax = hypers["max_radial"]
+    lmax = hypers["max_angular"]
+    nid = len(spx)
+    nel = spx.shape[1] // (nmax * (lmax + 1) ** 2)
+    return spx.reshape(
+        (nid, nel, nmax, (lmax + 1) ** 2)
+    )  # (lm) terms are stored in a compact form
+
+
+def mslice(l):
```
Luthaf, line 88: what about

```suggestion
def spherical_harmonics_angular_slice(l):
```
or something similar with a clearer name?

bindings/rascal/utils/cg_utils.py
```diff
@@ -0,0 +1,340 @@
...
+    """
+    Simple helper function to get the slice corresponding to m=-l..l in a dense
+    storage block for Y^l_m-like coefficients
+    """
+    return slice(l * l, (l + 1) * (l + 1), 1)
+
+
+class WDReal:
```
Luthaf, line 96:
```suggestion
class WignerDMatrixReal:
```

bindings/rascal/utils/cg_utils.py
```diff
@@ -0,0 +1,340 @@
...
+            frame.positions = frame.positions @ self._rotation.T
+            frame.cell = frame.cell @ self._rotation.T
+        else:
+            print(frame["positions"])
+            frame["positions"] = self._rotation@frame["positions"]
+            frame["cell"] = self._rotation@frame["cell"] 
+        return frame
+
+
+class CGReal:
```
Luthaf, line 169:
```suggestion
class ClebshGordanReal:
```

Luthaf (commented):
I have a few naming improvements suggestions for the new public cg_utils module. Since we are not using Fortran 77, we can use more than 6 characters for the class & functions names 😄

bindings/rascal/utils/cg_utils.py
max-veit: *Cleb**sch**

Deut**sch**e Sprache beste Sprache :)

bindings/rascal/utils/cg_utils.py
ceriottm: I wanted to keep a short name since this is meant to be used a lot. would `ylmslice` be a good compromise?

bindings/rascal/utils/cg_utils.py
ceriottm: as a compromise, it takes `max_radial` and `max_angular`, so you can also call it with **hypers

bindings/rascal/utils/cg_utils.py
ceriottm: went with lm_slice

bindings/rascal/utils/cg_utils.py
Luthaf: 👍

max-veit:
I'm having difficulty understanding most of what is going on in the `equivariant_demo.ipynb` notebook. It reads very much like a "scratch notebook" used for development, and I don't think it would be very useful in its current form to anyone trying to use librascal to compute spherical covariants or do Clebsch-Gordan iteration.

A few suggestions to make it into a useful demo/tutorial notebook for others:
- Add _clear_, high-level explanations of what is happening in each section, and what we are supposed to make of the results shown. I see some references to papers, but it would be more helpful to write the actual equations involved and (where appropriate) include figures or visualizations of what is happening.
- It seems there are many different ways of computing the same thing -- which one is "right"? Which way is preferable? Which way is the simplest?
- How do I use all this stuff to do a real covariant learning problem? How do these utilities simplify such a computation? (Ideally we'd have something that works very much like rascal's existing `SphericalCovariants`, but are actually correct)
- Be careful with overuse of manual computations (like the section "Manual computation of SOAP" - what is even happening here?! Why do I need to know all this? What equations do these horrible multi-line expressions correspond to?), including also the "hidden" functions from `cg_utils` (like `_r2c` or `_wigner_d`). Use high-level utilities whenever possible, and use explicit expressions only when necessary to demonstrate what the high-level utilities are doing under the hood.
- Variable names... most of them are not descriptive at all. This is an important part of making the notebook accessible to others, not only new users but also developers.

ceriottm:
@vmax - this is a WIP, definitely "scratch notebook" at the moment. the
objective would be that there is a "textbook" manual implementation
compared with the output of librascal. basically we want to dot all the i's
and dash all the t's and write "what librascal really computes" kind of
example.

max-veit:
Ok, so it sounds like the current notebook could be split into a WIP/scratch part, kept in `sandbox/` (that's what that folder is for), and a complete, clear, and pedagogical tutorial on using the new `cg_utils` - which IMO would ideally include a general, utility version of the "simple" lambda-soap implementation (the one that uses real C-G iteration) shown in the notebook. That could then be our interim replacement for the currently-broken `SphericalCovariants` while we figure out what exactly is going on there. How does this plan sound?

ceriottm:
Hi all. Finished polishing the comments in the two pedagogic notebooks. In a way the notebooks double as tests, given they verify consistency of different definitions - is it worth to add separate tests of the cg_utils routines? I think it would be good to merge this sooner than later, as it finally gives a proper and tested way of computing lambda-SOAP, plus some routines to get people excited about NICE without having to also understand the NICE libraries.

examples/nice_demo.ipynb
```diff
@@ -0,0 +1,780 @@
...
+    "spex = SphericalExpansion(**spherical_expansion_hypers)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "selframe = frames[8];         # frame and l value used for the test\n",
```
Luthaf, line 107:
```suggestion
    "selframe = frames[8];         # frame used for the test\n",
```

examples/equivariant_demo.ipynb
```diff
@@ -0,0 +1,1285 @@
...
+    "abc = np.random.uniform(size=(3))*np.pi"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# this is the Cartesian rotation matrix (helper function, follows ZYZ convention)\n",
```
Luthaf, line 179: should this be 
```suggestion
    "# this is the Cartesian rotation matrix (helper function, follows XYZ convention)\n",
```

examples/equivariant_demo.ipynb
```diff
@@ -0,0 +1,1285 @@
...
+    "# Clebsch-Gordan iteration"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "CG coefficients $\\langle l_1 m_1; l_2 m_2 | LM\\rangle$ are essential in the quantum theory of angular momentum, describing the combination of two angular momentum kets to form a coupled basis\n",
+    "$$\n",
+    "\\sum_{m_1,m_2} |l_1m_1\\rangle|l_2m_2\\rangle \\langle l_1 m_1; l_2 m_2 | LM\\rangle = |LM\\rangle\n",
```
Luthaf, line 403:
```suggestion
    "|LM\\rangle = \\sum_{m_1,m_2} |l_1m_1\\rangle|l_2m_2\\rangle \\langle l_1 m_1; l_2 m_2 | LM\\rangle \n",
```

Luthaf (commented):
Looks good overall, I have a couple of small improvements suggestions

bindings/rascal/utils/cg_utils.py
```diff
@@ -0,0 +1,652 @@
+"""
+Utilities to manipulate CG coefficients, rotations, and equivariant construction in general.
+Requires sympy to compute relevant coefficients.
+"""
+
+import numpy as np
+from copy import deepcopy
+from scipy.spatial.transform import Rotation
```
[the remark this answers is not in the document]
ceriottm: This is *scipy* not *sympy* and so there should be no major slowdown importing it in the header

examples/equivariant_demo.ipynb
ceriottm: no, it's ZYZ. there are a gazillion conventions for the definition of Euler angles, and this is the one consistent with D matrices.

bindings/rascal/utils/cg_utils.py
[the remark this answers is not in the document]
Luthaf: right, I got confused when reading this

bindings/rascal/utils/cg_utils.py
```diff
@@ -0,0 +1,652 @@
...
+                    # computes CG coefficients
+                    ccg = _cg(l1, l2, L)
+                    self._cgraw[(l1, l2, L)] = ccg
+
+                    # applies transformations that make them act and generate real
+                    # valued coefficients
+                    if (l1 + l2 + L) % 2 == 0:
+                        rcg = np.real(
+                            np.einsum(
+                                "abc, ax, by, zc -> xyz",
```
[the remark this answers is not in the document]
Luthaf: I still this this is worth some explanation for the future readers of this bit of code

src/rascal/math/spherical_harmonics.cc
```diff
@@ -134,7 +134,7 @@ void SphericalHarmonics::compute_assoc_legendre_polynom(double cos_theta) {
 
 void SphericalHarmonics::calc(
     const Eigen::Ref<const Eigen::Vector3d> & direction,
-    bool calculate_derivatives) {
+    bool calculate_derivatives, bool conjugate) {
```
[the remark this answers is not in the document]
ceriottm: I also thought a lot about it and I thought it might be. I sense this real/complex thing is going to haunt us again, and this (as well as the realconjugate code on the python side) might be helpful to have around.

bindings/rascal/utils/cg_utils.py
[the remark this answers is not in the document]
ceriottm: I added a comment already
```
                    # applies transformations that make them act and generate real
                    # valued coefficients. this is basically achieved taking the
                    # <l1m1; l2m2| LM> "complex" CG and apply a real-to-complex to
                    # the bra (r2c) and a complex-to-real transformation to the ket
                    # (c2r). Note that since the c2r is applied to the right, there
                    # has to be a transposition (zc rather than cz in einsum)
```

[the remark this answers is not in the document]
Luthaf: Ok, I missed this!

src/rascal/math/spherical_harmonics.cc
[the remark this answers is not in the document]
Luthaf: fair enough!

examples/nice_demo.ipynb
```diff
@@ -0,0 +1,780 @@
...
+    "feat_scaling = 1e6            # just a scaling to make coefficients O(1)\n",
```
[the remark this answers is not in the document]
ceriottm: How would you do that? An additional option "scaling"? That would also help making sure it is applied consistently when you define a potential. To be done in a separate PR though IMO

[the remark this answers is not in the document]
Luthaf: I was wondering whether we could determine the required scaling analytically before computing spherical expansion, and then automatically scale everything (maybe inside the radial integral?). Otherwise, we could rescale after computing the expansion by looking at the mean absolute value, although the very large self contribution coefficient will be an issue for that. I'd rather not have the user have to specify the exact scaling value.

Then once we have a way of automatically scaling the spherical expansion, we could either do it by default or include another hyper for it.

It is definitively for a future PR though!

Luthaf (approved):
let's squash & merge once CI is happy!

PR #381, opened 2021-09-07
Don't mention unimplemented features in the documentation

Luthaf:
The Python documentation mentions `gaussian_sigma_type` "PerSpecies" and "Radial", but none of these options are implemented, which is very confusing.

@rosecers just wasted 30 min re-building librascal to try to make this work since it was documented. Overall, it is very poor practice to put non-supported features in the documentation without any sign that they’re not supported.

Does anyone knows of another "desired" feature which is documented but not implemented? I would also remove the corresponding doc.

### DIFF
--- a/bindings/rascal/representations/spherical_covariants.py
+++ b/bindings/rascal/representations/spherical_covariants.py
@@ -32,8 +32,7 @@ class SphericalCovariants(BaseIO):
 
     gaussian_sigma_type : str
         How the Gaussian atom sigmas (smearing widths) are allowed to
-        vary -- fixed ('Constant'), by species ('PerSpecies'), or by
-        distance from the central atom ('Radial').
+        vary. Only fixed smearing width ('Constant') are implemented.
 
     gaussian_sigma_constant : float
         Specifies the atomic Gaussian widths, in the case where they're
--- a/bindings/rascal/representations/spherical_expansion.py
+++ b/bindings/rascal/representations/spherical_expansion.py
@@ -32,8 +32,7 @@ class SphericalExpansion(BaseIO):
 
     gaussian_sigma_type : str
         How the Gaussian atom sigmas (smearing widths) are allowed to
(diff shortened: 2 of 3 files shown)


Luthaf:
CI failure is strange: the error is from CircleCI (`Too long with no output (exceeded 10m0s): context deadline exceeded`), and restarting the build did not fix the issue.

max-veit:
> CI failure is strange: the error is from CircleCI (`Too long with no output (exceeded 10m0s): context deadline exceeded`), and restarting the build did not fix the issue.

That is indeed strange. I've seen similar things before (hanging with no output instead of just failing), but it's usually caused by an error in the notebook, and I don't see how these changes could introduce such an error given that the previous build completed successfully. Unless another dependency was updated, and that broke things...

max-veit:
Ok, great, now the notebook validation is timing out in the main branch as well and I can't figure out why. Seems to be hung up on the tenth code cell of `zundel_i-PI.ipynb`, which constructs a `CURFilter` and invokes it on the computed to do sample selection. But neither the scikit-cosmo nor the scikit-learn versions have been updated since the last working build, and it runs just fine on my local machine (in about a minute and using around 1 MiB of RAM, so I have no idea why running it in the CI is so slow!). So I'm going to put this aside for now and work on other things, but please let me know if you find anything else.

max-veit (approved):
Ok, I'm just going to approve these changes because the build failure seems to have nothing to do with them. Let's try to figure it out in a separate issue...

Luthaf:
Ok, I'll try re-running the failing build again, and merge after that. If it still fails on master, I'll open an issue

max-veit:
It is now failing on the main branch, so no point really in rerunning: https://app.circleci.com/pipelines/github/cosmo-epfl/librascal/2704/workflows/c3842d71-9c6f-4de2-8eaa-2dda95bd6af5/jobs/23781

Issue #383, opened 2021-09-22
Issues building librascal on Mac M1


rosecers:
Hey all,

Just trying to get librascal compiled on my new laptop, ran into some issues.

Output from `mkdir build && cd build && cmake .. && make VERBOSE=1`:
```
-- Build type is: Release
-- The CXX compiler identification is AppleClang 13.0.0.13000029
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /Library/Developer/CommandLineTools/usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Downloading wigxjpf 
-- Found PythonInterp: /opt/miniconda3/bin/python3 (found suitable version "3.9.5", minimum required is "3") 
-- Found PythonLibs: /opt/miniconda3/lib/libpython3.9.dylib
-- Performing Test HAS_FLTO
-- Performing Test HAS_FLTO - Success
-- Performing Test HAS_FLTO_THIN
-- Performing Test HAS_FLTO_THIN - Success
-- Installation ROOT: /opt/miniconda3
-- Warning: this does not have an effect - use NO_SYSTEM_FROM_IMPORTED if using imported targets
-- cpplint parser: /opt/miniconda3/bin/cpplint
-- clang-format not found, so the pretty-cpp target is unavailable
-- BLACK auto formatter: /opt/miniconda3/bin/black
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/rca/source_installs/librascal/build
/opt/miniconda3/bin/cmake -S/Users/rca/source_installs/librascal -B/Users/rca/source_installs/librascal/build --check-build-system CMakeFiles/Makefile.cmake 0
/opt/miniconda3/bin/cmake -E cmake_progress_start /Users/rca/source_installs/librascal/build/CMakeFiles /Users/rca/source_installs/librascal/build//CMakeFiles/progress.marks
/Library/Developer/CommandLineTools/usr/bin/make  -f CMakeFiles/Makefile2 all
/Library/Developer/CommandLineTools/usr/bin/make  -f src/CMakeFiles/rascal.dir/build.make src/CMakeFiles/rascal.dir/depend
[...]

Issue #335, opened 2021-04-08
Memory corruption related to march=native


Luthaf:
The use of `-march=native` **by default**  is not a good idea:
https://github.com/cosmo-epfl/librascal/blob/master/CMakeLists.txt#L94-L100

It can corrupt memory when using librascal as an external project. For example, compiling and running `examples/spherical_invariants_example.cc` externally with 

```bash
clang++ -std=c++14 -O3 spherical_invariants_example.cc -o spherical_invariants_example -I../src -I/usr/local/include/eigen3 -I../build/external/wigxjpf/inc ../build/src/librascal.a ../build/external/wigxjpf/lib/libwigxjpf.a
./spherical_invariants_example ../reference_data/inputs/molecular_crystal.json
```

Results in
```
spherical_invariants_example(79937,0x1146bc5c0) malloc: *** error for object 0x7ff1ba702200: pointer being freed was not allocated
spherical_invariants_example(79937,0x1146bc5c0) malloc: *** set a breakpoint in malloc_error_break to debug
zsh: abort      ./spherical_invariants_example ../reference_data/inputs/molecular_crystal.jso
```

Adding `-march=native` to the above command line fixes the issue and everything works. 

The origin of the problem is that some code in librascal.a/librascal.so contains code assuming one instruction set but the main binary assumes a separate instruction set. Different instructions set can have different ABI, and this results in caller/calle code not agreeing on where arguments should be passed. This is partially exacerbated by the fact that librascal is mostly a header library, meaning the end user will compile most of the code themselves, meaning they **MUST** match the exact compiler flags used to build librascal.

----

 Use of `-march=native` by default can also be a problem on some HPC cluster, where the head and compute nodes use different instruction sets (e.g. Broadwell vs Haswell for Intel), leading to binaries that run on the head node and crash on the compute nodes.

----

Overall, I think we should remove this default of `-march=native`, and evaluate the performance difference from doing so. If there is a large performance difference, then we can add suggestions in the documentation that people use if when building librascal & their own code.

felixmusil:
I am not quite sure how to label this issue but it certainly is not a bug, merely something that might hinder portability. I would keep this flag by default unless it is clear that it does not hit performances too much, especially on CPUs that have avx-512 instructions.

>Use of -march=native by default can also be a problem on some HPC cluster, where the head and compute nodes use different instruction sets (e.g. Broadwell vs Haswell for Intel), leading to binaries that run on the head node and crash on the compute nodes.

[...]

PR #384, opened 2021-10-01
Improve apple M1 support

felixmusil:
As referenced in #383, setting explicitly `-march` in the compilation arguments has yet to be supported by clang.
Thanks to CMake, for the specific case of arm64 we can build universal binaries which work on my machine. I made the change so that it does not affect the build settings on other architectures.

We don't have a test in the CI for the arm64 architecture so could you check that it works on your machines ? @rosecers 
Also, I am using clang from homebrew (built for arm64) so my current python does not use roseta.

Note that llvm 12 (current stable version) does not have arm64 as a target architecture even though it has apple-a14 which is basically the same). It seems that it should exist in llvm 13.


CMakeLists.txt
```diff
@@ -90,14 +90,27 @@ if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" OR
...
+if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+  set(CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE} -march=${TYPE_ARCHITECTURE}")
+  set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -march=${TYPE_ARCHITECTURE}")
+  set(CMAKE_C_FLAGS_RELWITHDEBINFO "${CMAKE_C_FLAGS_RELWITHDEBINFO} -march=${TYPE_ARCHITECTURE}")
+    set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO} -march=${TYPE_ARCHITECTURE}")
+elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR
+  "${CMAKE_CXX_COMPILER_ID}" STREQUAL "AppleClang")
+  # detect if the local architechture is arm64 or x86_64
+  if(CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL "arm64")
+    set(CMAKE_OSX_ARCHITECTURES "x86_64;arm64" CACHE STRING "")
```
Luthaf, line 106: Why build a universal binary instead of detecting which architecture we actually need?

Luthaf (commented):
I'll leave it to @rosecers to test if this fixes her problem, I have a small question in the meantime

CMakeLists.txt
felixmusil: I thought it would be needed to handle the case where you have a python distribution built with `x86_64` while working on an `arm64`. I need @rosecers input to check that it's actually working though.

felixmusil:
From the CI, it seems that compilation/runtime increased twofold causing the valgrind workflow to timeout. 
The weird aspect is that it happens also for the gcc-10 build (I compared the timings with the build timings from another PR) which should not be affected by my changes. I also compared the compilation arguments in these 2 cases and, as expected, they are identical. Except for the underlying hardware, what could cause such increase in the overall execution time of the workflow ?

CMakeLists.txt
Luthaf: My understanding is that if Python is an `x86_64` binary, it will only need `x86_64` libraries. Rosetta 2 will then deal with the emulation of x86 on ARM. The main issue with building universal binary is the increased compile time (and file size when we upload this to PyPI). 

If there is an easy way to detect the required architecture insead of doing both that would be nice, otherwise we'll survive.

felixmusil: I could not find one.

felixmusil:
Ok then there is no point compiling for x86_64 then. Could you try to setup a native python environment and check that it works as it does on my machine ?

Luthaf:
> Ok then there is no point compiling for x86_64 then.

Why? If the Python install is x86_64, we want to compile for that. For me the issue is also about `-march=native` when doing cross-compilation to x86_64

felixmusil:
> Why? If the Python install is x86_64, we want to compile for that.

From Rose's test it seems that the attempt to compile for both x86_64 and arm64 on arm64 failed.
 But it's true that it might also be the detection mechanism of arm64 that failed. This could be checked by looking into the cmake setup variables.
 Now that you mention it, it might also be that the cmake checks for arm64 fail catch it when using a Rosetta 2 install and then it falls back onto the native argument. The question is why it detects 'vortex' rather than something more generic?

> For me the issue is also about -march=native when doing cross-compilation to x86_64

The arch argument is set to native as default but can be very easily overiden at the cmake level so I am not sure to understand your concern.

Luthaf:
> The question is why it detects 'vortex' rather than something more generic?

Because `-march=native` asks for the most specific arch possible? From https://en.wikipedia.org/wiki/Comparison_of_ARMv8-A_cores, looks like Vortex is the name of one of Apple architecture. I more puzzled about why this is not selecting Firestorm/Icestorm instead, since these are the architectures used in the M1 CPU and would be the most specific ones. Maybe Vortex is the common base between firestorm & icestorm?

> The arch argument is set to native as default but can be very easily overiden at the cmake level so I am not sure to understand your concern.

My concern is 100% about the default value. By default, everything should work for everyone, and `-march=native` breaks here and in other ways (#335). I think we should have a more conservative default, while still allowing advanced users to build with something else.

Luthaf:
Fixed in #403 for now

PR #387, opened 2021-10-21
Feat/kvec generator

kvhuguenin:
Implemented class for generating the k-vectors (reciprocal space vectors) used for the k-space implementation of LODE and more general spherical expansion coefficients.
The details and derivations of the algorithm used to generate the vectors is described in a supplementary document that can be provided if desired. 

Tests are provided to check that the code behaves reasonably under realistic conditions as well as some more artificial tests to catch boundary cases.

### DIFF
--- a/src/rascal/math/kvec_generator.hh
+++ b/src/rascal/math/kvec_generator.hh
@@ -1,0 +1,137 @@
+/**
+ * @file   kvec_generator.cc
+ *
+ * @author  Andrea Grisafi <andrea.grisafi@epfl.ch>
+ * @author  Kevin Kazuki Huguenin-Dumittan <kevin.huguenin-dumittan@epfl.ch>
+ *
+ * @date   10 February 2021
+ *
+ * @brief Generate the k-vectors (also called reciprocal or Fourier vectors)
+ *        needed for the k-space implementation of LODE and SOAP.
+ *        More specifically, these are all points of a (reciprocal space)
+ *        lattice that lie within a ball of a specified cutoff radius.
+ *
+ * Copyright  2021  Andrea Grisafi, Kevin Kazuki Huguenin-Dumittan, COSMO(EPFL)
+ *
+ * Rascal is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation, either version 3, or (at
+ * your option) any later version.
+ *
+ * Rascal is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this software; see the file LICENSE. If not, write to the
+ * Free Software Foundation, Inc., 59 Temple Place - Suite 330,
+ * Boston, MA 02111-1307, USA.
+ */
+
+#ifndef SRC_RASCAL_MATH_KVEC_GENERATOR_HH_
+#define SRC_RASCAL_MATH_KVEC_GENERATOR_HH_
+
+#include "rascal/math/utils.hh"
+
+#include <Eigen/Dense>
+
+namespace rascal {
+  namespace math {
+    /**
+     * Produce the k-vectors (also called reciprocal space or Fourier
+     * vectors) needed for the k-space implementation of LODE and SOAP.
+     *
+     * More specifically, if b1, b2 and b3 are basis vectors of a reciprocal
+     * lattice, this class is used to get all vectors of the form:
+     * \f{equation}{k = n_1b_1 + n_2b_2 + n_3b_3,\f}
+     * where n1,n2,n3 are integers, which are inside a ball of a given
+     * cutoff radius \f$k_\mathrm{cut}\f$.
+     *
+     * Note that for any radius > 0, the origin \f$k=(0,0,0)\f$ is always
+     * in the ball and is the first entry of the returned vectors.
+     * For applications where k=0 should be excluded, there is the option
+     * to do so.
+     * For any other k-vector k, symmetry typically allows us to group the
+     * two vectors k and -k together. Thus, only one out of each such pair
+     * is returned as a representant.
+     *
+     */
+    class Kvectors {
+     private:
+      /* Store the quantities provided by the user that completely determine
+       * the problem:
+       * The first two quantities, the cutoff radius in reciprocal space as
+       * well as the three basis vectors of the reciprocal cell, determine
+       * the geometry, and thus are most important.
+       * The optional parameter include_origin is stored as well which
+       * determines whether the point (0,0,0) should be included or not.
+       */
+      double kcutoff{};
+      Eigen::Matrix3d basisvecs{};
+      bool include_origin{};
+
+      /* Variables for quantities that can be returned by this class:
+       * - the N x 3 Eigen matrix containing the k-vectors in the ball
+       * - the Eigen vector of size N containing the norms of the k-vectors
+       */
+      Matrix_t kvectors{};
+      Vector_t kvector_norms{};
+
+     public:
+      /** Constructor:
+       * Upon initialization, directly compute all the k-vectors on the
+       * lattice specified by three basisvectors lying within a ball.
+       * @param cutoff Cutoff radius of reciprocal space ball.
+       *     As a guideline, if the smearing parameter of the density
+       *     is given by \f$sigma\f$, use \f$pi/sigma\f$ as cutoff.
+       * @param basisvectors 3x3 matrix containing the three basis
+       *     vectors a1, a2, a3 of the real space unit cell, stored as:
+       *     a1 = basisvectors[0], a2=basisvectors[1], etc.
+       *     These are then automatically converted to the corresponding
+       *     reciprocal cell vectors b1, b2, b3 by the formulae:
+       *     \f{equation}{b_1 = 2\pi \frac{a_1 \times a_2}
+       *                 {a_1 \cdot (a_2 \times a_3)} \f}
+       *     and similarly for the periodic permutations.
+       *     It is also possible to provide the basis vectors of the
+       *     reciprocal lattice directly by using the option below.
+       * @param is_reciprocal_cell false by default.
+       *     If set to true, it indicates that the cell provided
+       *     in the variable basisvectors is already the reciprocal cell.
+       *     Thus, the formula above for converting the a_i's to the b_i's
+       *     is not used, and the provided cell is used directly.
+       * @param need_origin false by default.
+       *     If set to true, the list of found k-vectors will also
+       *     contain the origin k=(0,0,0), which will then be stored
+       *     as the very first element in the matrix of found vectors.
+       */
+      Kvectors(double cutoff, Eigen::Matrix3d basisvectors,
+               bool is_reciprocal_cell = false, bool need_origin = false);
+
+      /** Returns number of vectors found within cutoff radius
+       * without double counting pairs related by inversion.
+       * This is the actual number that can be used to iterate over all
+       * vectors and its norms obtained from get_kvectors and get_norms.
+       */
+      size_t get_numvectors() const { return this->kvector_norms.size(); }
+
+      /** Returns reference to the Eigen matrix containing all
+       * k-vectors within cutoff radius without double counting, where
+       * row(i) = i-th vector. If the origin k=(0,0,0) is included, it is
+       * always stored in row(0).
+       */
+      Eigen::Ref<Matrix_t> get_kvectors() {
+        return Eigen::Ref<Matrix_t>(this->kvectors);
+      }
+
+      /** Returns const reference to the Eigen vector containing the norm
+       * of all k-vectors within the cutoff without double counting.
+       */
+      Eigen::Ref<const Vector_t> const get_kvector_norms() {
+        return Eigen::Ref<const Vector_t>(this->kvector_norms);
+      }
+    };  // Kvectors class
+  }     // namespace math
+}  // namespace rascal
+
+#endif  // SRC_RASCAL_MATH_KVEC_GENERATOR_HH_
--- a/tests/test_math.hh
+++ b/tests/test_math.hh
@@ -458,6 +458,49 @@ namespace rascal {
     }    // for (auto inputs...) (function inputs)
   }
 
+  /**
+   * @brief Fixture used for testing kvecgenerator
+   *
+   * Multiple cells are defined that cover a range of relevant
+   * structures from the simple cubic structure having highest
+   * symmetry to triclinic cells whose extension along the three
+   * basis vectors is very different.
+   * The tests are then run over all these structures to ensure
+   * that the kvecgenerator works properly under various conditions.
+   */
+  struct KvecgenRefFixture {
+    KvecgenRefFixture() {
+      // Define basic cell quantities
+      Eigen::Matrix3d cell{};
+      double a = 10.;
+      double c = std::sqrt(8. / 3.) * a;
+
+      // Cells 1 and 2: Simple cubic cells of different dimensions
+      cell << 1., 0., 0., 0., 1., 0., 0., 0., 1.;
+      this->cells.push_back(a * cell);
+      this->cells.push_back(1.5 * a * cell);
+
+      // Cell 3: hcp cell (taken from test_structure.hh)
+      cell << a, -0.5 * a, 0., 0., std::sqrt(3.) / 2. * a, 0., 0., 0., c;
+      this->cells.push_back(cell);
+
+      // Cell 4: fcc cell (taken from test_structure.hh)
+      cell << a, 0.5 * a, 0.5 * a, 0., 0.5 * a, 0., 0., 0., 0.5 * a;
+      this->cells.push_back(cell);
+
+      // Cells 5 and 6: Two triclinic cells including high asymmetry
+      cell << 1., 0.3, 0.2, 0.4, 1., -0.1, 0., 0.3, 1.;
+      this->cells.push_back(a * cell);
+      cell << 0.8, 0.3, 0.2, 0.1, 3., -0.2, 0., 0.3, 0.7;
+      this->cells.push_back(a * cell);
+    }
+
+    ~KvecgenRefFixture() = default;
+
+    // Variable to store generated cells
+    std::vector<Eigen::Matrix3d> cells{};
+  };
+
   struct Hyp1F1RefFixture {
     Hyp1F1RefFixture() {
       this->ref_data =
--- a/tests/test_math_kvecgenerator.cc
+++ b/tests/test_math_kvecgenerator.cc
@@ -1,0 +1,337 @@
+/**
+ * @file   test_kvecgenerator.cc
+ *
+ * @author Kevin Kazuki Huguenin-Dumittan <kevin.huguenin-dumittan@epfl.ch>
+ *
+ * @date   2021
+ *
+ * @brief Test the class used to generate the k-vectors, also called
+ *        reciprocal space vectors, used in the k-space implementation
+ *        of the spherical expansion coefficients of SOAP and LODE
+ *
+ * Copyright  2019  Felix Musil, COSMO (EPFL), LAMMM (EPFL)
+ *
+ * Rascal is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation, either version 3, or (at
+ * your option) any later version.
+ *
+ * Rascal is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this software; see the file LICENSE. If not, write to the
+ * Free Software Foundation, Inc., 59 Temple Place - Suite 330,
+ * Boston, MA 02111-1307, USA.
+ */
+
+#include "test_math.hh"
+
+#include "rascal/math/kvec_generator.hh"
+
+#include <boost/test/unit_test.hpp>
+
+#include <Eigen/Eigenvalues>
+
+#include <cmath>
+
+namespace rascal {
+
+  BOOST_AUTO_TEST_SUITE(test_kvecgenerator);
+
+  /* ---------------------------------------------------------------------- */
+  /**
+   * Test that the kvecgenerator produces the desired vectors for small
+   * cells, where we can explicitly list the correct vectors
+   */
+  BOOST_AUTO_TEST_CASE(kvecgen_smallcells_test) {
+    // Define reciprocal space cell to be used in the following tests
+    // We use the identity matrix to avoid floating point errors
+    Eigen::Matrix3d reciprocal_cell;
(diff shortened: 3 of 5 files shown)


tests/test_math.hh
```diff
@@ -458,6 +458,50 @@ namespace rascal {
...
+      this->cells.push_back(cell);
+
+      // Cells 5 and 6: Two triclinic cells
+      cell << 1., 0.3, 0.2, 0.4, 1., -0.1, 0., 0.3, 1.;
+      this->cells.push_back(a * cell);
+      cell << 0.8, 0.3, 0.2, 0.1, 3., -0.2, 0., 0.3, 0.7;
+      this->cells.push_back(a * cell);
+    }
+
+    // Destructori
```
Luthaf, line 498:
```suggestion
    // Destructor
```

or event better 

```suggestion
```

tests/test_math_kvecgenerator.cc
```diff
@@ -0,0 +1,280 @@
...
+    // For small enough cutoff radii, namely for the intervals
+    // [1,sqrt(2)), [sqrt(2), sqrt(3)), [sqrt(3),2) we can explicitly
+    // list all vector that should be found by the algorithm.
+    // Store these solutions in the correct order for each cutoff
+    Eigen::MatrixXd kvec_sol0(3, 3), kvec_sol1(9, 3), kvec_sol2(13, 3);
+    kvec_sol0 << v001, v010, v100;
+    kvec_sol1 << v001, v01m, v010, v011, v1m0, v10m, v100, v101, v110;
+    kvec_sol2 << v001, v01m, v010, v011, v1mm, v1m0, v1m1, v10m, v100, v101,
+        v11m, v110, v111;
+    Eigen::MatrixXd * solutions[3] = {&kvec_sol0, &kvec_sol1, &kvec_sol2};
```
Luthaf, line 79:
```suggestion
    Eigen::MatrixXd kvec_expected_0(3, 3);
    kvec_expected_0 << v001, v010, v100;
    
    Eigen::MatrixXd kvec_expected_1(9, 3);
    kvec_expected_1 << v001, v01m, v010, v011, v1m0, v10m, v100, v101, v110;
    
    Eigen::MatrixXd kvec_expected_2(13, 3);
    kvec_expected_2 << v001, v01m, v010, v011, v1mm, v1m0, v1m1, v10m, v100, v101,
        v11m, v110, v111;
        
    Eigen::MatrixXd * expected[3] = {&kvec_expected_0, &kvec_expected_1, &kvec_expected_2};
```

tests/test_math_kvecgenerator.cc
```diff
@@ -0,0 +1,280 @@
...
+
+    // For various cutoff radii close to the thresholds after which
+    // we would get further vectors, we compare the obtained vectors
+    // to the correct solution. The variable i runs over the three intervals
+    double cutofflims[4] = {1., math::SQRT_TWO, math::SQRT_THREE, 2.};
```
Luthaf, line 84: You should use underscores to separate words in your variables names, it makes it much easier to read the code

```suggestion
    double cutoff_lims[4] = {1., math::SQRT_TWO, math::SQRT_THREE, 2.};
```

tests/test_math_kvecgenerator.cc
```diff
@@ -0,0 +1,280 @@
...
+    double cutoffeps = 1e-12;
```
Luthaf, line 85:
```suggestion
    double cutoff_eps = 1e-12;
```

tests/test_math_kvecgenerator.cc
```diff
@@ -0,0 +1,280 @@
...
+    for (int i = 0; i < 3; i++) {
+      math::Kvectors k_vectorsmin(cutofflims[i] + cutoffeps, reciprocalcell,
```
Luthaf, line 87:
```suggestion
      math::Kvectors k_vectors_min(cutoff_lims[i] + cutoff_eps, reciprocal_cell,
```

tests/test_math_kvecgenerator.cc
```diff
@@ -0,0 +1,280 @@
...
+                                  true);
+      BOOST_CHECK_EQUAL(k_vectorsmin.get_kvectors(), *solutions[i]);
+      math::Kvectors k_vectorsmax(cutofflims[i + 1] - cutoffeps, reciprocalcell,
+                                  true);
+      BOOST_CHECK_EQUAL(k_vectorsmax.get_kvectors(), *solutions[i]);
+    }
+
+    // Check whether the code also works if we use a different, highly
+    // distorted basis to describe the same lattice
+    Eigen::Matrix3d cell_distortedbasis;
```
Luthaf, line 97:
```suggestion
    Eigen::Matrix3d cell_distorted_basis;
```

tests/test_math_kvecgenerator.cc
```diff
@@ -0,0 +1,280 @@
...
+    double cutoff = 6.28;
+    for (auto & cell : this->cells) {
+      math::Kvectors kvectors(cutoff, cell);
+      int nvec = kvectors.get_numvectors();
+      Eigen::MatrixXd kvecs = kvectors.get_kvectors();
+      Eigen::VectorXd kvals = kvectors.get_norms();
+
+      // Make sure that all vector do indeed lie within
+      // the cutoff and that the lengths computed by the kvecgenerator
+      // agree with those values.
```
Luthaf, line 137: This comment is duplicated from the `@brief` above, no need to have it around twice

tests/test_math_kvecgenerator.cc
```diff
@@ -0,0 +1,280 @@
...
+    for (auto & cell : this->cells) {
+      math::Kvectors kvectors(cutoff, cell, false);
+      int nvec = kvectors.get_numvectors();
+      Eigen::MatrixXd kvecs = kvectors.get_kvectors();
+      Eigen::VectorXd kvals = kvectors.get_norms();
+
+      /* Define the used rotation matrices. In order to avoid
+       * discrepancies due to rounding errors, we only use rotations
+       * that are a combination of 90 degrees rotations, since these
+       * rotation matrices only contain 1, 0, and -1 as components.
```
Luthaf, line 240: It might be more interesting to use arbitrary rotation, and check for approximate equality using epsilon ~1e-12. Such test would be able to catch more issues.

src/rascal/math/kvec_generator.hh
```diff
@@ -0,0 +1,229 @@
...
+       * These are the quantities that will be used by LODE and the
+       * reciprocal space implementation of SOAP. They are:
+       * - the total number numvectors of vectors within the cutoff,
+       * - the numvectors times 3 Eigen matrix containing those vectors,
+       * - the Eigen vector of size numvectors containing the norms
+       *   of the found vectors.
+       */
+      size_t numvectors = 0;
+      Matrix_t kvecs{};
+      Vector_t kvecnorms{};
```
Luthaf, line 81: Instead of duplicating the information on the number of vectors, you could remove the `numvectors` member, and use `kvecsnorms.size()` to implement `get_numvectors()`.

src/rascal/math/kvec_generator.hh
```diff
@@ -0,0 +1,229 @@
...
+      // Eigen::Matrix_Xi kvecindices{};
```
Luthaf, line 82: leftover code?

src/rascal/math/kvec_generator.hh
```diff
@@ -0,0 +1,229 @@
...
+
+      // Auxiliary variables for developers/testing:
```
Luthaf, line 84: Can you add a comment for each one of these indicating what they contain/represent?

src/rascal/math/kvec_generator.hh
```diff
@@ -0,0 +1,229 @@
...
+        // Run the precompute function described below
+        this->precompute();
+      }  // end constructor
+
+      /**
+       * Using the provided parameters in the constructor, find
+       * all the k-vectors within the cutoff radius and store them
+       * as well as derived quantities that can be returned to the user.
+       */
+      void precompute();
```
Luthaf, line 149: this function is never supposed to be called by the user right? Then it should be at least made private, or even better removed and only be part of the constructor.

src/rascal/math/kvec_generator.hh
```diff
@@ -0,0 +1,229 @@
...
+       *     If set to true, it indicates that the cell provided
+       *     in the variable basisvectors is already the reciprocal cell.
+       *     Thus, the formula above for converting the a_i's to the b_i's
+       *     is not used, and the provided cell is used directly.
+       * @param need_origin false by default.
+       *     If set to true, the list of found k-vectors will also
+       *     contain the origin k=(0,0,0), which will then be stored
+       *     as the very first element in the matrix of found vectors.
+       */
+      explicit Kvectors(double cutoff, Eigen::Matrix3d basisvectors,
```
Luthaf, line 121: you don't need an `explicit` here

src/rascal/math/kvec_generator.hh
```diff
@@ -0,0 +1,229 @@
...
+                        bool is_reciprocal_cell = false,
+                        bool need_origin = false) {
+        // Set internal variables apart from cell
+        this->kcutoff = cutoff;
+        this->include_origin = need_origin;
+
+        /* Store provided basis vectors into the internal variable basisvecs.
```
Luthaf, line 128: the constructor implementation should be moved to the .cc file, to improve compile time

src/rascal/math/kvec_generator.hh
```diff
@@ -0,0 +1,229 @@
...
+       * all k-vectors within cutoff radius without double counting,
+       * where row(i) = i-th vector. If the origin k=(0,0,0) is included,
+       * it is always stored in row(0).
+       */
+      Matrix_t get_kvectors() const { return this->kvecs; }
+
+      /** Returns the Eigen vector of length this->numvectors containing
+       * the norm of all k-vectors within the cutoff without double counting.
+       */
+      Vector_t get_norms() const { return this->kvecnorms; }
```
Luthaf, line 183: I'm not familiar with Eigen enough, but this code looks like it is making a copy. Could we write these two functions to return const references instead?

src/rascal/math/kvec_generator.hh
```diff
@@ -0,0 +1,229 @@
...
+
+      // FUNCTIONS FOR DEVELOPERS / TESTING
```
Luthaf, line 185: If these are not going to be used by the rest of the code and were only useful when developing this class, they should be removed to decrease the amount of code to maintain.

src/rascal/math/kvec_generator.hh
```diff
@@ -0,0 +1,229 @@
...
+        return 2. * M_PI / 3. * kcut * kcut * kcut / sqrt(this->detM);
+      }
+
+      /**
+       * Return an auxilary factor used for tests.
+       * The estimate above obtained from get_numvectors_estimate() clearly
+       * cannot be exactly equal to the actual number of obtained k-vectors.
+       * There however exist rigorous bounds for the deviation between the
+       * estimate and the true value, which we check in the tests.
+       * The details can be found in /tests/test_math_kvec_generators.cc.
```
Luthaf, line 206: if this is used in tests, this functions should be moved there.

src/rascal/math/kvec_generator.hh
agoscinski: As far as I remember one can do this as in
https://github.com/cosmo-epfl/librascal/blob/master/src/rascal/representations/calculator_spherical_expansion.hh#L1368
https://github.com/cosmo-epfl/librascal/blob/master/src/rascal/representations/calculator_spherical_expansion.hh#L1374

using definitions
https://github.com/cosmo-epfl/librascal/blob/master/src/rascal/representations/calculator_spherical_expansion.hh#L222

tests/test_math_kvecgenerator.cc
agoscinski: I haven't seen any orthogonal matrix generator from Eigen, I think you could do

srand to set a fixed seed (as test hyperparameter, might need to include stdlib.h)
(according to eigen mail list https://listengine.tuxfamily.org/lists.tuxfamily.org/eigen/2010/07/msg00403.html)
and use Eigen's for each rotation rotation.setRandom() to make it random and use QR decomposition to get an orthogonal matrix as in
https://eigen.tuxfamily.org/dox/classEigen_1_1HouseholderQR.html#a3e8e56769bbaaed0616ad98c4ff99c7b

src/rascal/math/kvec_generator.hh
kvhuguenin: The norms will actually come in very handy during LODE, so this function has to stay, unless you would propose to recompute all the norms every time. Personally, I had the impression that recomputing the norm for every single vector would be a waste of time vs. the added cost of storing these separately, since the norms are computed during the generation of the vectors anyway.

tests/test_math_kvecgenerator.cc
kvhuguenin: Ok, so this test is a little more tricky than that. Before commit #9c310f3 , what @Luthaf suggests is what I had done as well , namely by explicitly generating some rotation matrices. However, if you want to perform the rotation test on realistic cells, you get into serious rounding error issues.
Most importantly, if you apply a rotation matrix, the number of k-vectors that are found within the cutoff is often not the same any more due to rounding errors. In that case, it makes no sense to check the equality of each pair of vectors, since there will be many vectors that only exist in the rotated cell or vice versa, and checking whether one vector was "left out" would also be a huge pain to make sure you don't fail due to edge cases.

There are some possible solutions to this problem. But before we call it a "problem" too much, I would like to remind you that we typically choose the cutoff to be so large that we don't really care about what happens at the boundary. So if some vector that have length "equal to" kcut aren't found due to these rounding errors, it doesn't really matter for applications.

This is why in commit #9c310f3, I have decided to change the rotation test to something numerically stabler, by choosing rotation matrices that only contain +1, -1 and 0's and thus preserve floating point arithmetic. (this is why in the scaling test, I only use factors of 0.5 and 2, by the way).
With the original rotation test with arbitrary matrices, I had a hand picked set of cells that worked out fine numerically, but I thought that this is not a good long term solution, if the failure of the tests depends too much on the floating point error you get from the specific implementation details (and I had the rotation test fail on me before because I changed something little).

kvhuguenin: TLDR: Right now, we have a rotation test that is very stringent, i.e. it tests for essentially perfect equality between the results before and after rotation. Such tests are numerically challenging unless you use rotation matrices that do not affect floating point errors, which limits us to compositions of 90 degree rotations in xyz directions.

I did have some more alternatives in mind when I made these changes, which would come in the form of some additional tests for arbitrary rotations. If you want, I could add those. But given that the test file is already longer than both the kvec_generator.cc and .hh files (potentially longer than both of them combined once I move some of the functionality of the .hh file into the tests as suggested in this review), I'm not sure whether this would be more beneficial relative to the amount of extra code to maintain, since we have already 5 tests for a simple class that performs exactly one job.

agoscinski: Does this problem occur for points exactly on the cutoff? Shouldn't this be fixable by adding a bit of numerical jitter to the cutoff? You use a cutoff which is a multiple of the cell parameters, only in such a case I can see that the double precision error of 1e-16 could change the number of k-vectors. What do you think

kvhuguenin: It occurs for points close to the cutoff, yes. Back when I still had that in the test, I mostly used a cutoff that was not an integer multiple of the cell parameters, namely 6.28, and I still had these issues.
In principle, it wouldn't be difficult to add jitters to the cutoff until all tests pass, but that was exactly the kind of "fine-tuning" I was trying to avoid by rewriting the tests in the current form, since the tests suddenly started to fail when I changed something else in the code, although I don't recall which cell caused that.

src/rascal/math/kvec_generator.hh
```diff
@@ -0,0 +1,229 @@
...
+       * Return the expected number of obtained vectors based on
+       * continuous approximation, which is simply the volume of a ball
+       * of radius kcut divided by the reciprocal space unit cell.
+       * In practice, we have to divide by 2 since we are only generating
+       * half of the k-vectors, which is why we use 2pi/3 rather than 4pi/3.
+       * The volume of the unit cell in reciprocal space is sqrt(detM).
+       */
+      double get_numvectors_estimate() const {
+        const double kcut = this->kcutoff;
+        return 2. * M_PI / 3. * kcut * kcut * kcut / sqrt(this->detM);
```
agoscinski, line 197: I don't know why, but we have defined our own pi constant in math/utils.hh I would just use it here for consistency within the project

src/rascal/math/kvec_generator.hh
```diff
@@ -0,0 +1,229 @@
...
+        this->include_origin = need_origin;
+
+        /* Store provided basis vectors into the internal variable basisvecs.
+         * We need to distinguish whether the provided vectors are already
+         * reciprocal cell vectors or need to be converted first
+         */
+        if (not is_reciprocal_cell) {  // real space cell is given
+          // Create reciprocal cell from real space cell
+          Eigen::Matrix3d tcell = basisvectors.transpose();
+          this->basisvecs = 2.0 * M_PI * tcell.inverse();
```
agoscinski, line 135: I don't know why, but we have defined our own pi constant in math/utils.hh I would just use it here for consistency within the project

tests/test_math_kvecgenerator.cc
agoscinski: But don't you find it weird: One rotation can only introduce an error in the range 1e-16 (maybe 1e-15), so the k vector count can only change when a k vector is exactly on the cutoff. The probability that this happens is for a random cell should be super low. The basis vector are multiplied by pi so the basis vectors are not a multiple of your cutoffs. I don't understand how the count can change

kvhuguenin: Yes, I agree. That's why I had printed all the vectors to see whether that was really the issue, and at least then, it was. I probably could have just shifted the cutoff by 1e-10 or so to get rid of that issue. The way I was generating the rotation matrices had huge potential for rounding errors, and combined with the loop over n3 that would accumulate those, I believe that errors on the order of 1e-14 might be realistic actually, but not much more.

kvhuguenin:
Thank you for all the suggestions. I have now pushed a version that has implemented all these changes. The largest two of which are:
  - all functions in the .hh file used only for tests was moved to the "kvecgen_numvectors_reasonable_test". The class now only contains the constructor + the three functions that will be called by LODE in its current implementation. It really is true that the class looks much cleaner now, so thank you @Luthaf for the suggestion!
  - since we had some discussions regarding the rotation test, I have now made a new test using arbitrary rotation matrices by the method suggested by @agoscinski . When I ran the tests, I found that the sup/max/infinity-norm between the original and rotated vectors obtained could be as large as 4e-14 with a relatively small cutoff, so this combined with a very numerical error-prone way to implement rotation matrices originally might have led to my "bad luck" of failing rotation tests in the past.

kvhuguenin:
Right now, for some reason I don't yet understand, the tests compiled with the clang compiler (that's what the clang-n tests are, right?) produce errors which seem to be due to the fact that the "rotation" matrices that are produced by the algorithm are not actual rotation matrices, i.e. they do not satisfy R^TR = RR^T=1. On my machine, the test runs fine. Is anybody familiar with this isse?

agoscinski:
I am looking into this now

tests/test_math_kvecgenerator.cc
```diff
@@ -0,0 +1,339 @@
...
+    cell0 << 2., 0., 0., 0., 1., 0., 0., 0., 3.;
+    cells.push_back(cell0);
+
+    // Generate rotation matrices with fixed seed
+    srand(4653056);
+    size_t num_random = 100;
+    std::vector<Eigen::Matrix3d> rotation_matrices{};
+    for (size_t index = 0; index < num_random; index++) {
+      auto random_matrix = Eigen::Matrix3d::Random();
+      const auto rotation_matrix_0 =
```
agoscinski, line 295:
```suggestion
      Eigen::Matrix3d rotation_matrix_0 =
```
I think auto makes the type `const Eigen::HouseholderSequence<Eigen::Matrix<double, 3, 3, 0, 3, 3>, Eigen::Matrix<double, 3, 1, 0, 3, 1>, 1>`
I am not sure what then in the multiplication later happens, but it is probably not a matrix multiplication

Because auto is interpreted differently interpreted from interpret compiler, it can work on your machine probably using a gnu compiler.

kvhuguenin:
Oh nice! Thank you for the quick fix and lesseon @agoscinski ! I thought that using auto more might increase readability, but I should be more careful in the future then.
Anyway, since this is the first time my code gets reviewed by somebody who actually knows how to do this, let me know if there is anything else that can be improved, both for this code specifically and also in terms of general programming practices @Luthaf and @agoscinski .

tests/test_math_kvecgenerator.cc
Luthaf: > Because auto is interpreted differently from different compiler, it can work on your machine using a gnu compiler while failing with lang

That's false. `auto` deduction rules are defined by the standard (they are the same rules applied during template substitution) and the same over different compilers (or this is a compiler bug). What might be happening here is that Eigen is doing strange things with templates, which auto take as they are; and using an explicit type forces a cast & evaluate the Eigen template. Overall, that's an Eigen issue more than a compiler/auto issue.

src/rascal/math/kvec_generator.hh
Luthaf: I was talking about the functions below, in `// FUNCTIONS FOR DEVELOPERS / TESTING`. I agree that the norm makes sense!

src/rascal/math/kvec_generator.hh
```diff
@@ -0,0 +1,137 @@
...
+      Matrix_t get_kvectors() const {
+        return Eigen::Ref<const Matrix_t>(this->kvectors);
+      }
+
+      /** Returns const reference to the Eigen vector containing the norm
+       * of all k-vectors within the cutoff without double counting.
+       */
+      Vector_t get_kvector_norms() const {
+        return Eigen::Ref<const Vector_t>(this->kvector_norms);
+      }
```
Luthaf, line 132: You need to also change the return type of these functions. Here you are creating a const reference, which is then converted to a `Matrix_t`/`Vector_t` value, making a copy.

tests/test_math_kvecgenerator.cc
Luthaf: Still the case 😃!

Luthaf (commented):
I have two small nitpicks; and then this should be good to go! 

As @agoscinski mentionned, we will have to either rebase this branch on top of master & cleanup commits, or squash-merge it. Let me know if you want to try to rebase this together.

src/rascal/math/kvec_generator.hh
```diff
@@ -120,14 +120,14 @@ namespace rascal {
        * row(i) = i-th vector. If the origin k=(0,0,0) is included, it is
        * always stored in row(0).
        */
-      Matrix_t get_kvectors() const {
+      Eigen::Ref<const Matrix_t> get_kvectors() {
```
Luthaf, line 123: the function can still be declared as const! Any specific reason why you removed the qualifier?

src/rascal/math/kvec_generator.hh
```diff
@@ -120,14 +120,14 @@ namespace rascal {
...
         return Eigen::Ref<const Matrix_t>(this->kvectors);
       }
 
       /** Returns const reference to the Eigen vector containing the norm
        * of all k-vectors within the cutoff without double counting.
        */
-      Vector_t get_kvector_norms() const {
+      Eigen::Ref<const Vector_t> get_kvector_norms() {
```
Luthaf, line 130: Ideally, the function should be marked as const if it returns a const reference.

```suggestion
      Eigen::Ref<const Vector_t> get_kvector_norms() const {
```

PR #394, opened 2021-11-22
Feat/cpp gradient interface enhancements

curiosity54:
Implemented "get number of coefficients" for spherical expansion and invariants on the C++ side with Alex. 

Tasks before review:

          elsewhere in the code.
          auto-formatting is sensible
          own branch


src/rascal/representations/calculator_spherical_invariants.hh
```diff
@@ -254,6 +254,41 @@ namespace rascal {
...
 
+    int get_num_coefficients(int n_species) {
+      auto soap_type = hypers.at("soap_type").get<std::string>();
+      if (soap_type == "RadialSpectrum") {
+        return (n_species * this->max_radial);
+      } else if (soap_type == "PowerSpectrum") {
+        return (n_species * (n_species + 1) * 0.5 * this->max_radial *
+                this->max_radial * (this->max_angular + 1));
+      } else if (soap_type == "BiSpectrum") {
+        int count = 0;
```
felixmusil, line 269:
```suggestion
        int count{0};
```

src/rascal/representations/calculator_spherical_invariants.hh
```diff
@@ -254,6 +254,41 @@ namespace rascal {
...
+                continue;
+              }
+
+              count += 1;
+            }
+          }
+        }
+
+        return (count * math::pow(this->max_radial, 3) *
+                ((n_species + 2) * (n_species + 1) * n_species) / 6);
```
felixmusil, line 290: I trust you here as it's different from the python impl

src/rascal/structure_managers/adaptor_center_contribution.hh
```diff
@@ -258,6 +261,42 @@ namespace rascal {
...
+        throw std::runtime_error(
+            R"(There are no structure to get features from)");
+      }
+
+      size_t n_neighbors{this->get_nb_clusters(2)};
+      Eigen::Matrix<int, Eigen::Dynamic, 5> gradients_info(n_neighbors, 5);
+      gradients_info.setZero();
+      int i_row{0};
+      for (int i = 0; i < static_cast<int>(this->get_size()); i++) {
+        for (int j = 0; j < static_cast<int>(this->nb_neigh[1][i]); j++) {
```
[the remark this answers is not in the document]
felixmusil: Could you use more meaningful iteration variables like `i_center` and and `j_neigh` ?

felixmusil (changes requested):
I feel there should be a way to have a generic version of `get_gradients_info`. Probably it could be a function that takes a manager and returns the matrix, or a generic implementation in `StructureManager`. Why did you repeat the implementation in several managers ?

src/rascal/representations/calculator_spherical_invariants.hh
```diff
@@ -254,6 +254,41 @@ namespace rascal {
     //! the coefficient_subselection input
     bool is_sparsified{false};
 
+    int get_num_coefficients(int n_species) {
+      auto soap_type = hypers.at("soap_type").get<std::string>();
+      if (soap_type == "RadialSpectrum") {
+        return (n_species * this->max_radial);
+      } else if (soap_type == "PowerSpectrum") {
+        return (n_species * (n_species + 1) * 0.5 * this->max_radial *
+                this->max_radial * (this->max_angular + 1));
```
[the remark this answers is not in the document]
agoscinski: Could you tell me the advantage in this case. If we do not return an `int` here, we get a compile error because the output is declared as `int`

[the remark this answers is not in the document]
agoscinski: I put it now around the species, because this should evaluate to an int. Besides removing some unnecessary floating point computation, I however do not see any advantage.

src/rascal/representations/calculator_spherical_invariants.hh
agoscinski: yes Bispectrum number of coefficient was wrong, we checked it

agoscinski:
Because the CRTP made the error messages hard to understand and I thought I just try this implementation, since we anyway don't use any other stacking for gradients (except the upcoming LODE stuff). In terms of development time this was faster than now figuring out the within structure manager solution.

The problem was that I used a `const` function, which works in the adaptors and structure manager collection but not in the structure, since we are not changing any member variable. But in structure manager one needs to create a non `const` Iterator passing `this` to it which changes the `this` object, so it cannot be `const`.

src/rascal/representations/calculator_spherical_invariants.hh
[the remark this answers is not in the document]
felixmusil: doing the static cast here is just to specify clearly what happens implicitly with the return type.

src/rascal/representations/calculator_spherical_covariants.hh
```diff
@@ -290,8 +290,13 @@ namespace rascal {
...
 
-    int get_num_coefficients(int /*n_species*/) {
-      throw std::runtime_error("Not implemented.");
+    int get_num_coefficients(int /*n_species*/) const {
+      std::stringstream error{};
+      error << "Function `get_num_coefficients` in file " << __FILE__
+            << std::endl;
+      throw std::runtime_error(error.str());
+      // throw std::runtime_error(,  " in line ", __LINE__, " is not
+      // implemented.");
```
felixmusil, line 299: remove

src/rascal/structure_managers/structure_manager.hh
```diff
@@ -634,6 +634,38 @@ namespace rascal {
       return this->get_previous_manager()->get_direction_vector(pair);
     }
 
+    /**
+     * Get informations necessary to the computation of gradients. It has
+     * as many rows as the number gradients and they correspond to the index
+     * of the structure, the central atom, the neighbor atom and their atomic
+     * species.
+     *
+     * The shape is (n_structures * n_centers * n_neighbor, 5) while the
```
felixmusil, line 643:
```suggestion
     * The shape is (n_centers * n_neighbor, 5) while the
```

felixmusil (changes requested):
Looks good. After addressing the few more points I will approve.

src/rascal/structure_managers/structure_manager.hh
agoscinski: thanks

felixmusil (approved):
Good to go

PR #401, opened 2022-03-18
Fix imports on Python 3.10

Luthaf:
`collections.Iterable` has been deprecated for a while and was removed in Python 3.10. The correct import is `collections.abc.Iterable`, which was already used in other parts of the code.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,6 +11,7 @@ sphinx>=2.1.2,<=4.0.2
 breathe>=4.14.1
 # Jinja2 v3 breaks nbsphinx
 jinja2==2.11.3
+MarkupSafe==2.0.1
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -11,6 +11,7 @@ sphinx>=2.1.2,<=4.0.2
 breathe>=4.14.1
 # Jinja2 v3 breaks nbsphinx
 jinja2==2.11.3
+MarkupSafe==2.0.1
```
Luthaf, line 14: this is needed because of the pin of jinja to version 2 above, newer versions of MarkupSafe are no longer compatible with it

agoscinski (approved):
Okay

PR #406, opened 2022-05-06
Bugfix: Overlap matrix had extra normalization

kvhuguenin:
Fixes the wrong orthonormalization in rascal for GTO basis.

The code was mixing two conventions for the GTOs.
The overlap matrix was defined using the normalized GTOs, while the rest of the code is using the primitive GTOs (just the r^n times Gaussian part). Now, the coefficients obtained from librascal agree with those from "exact calculations" and those from pyLODE up to a global factor.

The original code was trying to take the extra normalization into account by multiplying the expansion coefficients again with the same normalization factors. Most likely, there was something wrong in the execution of these steps. We are trying to understand this in more detail, but for the time being, the new coefficients agree with the approaches mentioned above and seem to be correct.

Still TODO:
- [ ] Fix some errors associated with this in the documentation.
- [ ] Clean up code to remove code fragments belonging to redundant normalization steps

### DIFF
--- a/src/rascal/representations/calculator_spherical_expansion.hh
+++ b/src/rascal/representations/calculator_spherical_expansion.hh
@@ -403,8 +403,7 @@ namespace rascal {
       void precompute() override {
         this->precompute_radial_sigmas();
         this->precompute_radial_overlap();
-        this->ortho_norm_matrix =
-            this->radial_norm_factors.asDiagonal() * this->radial_ortho_matrix;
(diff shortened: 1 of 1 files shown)


Luthaf:
We will have to also change the regression tests data and the python implementation of GTO in https://github.com/lab-cosmo/librascal/blob/fed28b910d9ba6f8549c935aaf67bdc14051d511/bindings/rascal/utils/radial_basis.py.

Luthaf:
@kvhuguenin should we close this?

PR #410, opened 2022-06-07
Soap test 01

UnixJunkie:
Librascal beginner trying to dump SOAP features of atoms in molecules

### DIFF
--- a/soap_test.py
+++ b/soap_test.py
@@ -1,0 +1,41 @@
+#!/usr/bin/env python3
+
+# Try to dump SOAP vectors for each atom of a molecule
+
+from ase.io import read as ase_read
+from rascal.representations import SphericalInvariants as SOAP
+
+max_radial = 6 # What is this integer? Unit? Reasonable range to explore?
+max_angular = 6 # What is this integer? Unit? Reasonable range to explore?
+soap_type = "PowerSpectrum"  # only physicists know what is a power spectrum...
+#            "RadialSpectrum" # the other possibility; also only known to physicists...
+if soap_type == "RadialSpectrum":
+    max_angular = 0
+
+input_fn = "reference_data/inputs/small_molecules-20.json"
+
+start = 0
+length = 1
+representation_name = "spherical_invariants" # Any other choice?
+target_type = "Atom" # The other choice is Structure; i.e. granularity is per atom or per molecule
+
+hyper_params = {
+    "interaction_cutoff": 3.5, # Angstroms? what is the reasonable range to explore for this value?
+    "cutoff_smooth_width": 0.5, # Angstroms? what is the reasonable range to explore for this value?
+    "max_radial": max_radial,
+    "max_angular": max_angular,
+    "gaussian_sigma_type": "Constant", # other possible choices for this value?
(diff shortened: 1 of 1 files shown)


soap_test.py
```diff
@@ -0,0 +1,73 @@
...
+    "gaussian_sigma_constant": 0.5,
+    "soap_type": soap_type,
+    # 'RadialScaling' is the only other choice
+    "cutoff_function_type": "ShiftedCosine",
+    "normalize": True, # recommended
+    # 'DVR' is the only other choice
+    "radial_basis": "GTO",
+    # # as far as I can tell, global_species has no effect on fixing the
+    # # dimensionality of the SOAP feature vectors
+    # "global_species": global_species
```
Luthaf, line 47: Try also specifying `expansion_by_species_method`:

```suggestion
    "global_species": global_species,
    "expansion_by_species_method": "user defined",
```

It should not be required but something might have felt between the cracks.

soap_test.py
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env python3
+
+# Try to dump SOAP vectors (features) for each atom of each molecule
+
+import sys
+import numpy as np
+from ase.io import read as ase_read
+from rascal.representations import SphericalInvariants as SOAP
+
+# What is this integer? Unit? Reasonable range to explore?
```
Luthaf, line 10: that's the number of radial basis, i.e. the radial resolution. Typical range is 4-9, GTO can go up to 22, DVR is mostly unlimited.

soap_test.py
```diff
@@ -0,0 +1,73 @@
...
+max_radial = 6
+# What is this integer? Unit? Reasonable range to explore?
```
Luthaf, line 12: that's the number of spherical harmonics being used for the basis, i.e. the angular resolution. Typical range is 4-8, you can go higher if you want/need

soap_test.py
```diff
@@ -0,0 +1,73 @@
...
+max_angular = 6
+# what is a power spectrum?
```
Luthaf, line 14: The power spectrum is the standard SOAP descriptor, but there are many other that can be constructed. Power spectrum is a 3-body descriptor, while the radial spectrum is a 2-body representation. 

Bi-Spectrum is the name of the 4-body representation, and then we just refer to them with the body order.

soap_test.py
```diff
@@ -0,0 +1,73 @@
...
+    "interaction_cutoff": 3.5,
+    # Angstroms? what is the reasonable range to explore for this value?
+    "cutoff_smooth_width": 0.5,
+    "max_radial": max_radial,
+    "max_angular": max_angular,
+    "gaussian_sigma_type": "Constant", # only possible choice
+    # Angstroms? what is the reasonable range to explore for this value?
+    "gaussian_sigma_constant": 0.5,
+    "soap_type": soap_type,
+    # 'RadialScaling' is the only other choice
```
Luthaf, line 40: RadialScaling is defined in https://pubs.rsc.org/en/content/articlelanding/2018/CP/C8CP05921G, for the QM9 example

UnixJunkie:
Indeed, the combination of both 
```python
"global_species": global_species,
"expansion_by_species_method": "user defined",
```
fixes the length of the SOAP vectors.
Thanks a lot!

PR #414, opened 2022-06-14
Adapt cpp representation input paramaters compatible with the current lammps-rascal interface

agoscinski:
This is a PR from this commit.
https://github.com/lab-cosmo/librascal/pull/376/commits/2a879b7ffabefb21f937366c0b944a6edee895ec

To use the current lammps-rascal interface conveniently, we however don't take the newest format version #376 which is cleaner, but the older version compatible with the branch `feat/lammps_preparations` which is used for the current version of the lammps-rascal interface.

EDIT:
This is so people do not need to have to install a librascal version just to format the ML model json correctly compatible with lammps-rascal

### DIFF
--- a/bindings/rascal/models/kernels.py
+++ b/bindings/rascal/models/kernels.py
@@ -107,9 +107,21 @@ class Kernel(BaseIO):
 
     def _set_data(self, data):
         super()._set_data(data)
+        # allows to load deprecated models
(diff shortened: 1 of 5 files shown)


PicoCentauri (approved):
LGTM! Thanks @agoscinski

PR #416, opened 2022-08-04
Bayesian optimization tutorial

hurricane642:
In this PR I add an example of using Bayesian optimization for Power Spectrum. Most of the code is based on the first `librascal` example, I have not changed the meaningful part of it, adding only a few things about the optimization itself. This is a fairly simple example with the simplest functions from the `scikit-optimize` library, but I hope it is enough for the first acquaintance. This is my first time writing examples for the library, so any suggestions and edits are welcome.

### DIFF
--- a/examples/Bayes_tutorial-librascal.ipynb
+++ b/examples/Bayes_tutorial-librascal.ipynb
@@ -192,12 +192,20 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "/home/kliavine/molecular-clouds-models/i-PI/zundel\n"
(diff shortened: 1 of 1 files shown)


max-veit:
Very good, thanks for adding this! I probably won't have time for a full review anytime soon (maybe late next week but no promises). But right now my main comment is that there's a lot of utility code, especially in the second-to-last cell, that would benefit from either being moved to its own "notebook utilities" module or broken down and explained step-by-step in the notebook.

BTW the tests are failing because of an update in scikit-cosmo a while back; there's a fix in 95aedf9c

PR #417, opened 2022-10-19
solvers changes 

DavideTisi:
(Describe your pull request in 1-2 sentences)
added solvers rkhs, rkhs-qr and qr from gaptools also added a option "lstsq" to uses lstsq with the old method ( numpy.linalg.lstsq(rcond=-1) )

train_gap_model works out-of-the-box with the older versions. The default solver stays the same, other solver as simply made available by the PR

Fix # (optional)

- added solvers rkhs, rkhs-qr and qr from gaptools
- train_gap_model is "retro-compatible" but uses as default RKHS-QR
- added solver "lstsq" which uses numpy.linalg.lstsq(rcond=-1)

Tasks before review:

          auto-formatting is sensible

### DIFF
--- a/bindings/rascal/models/krr.py
+++ b/bindings/rascal/models/krr.py
@@ -10,8 +10,191 @@ Public functions:
 from ..utils import BaseIO
 from ..lib import compute_sparse_kernel_gradients, compute_sparse_kernel_neg_stress
 
+import scipy
 import numpy as np
 import ase
+import warnings
+
+
+class SparseGPRSolver:
+    """
+    A few quick implementation notes, docs to be done.
+
+    This is meant to solve the sparse GPR problem::
+
+        b = (KNM.T@KNM + reg*KMM)^-1 @ KNM.T@y
+
+    The inverse needs to be stabilized with application of a numerical jitter,
+    that is expressed as a fraction of the largest eigenvalue of KMM
+
+    Parameters
+    ----------
+    KMM : numpy.ndarray
+        KNM matrix
+    regularizer : float
+        regularizer
+    jitter : float
+        numerical jitter to stabilize fit
+    solver : {'RKHS-QR', 'RKHS', 'QR', 'solve', 'lstsq'}
+        Method to solve the sparse KRR equations.
+
+        * RKHS-QR: TBD
+        * RKHS: Compute first the reproducing kernel features by diagonalizing K_MM
+          and computing `P_NM = K_NM @ U_MM @ Lam_MM^(-1.2)` and then solves the linear
+          problem for those (which is usually better conditioned)::
+
+              (P_NM.T@P_NM + 1)^(-1) P_NM.T@Y
+
+        * QR: TBD
+        * solve: Uses `scipy.linalg.solve` for the normal equations::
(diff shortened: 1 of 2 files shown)


bindings/rascal/models/krr.py
```diff
@@ -7,13 +7,168 @@
     compute_KNM             Compute GAP kernel of a set of structures
     train_gap_model         Train a GAP model given a kernel matrix and sparse points
 """
+from ast import Raise
```
Luthaf, line 10: wat? Not sure why you would need this =)

bindings/rascal/models/krr.py
```diff
@@ -433,6 +593,19 @@ def train_gap_model(
...
+           (K_NM.T@K_NM + K_MM)^(-1) K_NM.T@Y
+        "lstsq" uses numpy.linalg.solve( rcond=-1) for the normal equations:
+           (K_NM.T@K_NM + K_MM)^(-1) K_NM.T@Y
+        "RKHS" computes first the reproducing kernel features by diagonalizing K_MM
+        and computing P_NM = K_NM @ U_MM @ Lam_MM^(-1.2) and then solves the linear
+        problem for those (which is usually better conditioned)
+           (P_NM.T@P_NM + 1)^(-1) P_NM.T@Y
+        "RKHS-QR"
+        "QR"
+        by default, "RKHS"
```
Luthaf, line 608: I think we should leave the default at lstsq for now, since that's what we used most of the time. Alternatively, I could see an argument for QR, which is what QUIP is using, and should be more stable. RKHS solvers seems more experimental

bindings/rascal/models/krr.py
PicoCentauri: I think this is an automatic change from vscode :-)

bindings/rascal/models/krr.py
agoscinski: Just for information: according to michele the RKHS-QR solver is the most stable, then the RKHS, then the QR.

I would add it in the documentation that the QR solver is the same as in QUIP, but I am fine with any solver as the default one. I can see an argument in each case.

Luthaf: My point is we want this to be a small & quick PR adding the option to use other solvers so that people using librascal can have model that learn, and for that, I think we need to have the smallest set of changes possible. 

Changing the default can be done later, or in equisolve/equilearn once we understand what's going on.

bindings/rascal/models/krr.py
```diff
@@ -444,6 +645,9 @@ def train_gap_model(
         In Handbook of Materials Modeling (pp. 1–27). Springer, Cham.
         https://doi.org/10.1007/978-3-319-42913-7_68-1
     """
+    if solver=='lstsq' and rcond is None : 
+        print('Warning solver = lstsq and rcond=None is deprecated.\
```
Luthaf, line 649:
```suggestion
import warnings
        warnings.warn('Warning solver = lstsq and rcond=None is deprecated.\
```

Luthaf (approved):
Thanks a lot!!!!

PR #429, opened 2023-09-14
Changed dependencies to reflect rebranding of skcosmo

ceriottm:
Librascal still depends on skcosmo for feature selection. Given the rebranding, we must update to skmatter.
There are no open bugs, but dependency on a dead package is a bad idea so it's worth to do the renaming.

Just sed s/skcosmo/skmatter. Fingers crossed that tests won't break

No more tests or documentation are needed.


ceriottm:
OK sorry seems I had just failed to update my librascal installation. Sorry about the noise

